PR #9, opened 2023-10-20
Added option for custom logger

### LINKED ISSUE #2: Log with a custom logger instead of the `slog.DefaultHandler`
I want to use my own logger.

An option in the app declaration, of type `slog.Handler`.

nhAnik:
Closes #2

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -1,6 +1,7 @@
 package op
 
 import (
+	"log/slog"
 	"net/http"
 )
 
@@ -24,7 +25,7 @@ func Register[T any, B any](s *Server, method string, path string, controller fu
 	if isGo1_22 {
 		fullPath = method + " " + path
 	}
-	s.logger.Debug("registering openapi controller " + fullPath)
+	slog.Debug("registering openapi controller " + fullPath)
 	s.mux.Handle(fullPath, withMiddlewares(http.HandlerFunc(httpHandler[T, B](s, controller)), s.middlewares...))
 
 	RegisterOpenAPIOperation[T, B](s, method, path)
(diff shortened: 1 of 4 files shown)


EwenQuim:
Thank you for your PR !

I see that there are still occurences of slog.Debug in `ctx.go` and `deserialization.go`.
I know that `s.logger` is not accessible from there. So there are 2 options :

- set a logger field on Ctx (but it will not be available on serialization.go)
- use a package-level global variable `Logger`

I think it might be a good idea to use a global variable. I don't see why it would cause a problem : nobody will use 2 instance of a op.Server with 2 different loggers. 
The only issue about this is that it have to be accessible from another package if we create one (for middlewares for example). So we need to make this variable public. Once again, I don't see why it would be a problem.

What do you think about using a global variable for the logger ?

Also, could you please run `make ci` before pushing please? Thank you!

Sorry for the rework and thank you again for the PR.

nhAnik:
Thanks for your recommendation!

Currently there is no way to access `s.logger` from `deserialization.go`. So, I think we have no option but to use a default logger.

Also, IMHO there is no need to define a public default logger in a package as we can already use `slog.Default()`.  And if we need to pass a custom handler, we can use `slog.SetDefault` like below:
```go
func WithHandler(handler slog.Handler) func(*Server) {
	return func(*Server) {
		slog.SetDefault(slog.New(handler))
	}
}
```
In this case we can use the current log APIs from slog package like `slog.Info` etc.

What do you think about it? If you think if it is okay I can push my new changes. Also, you can give any different idea.

Thanks again.

EwenQuim:
> Thanks for your recommendation!
> 
> Currently there is no way to access `s.logger` from `deserialization.go`. So, I think we have no option but to use a default logger.
> 
> Also, IMHO there is no need to define a public default logger in a package as we can already use `slog.Default()`.

I was afraid that this would cause the user's app to use our logger. I thought the `slog.Default` would be shared between our lib and the user's app. That would have been bad. 

So I tested and it looks that it isn't the case, proof here:
image


With this code in user's app:

```go
        myHandler := tint.NewHandler(os.Stderr, &tint.Options{
		AddSource:  true,
		Level:      logLevel,
		TimeFormat: "15:04:05",
	})

	slog.SetDefault(slog.New(
		myHandler,
	))

	app := op.NewServer(
		op.WithHandler(slog.NewJSONHandler(os.Stderr, nil)),
	)
```

So your idea seems to be a very good one :) Thank you very much for your work. I will approve it, just don't forget to run `make ci` please!

EwenQuim:
Thank you very much for your contribution! 💪🏼

PR #18, opened 2023-10-30
feat: add generate openapi options

rizerkrof:


openapi.go
```diff
@@ -47,35 +58,79 @@ func (s *Server) GenerateOpenAPI() {
...
+	if options.disableSaveOnDisk == false {
+		jsonLocalPath := "docs/openapi.json"
+		if options.jsonLocalPath != "" {
+			jsonLocalPath = options.jsonLocalPath
+
+			match, _ := regexp.MatchString(`^[\.a-zA-Z0-9\-\_]+(.json)$`, jsonLocalPath)
+
+			if match == false {
+				err := errors.New("Value of 'jsonLocalPath' option is not valid")
+				slog.Error("Error writing json spec", "error", err)
```
EwenQuim, line 70: It's a bit confusing to create an error when it's just being used in a log, so I think it's better to write it directly.

openapi.go
```diff
@@ -47,35 +58,79 @@ func (s *Server) GenerateOpenAPI() {
...
+		// Serve spec as JSON
+		jsonUrl := "/swagger/doc.json"
+		if options.jsonUrl != "" {
+			jsonUrl = options.jsonUrl
+
+			match, _ := regexp.MatchString(`^\/[\/\.a-zA-Z0-9\-\_]+(.json)$`, jsonUrl)
+
+			if match == false {
+				err := errors.New("Value of 'jsonUrl' option is not valid")
+				slog.Error("Error serving openapi json spec", "error", err)
```
EwenQuim, line 102: Same

serve.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 func (s *Server) Run() {
-	go s.GenerateOpenAPI()
+	go s.GenerateOpenAPI(openApiOptions{})
```
EwenQuim, line 10: C'est bien mais du coup pas moyen de changer les options ?

Regarde `options.go`, on utilise le pattern des "function options", en gros je verrai bien un WithOpenApiConfig, avec en paramètre `openApiOptions` qu'on renommerai `OpenApiConfig` (avec une majuscule et finissant par config

openapi.go
```diff
@@ -47,35 +58,79 @@ func (s *Server) GenerateOpenAPI() {
...
 
 	// Write spec to docs/openapi.json
-	err = os.MkdirAll("docs", 0o750)
-	if err != nil {
-		slog.Error("Error creating docs directory", "error", err)
-	}
-	f, err := os.Create("docs/openapi.json")
-	if err != nil {
-		slog.Error("Error creating docs/openapi.json", "error", err)
+	if options.disableSaveOnDisk == false {
```
EwenQuim, line 61: The fact of having 2 large, hyper-long `if` is a bit of an indicator of poor code structure. Personally, I'd refactor it into 2 new functions that are called or not depending on whether the option is called or not. What do you think?

openapi.go
```diff
@@ -47,35 +58,79 @@ func (s *Server) GenerateOpenAPI() {
...
+			w.Header().Set("Content-Type", "application/json")
+			_, _ = w.Write(dataJSON)
+		})
+
+		// Swagger UI
+		swaggerUrl := "/swagger"
+		if options.swaggerUrl != "" {
+			swaggerUrl = options.swaggerUrl
+
+			match, _ := regexp.MatchString(`^\/[\/\.a-zA-Z0-9\-\_]+[\.a-zA-Z0-9\-\_]$`, swaggerUrl)
```
EwenQuim, line 117: It's nice to have a param validation ! Very good idea 🎉 
1. Note that building a regex takes time: don't do it in a loop, for example. In this case, you only have to do it once, so don't worry. Alternatively, you can compile it at build time, as a local variable, as in the `params.go` file, instead of building it at runtime.
2. Every regex must have test cases. Sorry but we're making a premium quality framework! 😉

EwenQuim (changes requested):

EwenQuim:
Thank you @rizerkrof for this PR 🎉

PR #20, opened 2023-11-08
feat: add http logger middleware

rizerkrof:


EwenQuim:
@rizerkrof we're dropping this in favor of something more up to date, hope you don't mind!

PR #42, opened 2024-02-14
Add search to documentation & minor cleanup

Jeremie-Chauvel:
Add search to the documentation using https://github.com/praveenn77/docusaurus-lunr-search

then some minor cleanup of the documentation: the github workflow use npm so I removed the yarn.lock, updated the dependencies and the node version to the latest lts

image


.github/workflows/pages.yml
```diff
@@ -32,11 +32,10 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: "18"
-          
+          node-version: "20"
 
       - name: Install dependencies
-        run: npm install
+        run: npm ci
```
EwenQuim, line 38: Normal @Jeremie-Chauvel ?

PR #43, opened 2024-02-14
Fix video size on documentation home page

Jeremie-Chauvel:


hugoleborso:
What a beautiful title !

PR #48, opened 2024-02-19
feat: add ability to set header parameters

kaazedev:


EwenQuim:
Very good idea. 

I'll work on a more complete API to support these: not only the title, but more data  :)

EwenQuim:
Is it possible to add some tests ?

I will merge it into `main` then work before next release to add more parameters (description, example, required...)

PR #52, opened 2024-02-22
Update ci/cd go version to 1.22

Kronopt:
I noticed the testing ci/cd pipeline is using go version **1.21**.
Both go.mod and go.work target go version **1.22**.
I just thought the project should be tested using the same version as it targets.

Feel free to close the PR if you think this is incorrect.

Cheers! 🚀


EwenQuim:
Indeed! 

Thank you for your contribution

PR #61, opened 2024-03-02
fix: Better CORS handling

guerinoni:
This inject cors as option using `WithCors` and also fix the usage
because before of this cors were not applied to server.

### DIFF
--- a/options.go
+++ b/options.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/golang-jwt/jwt/v5"
+	"github.com/rs/cors"
 )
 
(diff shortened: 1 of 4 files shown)


options.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/golang-jwt/jwt/v5"
+	"github.com/rs/cors"
 )
```
EwenQuim, line 15: `rs/cors` is a middleware, I don't want to add it to fuego : users should be free to chose the middlewares they want (some users don't need cors)

EwenQuim:
Thanks for the work and sorry to close this pull request!

Issue #60, opened 2024-03-01
Hide routes from swagger UI


epentland:
**Is your feature request related to a problem? Please describe.**
I am building an application that has both web and api interfaces, and I want the web interfaces hidden from the UI. It shows two getConversations on the UI, one for the api, one for the web.

**Describe the solution you'd like**
Be able to add a function to fuego.Route, or fuego.Server (as a part of a group) to hide that interface from the UI

**Describe alternatives you've considered**
I tried seeing if adding or removing tags would solve the problem, but I wasn't able to do it this way.

**Additional context**
Awesome repo!

EwenQuim:
Good idea, will work on it the following weeks.

Do you think we should not declare these routes on the OpenAPI spec or only hide them in the UI?

epentland:
Much cleaner! I was able to put something together for the group.Hide() quickly, I will make a PR tomorrow.

I just added a DisableOpenapi bool field to the server, made the server.Hide() function, and on the mux.register function, if s.DisableOpenapi is true, it just returns the route with an empty openapi3.Operation{}

That being said, taking a look at the Route struct, it seems like it would be a bit harder to add the Route.Hide() function. It doesn't have access to the server, so you would need to pass the server as an arg and delete the api spec that was already registered vs preventing it from being created in the first place.

EwenQuim:
Hi @epentland.

Thank you for working on it!

Related Resources:
1. Just so you know, in the next release wednesday, we will group the HTML routes under a common tag.
2. One of the next features will be custom Group Tags #62

<img width="403" alt="image" src="https://github.com/go-fuego/fuego/assets/46993939/01e43d82-92bc-4099-b621-a09b759fa648">

<img width="353" alt="image" src="https://github.com/go-fuego/fuego/assets/46993939/7af41b65-b57f-44f8-93aa-f62858c8ffb4">



---

This issues tracks
- disable OpenAPI for a group
- disable OpenAPI for a route

Have you made progress on the PR? I want to know if I can spend time on this feature or if you've already done it.

Thanks again!

EwenQuim:
Closed by #63

PR #63, opened 2024-03-04
Hide from Openapi Spec

epentland:
Added being able to hide routes from the server, by using server.Hide()

Implements server.Hide() from #60

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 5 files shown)


mux_test.go
```diff
@@ -245,6 +246,19 @@ func TestDeleteStd(t *testing.T) {
 	require.Equal(t, w.Body.String(), "test successful")
 }
```
EwenQuim, line 248: Thank you for this tests ! More tests that cover more cases would be wonderful

- [ ] with groups
- [ ] nested groups

EwenQuim:
This closes #60 and the release will be made before the end of the week

PR #65, opened 2024-03-05
feat: add option for custom validator

vladz:
This pull request adds WithValidator option to Server which replaces default validator.


EwenQuim:
Very good use case & tech, a pleasure to read PRs like these.

Thank you very much for your contribution!

PR #70, opened 2024-03-11
add WithAddr option, adjusted WithPort func to only handle port

nengberg:
# Description

This PR adds the option `WithAddr` to the server, which lets you specify the full "host:port" for the server to be run at. The `WithPort` func currently allows you to do this, but it's not obvious that you can. I believe it's better to be explicit about this, hence the addition.

With this change I've also deprecated `WithPort` but kept it for backwards compatibility. Up to you how you want to do it. One case that is not covered is if `WithPort` has been used to provide the complete addr: `WithAddr(localhost:8080)`.

I've also updated the log rows and the example in the README.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ type MyResponse struct {
 
 func main() {
 	s := fuego.NewServer(
-		fuego.WithPort("8088"),
+		fuego.WithAddr("localhost:8088"),
 	)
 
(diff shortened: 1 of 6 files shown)


EwenQuim:
Good job ! I will read this PR more carefully this afternoon but it's a good idea :)

PR #72, opened 2024-03-13
feat: support decoding x-yaml content-type in ContextWithBody

dylanhitt:
The following adds the ability to submit yaml as a content type to be decoded ContextWithBody. I'm not quite sure if I did this probably as this comment [is throwing me off but it could just be a mistake?](https://github.com/go-fuego/fuego/commit/3bb79ad6313b8b81a4d9c96ad07954648ebf978f#diff-c4dd82978a1a49bcd620aff72208bbd8114f09341d701e188c247ce3e3511117R104)

Anyhow let me know your thoughts. I'm open to changing things up if this doesn't suffice. If there isn't any interest let me know as well. 

The pattern I chose here seems like it could be extended to XML easily as well.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -363,6 +363,8 @@ func body[B any](c ContextNoBody) (B, error) {
 		err = errReadingString
 	case "application/x-www-form-urlencoded", "multipart/form-data":
 		body, err = readURLEncoded[B](c.Req, c.readOptions)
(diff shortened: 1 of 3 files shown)


EwenQuim:
Very good idea, I didn't know it could be useful to decode yaml to be fair. 

Thank you for your contribution 🎉

PR #73, opened 2024-03-13
docs: ContextWithBody clarification

dylanhitt:
Found this and tripped me up today as it seemed like the usage of ContextWithBody was on it's way to being deprecated with the current language that is there.


EwenQuim:
Thanks !

PR #96, opened 2024-03-22
Tutorial CRUD Modification

minpeter:
Update the document to match the updates in #95.


minpeter:
The Tooltitude for Go (GoLang) extension was the cause. I'll repair the dependencies :)

minpeter:
If you have the time, please try following the tutorial. There seems to be a bug where the delete method leaves the route group.

image

EwenQuim:
Thank you for this docs clarification!

PR #99, opened 2024-03-22
feat: add the ability to read xml from body

dylanhitt:
Add the ability to read XML from body.

I elected not to duplicate all the same test that is there for json i.e transform readtwice etc as it's covered pretty well. If you want that for this let me know.


EwenQuim (approved):
Very clean, thanks!

PR #102, opened 2024-03-25
Draft: fix: middleware ordering

### LINKED ISSUE #98: Clarification on Middleware ordering
Ran into this today and it tripped me up. 

```
// Admin routes
admin := fuego.Group(root, "")
fuego.Use(admin, auth.EnsureValidToken)
fuego.Use(admin, auth.IsAdmin)
```

Typically every other lib i've used when registering middleware they are invoked in the order they are registered. So for instance EnsureValidToken would be invoked first in the above example. 

However, in the case of fuego this seems to be the opposite. In this case IsAdmin is invoked first. Is this intentional? 

Current workaround is to just flip them of course. If it is a bug let me know I'd be happy to fix it. However,  you're probably gonna beat me to it 😄, by the time I get back to a computer.

dylanhitt:
closes #98 

Here's the fix. I'll get the tests up soon. Currently need this to get through the work day. 

Thanks

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -156,8 +156,8 @@ func RegisterStd(s *Server, method string, path string, controller func(http.Res
 }
 
 func withMiddlewares(controller http.Handler, middlewares ...func(http.Handler) http.Handler) http.Handler {
-	for _, middleware := range middlewares {
-		controller = middleware(controller)
+	for i := len(middlewares) - 1; i >= 0; i-- {
+		controller = middlewares[i](controller)
 	}
 	return controller
 }
--- a/mux.go
+++ b/mux.go
@@ -88,7 +88,7 @@ func Register[T any, B any, Contexted ctx[B]](s *Server, method string, path str
 func register[T any, B any](s *Server, method string, path string, controller http.Handler, middlewares ...func(http.Handler) http.Handler) Route[T, B] {
 	fullPath := method + " " + s.basePath + path
 
-	allMiddlewares := append(middlewares, s.middlewares...)
+	allMiddlewares := append(s.middlewares, middlewares...)
 	s.Mux.Handle(fullPath, withMiddlewares(controller, allMiddlewares...))
 
 	if s.DisableOpenapi {
--- a/mux_test.go
+++ b/mux_test.go
@@ -20,6 +20,101 @@ func dummyMiddleware(handler http.Handler) http.Handler {
 	})
 }
 
+// orderMiddleware sets the X-Test-Order Header on the request and
+// X-Test-Response header on the response. It is
+// used to test the order execution of our middleware
+func orderMiddleware(s string) func(http.Handler) http.Handler {
+	return func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			orderValue := strings.Split(r.Header.Get("X-Test-Order"), ",")
+			orderValue = append(orderValue, s)
+
+			r.Header.Set("X-Test-Order", strings.Join(orderValue, ","))
+			w.Header().Set("X-Test-Response", "response")
+			handler.ServeHTTP(w, r)
+		})
+	}
+}
+
+// TestUse is used to mainly test the ordering of middleware execution
+func TestUse(t *testing.T) {
+	t.Run("base", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


EwenQuim:
That's weird. We can clearly see here that it makes future uses of `Use` after 

```go
// Your proposition
func Use(s *Server, middlewares ...func(http.Handler) http.Handler) {
	s.middlewares = append(middlewares, s.middlewares...)
}
```


```go
fuego.Use(cache)
fuego.Use(compress)
```

equivalent to 

```go
s.middlewares = append(cache, s.middlewares...) // [cache]
s.middlewares = append(compress, s.middlewares...) // [compress, cache], not [cache, compress] as wanted
```

I might need some tests to understand, I'll see this tomorrow or the day after

dylanhitt:
Right your ordering there makes sense, in terms of when you use `Use`. However, `register` loops from the beginning of the slice. So it's in reverse. I think the code I places here isn't fully correct. And it may be simpler to just reverse the registration loop.  

```
func withMiddlewares(controller http.Handler, middlewares ...func(http.Handler) http.Handler) http.Handler {
	for _, middleware := range middlewares {
		controller = middleware(controller)
	}
	return controller
}
```

I think using numbers would make it clearer

```golang
group := fuego.Group(s, "/v1"
fuego.Use(group, 1)
fuego.Use(group, 2)
fuego.Use(group, 3)
```

My expectation here would 1 executes before 2 and 3. So when this proposed changed we end up with a slice that looks like

```
[3, 2, 1]
```

The loop then registers 3 -> 2 -> 1 -> controller which is the equivalent of a line code like this

```goalnge
1middleware(2middleware(3middleware(conrtoller)))
```

EwenQuim:
Yes exactly ! We'll reverse the loop order of `withMiddlewares`

dylanhitt:
@EwenQuim Here's my proposal for how we should test it. I will of course add more testing around groups and the other possible combinations of adding middleware i.e Group and on routes directly. 

Let me know what you think or if you have a better idea or if you like this and have a thought to add another subset of tests somewhere else be my guest. 

Thanks. Fun bug.

mux.go
```diff
@@ -88,7 +88,7 @@ func Register[T any, B any, Contexted ctx[B]](s *Server, method string, path str
 func register[T any, B any](s *Server, method string, path string, controller http.Handler, middlewares ...func(http.Handler) http.Handler) Route[T, B] {
 	fullPath := method + " " + s.basePath + path
 
-	allMiddlewares := append(middlewares, s.middlewares...)
+	allMiddlewares := append(s.middlewares, middlewares...)
```
dylanhitt, line 91: Turns out this was still needed when adding middlewares at route level.

mux_test.go
```diff
@@ -20,6 +20,101 @@ func dummyMiddleware(handler http.Handler) http.Handler {
 	})
 }
 
+// orderMiddleware sets the X-Test-Order Header on the request and
+// X-Test-Response header on the response. It is
+// used to test the order execution of our middleware
+func orderMiddleware(s string) func(http.Handler) http.Handler {
+	return func(handler http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			orderValue := strings.Split(r.Header.Get("X-Test-Order"), ",")
```
EwenQuim, line 29: `r.Header.Add("X-Test-Order", s)` would do the job

mux_test.go
```diff
@@ -20,6 +20,101 @@ func dummyMiddleware(handler http.Handler) http.Handler {
...
+
+		orderValue := strings.Split(r.Header.Get("X-Test-Order"), ",")
+		require.Equal(
+			t, []string{
+				"Start!", "First!", "Second!", "Third!", "Fourth!", "Fifth!",
+			},
+			orderValue,
+		)
+	})
+}
```
EwenQuim, line 116: Perfect tests, thanks!

EwenQuim (commented):
Thank you for this fix and great tests!

I think a test with `Group` might be a final touch to this PR

I'll be merging+release this as soon as it is done!

dylanhitt:
I always forget headers.Add exists..

dylanhitt:
I always forget headers.Add exists..

PR #104, opened 2024-03-28
feat: add option to pretty print json doc output

dylanhitt:
Provide the ability to pretty print to the .json file

Bit of a vanity thing, but I have some colleagues asking for it. What does make this nice is that it provides the ability to pretty print in CI and we can place the file into a npm package (that's generated from the spec) to be consumed by downstream API users. It also makes diffs useful source control managers.

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -59,7 +59,7 @@ func (s *Server) OutputOpenAPISpec() openapi3.T {
 	}
 
 	// Marshal spec to JSON
-	jsonSpec, err := json.Marshal(s.OpenApiSpec)
+	jsonSpec, err := s.marshalSpec()
 	if err != nil {
 		slog.Error("Error marshalling spec to JSON", "error", err)
 	}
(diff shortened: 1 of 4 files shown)


EwenQuim:
Good idea.

Very easy to implement with [`json.MarshalIndent`](https://pkg.go.dev/encoding/json#MarshalIndent).

dylanhitt:
I'm a bit confused this is the implementation?

```
type OpenAPIConfig struct {
	DisableSwagger   bool                              // If true, the server will not serve the swagger ui nor the openapi json spec
	DisableLocalSave bool                              // If true, the server will not save the openapi json spec locally
	SwaggerUrl       string                            // URL to serve the swagger ui
	UIHandler        func(specURL string) http.Handler // Handler to serve the openapi ui from spec url
	JsonUrl          string                            // URL to serve the openapi json spec
	JsonFilePath     string                            // Local path to save the openapi json spec
	PrettyFormatJson bool                              // Pretty prints the open api spec with proper json indentation
}
```

Is configurable with `PrettyFormatJson`. Are you confusing this pull request as an issue?

EwenQuim:
Oh sorry I thought it was an issue, I didn't look at the code. Reviewing it right now!

EwenQuim:
Very good work here! Thanks 🎉

EwenQuim:
Released with patch version 0.13.4 if you want to try (Go packages proxy might be a little slow to index new tags so it might be available only in a few minutes)

PR #112, opened 2024-04-05
Read Binary body with application/octet-stream Content-Type header

### LINKED ISSUE #108: How can I send binary data to the endpoint?
I am trying is to send binary blob data into endpoint and I see error `Error Decoding Failed`, so I guess I am doing something wrong.

```
	fuego.Put(s, "/blob", func(c *fuego.ContextWithBody[any]) (any, error) {
		body, err := c.Body()
		if err != nil {
			return nil, err
		}

		return body, nil
	})
```

with `curl --request PUT --url http://localhost:9999/blob --data-binary @blob`

EwenQuim:
Fixes #108


ctx.go
```diff
@@ -423,11 +426,22 @@ func body[B any](c ContextNoBody) (B, error) {
...
 	case "application/xml":
-		return readXML[B](c.Req.Context(), c.Req.Body, c.readOptions)
+		body, err = readXML[B](c.Req.Context(), c.Req.Body, c.readOptions)
 	case "application/x-yaml":
-		return readYAML[B](c.Req.Context(), c.Req.Body, c.readOptions)
+		body, err = readYAML[B](c.Req.Context(), c.Req.Body, c.readOptions)
 	case "application/json":
 		fallthrough
+	case "application/octet-stream":
+		// Read c.Req Body to bytes
```
dylanhitt, line 435: It may seem like overkill, but I really like the how we've been using helper functions for this. How about a helper function here too?

Just a nitpick feel free to close.

documentation/docs/guides/serialization.mdx
```diff
@@ -6,6 +6,23 @@ Fuego automatically serializes and deserializes inputs and outputs with standard
 
 <FlowChart selected="resp" />
 
+## Deserialize binary data
+
+If you just want to read the body of the request as a byte slice, you can use the `[]byte` receiver type.
+
+Don't forget to set the request `Content-Type` header to `application/octet-stream`. 
+
+```go
```
dylanhitt, line 15: 👍

ctx.go
```diff
@@ -423,11 +426,22 @@ func body[B any](c ContextNoBody) (B, error) {
 	case "application/x-www-form-urlencoded", "multipart/form-data":
 		body, err = readURLEncoded[B](c.Req, c.readOptions)
 	case "application/xml":
-		return readXML[B](c.Req.Context(), c.Req.Body, c.readOptions)
+		body, err = readXML[B](c.Req.Context(), c.Req.Body, c.readOptions)
 	case "application/x-yaml":
-		return readYAML[B](c.Req.Context(), c.Req.Body, c.readOptions)
+		body, err = readYAML[B](c.Req.Context(), c.Req.Body, c.readOptions)
 	case "application/json":
 		fallthrough
```
dylanhitt, line 433: Is this not a bug now? Wouldn't the fallthrough hit the next case?

Like [this playground](https://go.dev/play/p/0DxTOo7PbuX)

ctx.go
EwenQuim: I'm just doing it as a way to harmonize the serialization within the Fuego controllers.

As it is really easy to do it with std lib, people using Fuego helper functions might better use directly the std lib I guess.

ctx.go
EwenQuim: There is no fallthrough for yaml ! Only for JSON, to allow the default case to be JSON

EwenQuim:
Thank you for reading this @dylanhitt, very interesting comments.

I hope you take no offense even if I merge this without modifying my code

ctx.go
dylanhitt: Yeah we don't have a test where we explicitly declare `application/json` content header. Which is why we're missing this.

EwenQuim: Oh yes you're right, I thought you were talking about the lines above. We definitely need a test for this case so it never happens again

EwenQuim:
Thank you for your catch @dylanhitt , you saved next release ;) 

I added a test to explicitely cover this case

dylanhitt (approved):
👍

PR #114, opened 2024-04-05
docs: remove usage of WithPort from docs in favor of WithAddr

dylanhitt:


documentation/docs/guides/options.md
```diff
@@ -27,13 +27,13 @@ func main() {
 
 ## Some options
 
-### Port
+### Address
```
EwenQuim, line 30: We still need to explicitely mention Port for people searching for "port" in the docs.

```suggestion
### Address (and port)
```

EwenQuim (changes requested):

documentation/docs/guides/options.md
dylanhitt: Good point

dylanhitt:
Thanks looking forward to pinning to the next release version!

😄

PR #115, opened 2024-04-05
Default tags are based on groups + possibility to disable auto tags

### LINKED ISSUE #105: Ability to disable default tag on routes
Currently we append tags to every route using the ResponseBody. In some cases users do not want this. For instance I would like only the `v2` tag for all routes in a server group. As I want my outputted node client to have all of it's function in the `v2api.js` file. This is still the result however a consequence of this is that we end up with another file exporting the same function name, ultimate failing the typescript buld. This leads to needed to do something along the lines of

```golang
rootProtected := fuego.Group(root, "/v2")

fuego.Get(rootProtected, "/accounts", h.Accounts).Tags("v2")

fuego.Get(rootProtected, "/clusters", h.Clusters).Tags("v2")
```

This essentially makes using tags at the group level not very useful. 

Use the [OpenAPIConfig](https://github.com/go-fuego/fuego/blob/main/options.go#L19) struct to configure this behavior. I.e add a field of `NoDefaultTags` or whatever (open to name suggestions) 

The alternative is above.

EwenQuim:
Fixes #105


dylanhitt (approved):
I love nice it is to add features to this repos 👍

Issue #113, opened 2024-04-05
Make it easier for Fuego to use TLS 


redecs:
Starting from my inquiry about [HTTP2 support](https://github.com/go-fuego/fuego/issues/111), it is pretty clear that TLS configuration for Fuego could be easier. 

It looks like the consensus is to add a new method, `RunTLS(cert, key)` that will do the same thing as Run does but call `ListenAndServeTLS` at the end, with the supplied certificate and key.

I also think nobody wants to deal with certificate management manually since LetsEncrypt and the ACME protocol made that a thing of the past. So while suppling paths to certificate files would work fine, I would also like to explore the options to make TLS usage less of a pain. I also realize that using an ACME client library to make it easy to get certificates might fall out of the scope of the initial work we want to do, I think it's good to consider a path forward here before making changes.

My first thought is around how an implementation using something like [autocert](https://pkg.go.dev/golang.org/x/crypto/acme/autocert) would be like. Where are the integration points with Fuego?

What do you think?

dylanhitt:
I'm pretty against having fuego do any automated certificate stuff. That's on the caller in my opinion. 

But that's just my opinion.

redecs:
@dylanhitt I agree, also don't think Fuego should manage certificates automatically. What I meant is that we should look at how some of the libraries that do that work to make sure the API makes the integration easier.

Here an example: let's say I want to use TLS with certs from LetsEncrypt, I will setup `autocert` in my main.go and then I'll just hook it into Feugo somehow. Managing the certs (storage, renewal, etc) would my programs responsibility, not the framework's. I will try to come with some code to make it clear.

dylanhitt:
Ah I see, yes, some thought into how we expose the server TLS config could be fruitful. It is annoying when you need to do anything a bit more low level.

dylanhitt:
Hey @redecs curious if you were gonna work on this?

Thanks

redecs:
@dylanhitt Yes, I am. Will try to open a PR later today to get some feedback.

dylanhitt:
closed by #120

PR #120, opened 2024-04-16
Easier TLS setup

redecs:
Work to address the easier TLS setup discussed here: https://github.com/go-fuego/fuego/issues/113.

The goals would be to:

- [x] run Fuego in TLS mode
- [x] ~~support custom TLS configuration to have easy option to hook TLS library (for ACME protocol TLS managers that can handle everything automatically)~~ not required not, since you can access Fuego's underling `http.Server` to [set the `tls.Config` directly](https://github.com/go-fuego/fuego/pull/120/files#diff-5c0f29959b10e9d30e8a9253023cb65131d9844ed82b9c0c6bc41e3b8ff21fc0R73).


serve.go
```diff
@@ -9,25 +9,37 @@ import (
...
-// It is blocking.
-// It returns an error if the server could not start (it could not bind to the port for example).
-// It also generates the OpenAPI spec and outputs it to a file, the UI, and a handler (if enabled).
-func (s *Server) Run() error {
+func (s *Server) setupRun(proto string) {
 	go s.OutputOpenAPISpec()
 	elapsed := time.Since(s.startTime)
 	slog.Debug("Server started in "+elapsed.String(), "info", "time between since server creation (fuego.NewServer) and server startup (fuego.Run). Depending on your implementation, there might be things that do not depend on fuego slowing start time")
-	slog.Info("Server running ✅ on http://"+s.Server.Addr, "started in", elapsed.String())
+	slog.Info("Server running ✅ on "+proto+"://"+s.Server.Addr, "started in", elapsed.String())
```
redecs, line 16: Most of the time I just bind my servers to all interfaces using the shorthand syntax `:8080` so this log message becomes 
> Server running ✅ on http://:8080 
witch bugs me a bit. 
I also noticed Fuego deprecated the `WithPort` option in favor of the `WithAddress` one, so maybe we want to  update this log message to better reflect that?

redecs:
I made this a draft until I get a better idea on how the integration with a TLS library would look like.
I're looked at [https://pkg.go.dev/golang.org/x/crypto/acme/autocert](autocert) but I found it lacking some features.
The next place I looked is Caddy, and I think [https://github.com/caddyserver/certmagic](certmagic) it's a better choice for the average developer that would want TLS (and HTTP/2) with Fuego, while of course keeping any of the functionality out of the Fuego code. 
Right now I think both library would work by hooking into the `fuego.Server.TLSConfig` and I thinking to add a `WithTLSConfig` option, but I want to build an examples and figure out some details. I'll try go get this done in the following days (I need to get a VPS for testing).

serve.go
dylanhitt: What do you have in mind?

dylanhitt:
> I made this a draft until I get a better idea on how the integration with a TLS library would look like.

What you've implemented here is fine for the scope of the issue and you shouldn't worry about the other integration in this Pull Request as I'm sure the next PR will have a lot more discussion 😉

serve.go
dylanhitt: Also when it comes to switching the proto type can we use a boolean flag as a the input to setupRun so like

```golang
proto := protoHTTP
if isTLS {
	proto = protoHTTPS
}
```

This may seem like overkill at the moment but, once you expose the TLS configuration option in the future our users will be calling `s.Run()`. So we'd need to interpret if the user is starting the server with TLS by inspecting the configuration in `Server` so I'd expect to see code like this with `Run()` eventually

```golang
func (s *Server) Run() error {
	s.setupRun(s.TLSConfig != nil)
	return s.Server.ListenAndServe()
}
```

dylanhitt (changes requested):

serve.go
redecs: OK, this makes sense to me too. But pushing this a step forward if we have a TLSConfig (or a key/cert) we could also have an if inside `Run` for `ListenAndServe` or `ListenAndServeWithTLS`.

I think our initial thoughts around this were to mirror the `net/http` approach but since Fuego's function si called `Run` (no implied connection on TLS or not) I think it will a better developer experience to use options for TLSConfig and cert/key configuration and determine what mode we need to run in based on the options set. I think this would be inline with Fuegos API design (similar pattern to CORS MIddleware but for TLS).  What do you think?

redecs:
I've added the custom TLSConfig option and removed the `RunTLS` function I initially added because to me that simplifies the Fuego API for the developers that are using it. We can easily determine if the server should run in TLS mode based on the configured options (via TLSConfig or the provided TLS files). If you don't agree, I think we can roll this back quite easily.

I would like to also contribute on example of TLS server setup to show people how this can be leveraged better but I still have some bits to figure out here.

serve.go
redecs: > What do you have in mind?

Regarding logging: now that've seen the `disableStartupMessages` option, I think we can leave this as is. If people set the `Addr` up via an environment variable or other config to be something like `localhost:8080` and they have this enabled they can CTRL/CMD+Click the link to open it from a terminal. This might help newcomers and for people that set the listener address to `:8080` but don't have startup messages disabled they can just ignore that.

serve_test.go
```diff
@@ -188,6 +189,46 @@ func TestServer_Run(t *testing.T) {
...
+func TestServer_RunTLS(t *testing.T) {
+	// This is not a standard test, it is here to ensure that the server can run.
+	// Please do not run this kind of test for your controllers, it is NOT unit testing.
+	tt := []struct {
+		name string
+		opts []func(*Server)
+	}{
+		{
+			name: "can run TLS server with TLS config",
+			opts: []func(*Server){WithTLSConfig(&tls.Config{}), WithoutLogger()},
```
dylanhitt, line 201: This fails to start up the server. 

If you remove the go func and check for the err you can see it fail with. 

```
open : no such file or directory
```

As in this case no file is given: `TLSConfig.Certificates` or `TLSConfig.GetCertificate` has to be populated

serve_test.go
```diff
@@ -188,6 +189,46 @@ func TestServer_Run(t *testing.T) {
...
+		},
+		{
+			name: "can run TLS server with TLS files",
```
dylanhitt, line 204: This case also never runs ListenAndServerTLS as you're checking for `!= ""` in `isTLS`

serve.go
```diff
@@ -22,15 +22,21 @@ func (s *Server) Run() error {
 	if s.corsMiddleware != nil {
 		s.Server.Handler = s.corsMiddleware(s.Server.Handler)
 	}
-
+	if s.isTLS() {
```
dylanhitt, line 25: The issue with this approach is that it's very easy for the user to misconfigure something. We should provide a simple way of starting the server that is similar to the stdlib.

For users that need more the WithTLS config is fine, but more work is needed to support that a bit better (hence why it should be it's own PR). With the RunTLS approach we cover a large portion of use cases. For users that need to get into the nuts and bolts of a TLSConfig they can still do that as `Server` is exposed. And they can simply call `app.RunTLS("","")`.

dylanhitt (changes requested):

serve_test.go
redecs: Thanks, I'll try to improve the tests.

serve_test.go
redecs: Didn't see the forest from the trees. Sorry, I'll fix ths.

redecs: I might have exaggerated with a full blown functional test for the TLS server, but since I messed up on my first attempt I wanted to stop making any assumptions and make sure everything actually works.
Looking forward to any feedback that would help simplifying the TLS server tests.

serve.go
redecs: Now that I have real tests to check all of this I understand what you mean. Adding some errors in `Run` to account for invalid configuration didn't seem like a big deal but with certificates being provided by `tls.Config.Certificates` or `tls.Config.GetCertificate` or files for ListenAndRunTLS is not as simple as I initially though. 

I'm afraid that once we added `RunTLS` that would be in there forever, I would really like to have that magic Run option but for now I've followed your wise advice and simplified things.

redecs:
Added `RunTLS` back in and removed the options I've previously added. Proper testing for TLS Server. I think we need to updated the test for `Run` to use a real http client if we want to really test the `Run` method, not just `s.Mux`, but I haven't made that change yet. I did add graceful shutdown for both HTTP and HTTPS servers to make sure there are not left overs between test (I had issues when I was running the full test suite).

redecs:
Also pushed my working example for TLS with ACME Server (Let's Encrypt Staging) using certmagic. Let me know if think it's useful otherwise I can remove it.

redecs:
@dylanhitt removed the example, will open a new PR with that after we are done with this one, since it requires the code.

dylanhitt (approved):

PR #122, opened 2024-04-16
feat: add ability to disable startup message print

dylanhitt:
This is just noise in some cases. 

Should probably provide the ability to disable the openapi printing as well i.e

```
2024/04/16 17:01:30 INFO JSON spec: http://0.0.0.0:9999/swagger/openapi.json
2024/04/16 17:01:30 INFO OpenAPI UI: http://0.0.0.0:9999/swagger/index.html
2024/04/16 17:01:30 INFO JSON file: doc/openapi.json
```

The only thing I'm really debating about though is should we just always print the debug message here? Could be useful.

### DIFF
--- a/options.go
+++ b/options.go
@@ -53,8 +53,9 @@ type Server struct {
 
 	middlewares []func(http.Handler) http.Handler
 
-	disableAutoGroupTags bool
-	groupTag             string
+	disableStartupMessage bool
+	disableAutoGroupTags  bool
+	groupTag              string
 
 	basePath string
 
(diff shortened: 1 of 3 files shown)


EwenQuim:
I think this option should also disable the OpenAPI message as well.

That's debatable though, but they are noise on the same level as the startup message : useful to start but maybe not necessary to develop a full custom app, and definitely replaceable.

dylanhitt:
If that's the case give me like 10 more minutes.

EwenQuim:
Closed in favor of #123

PR #123, opened 2024-04-16
Feat/disable print openapi

dylanhitt:
Same ad #122 but extends the options to also disable the printing startup of openapi resources.


EwenQuim:
Nice work! 

Thank you 🎉

EwenQuim (approved):

PR #125, opened 2024-04-19
ACME Server TLS example with Let's Encrypt and certmagic

redecs:
Full example for a [TLS](https://github.com/go-fuego/fuego/pull/120) setup with Fuego. This example should get you very close to a production setup with HTTPS (and HTTP/2), automatic TLS certificates (with auto-renew) and HTTP to HTTPS redirect.

Key points:
- Used [certmagic](https://github.com/caddyserver/certmagic) for ACME Server integration: Let's Encrypt Staging
- Runs also an HTTP server (using net/http) to solve the HTTP challenge for the ACME server 
- Set's a custom `tls.Config` from `certmagic` on the Fuego's underlying `http.Server`
- Graceful shutdown for both HTTP and HTTPS servers

This is examples is meant to be run on server that has a real DNS record pointing to it, with ports 80 and 443 opened. You need to provide the domain name and email address using flags.

While initially [I was considering autocert](https://github.com/go-fuego/fuego/issues/113) I found to `certmagic` from Caddy to be more powerful.


EwenQuim (approved):

EwenQuim:
Thanks! Could you remove the `chi` import by running `go mod tidy` ? I think it isn't useful here

EwenQuim:
Thank you very much! 🔥

PR #126, opened 2024-04-20
Chore/server setup

dylanhitt:
- Refactors the server start ups to use the same setup function
- Adds test coverage on the with middleware during server startup. Is it worth maybe sending invalid request to the server and see it fails? 🤷 feels like we're just testing the stdlib at that point.
- also found some doc oopsies around cor middleware.

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -14,15 +14,7 @@ import (
 // It returns an error if the server could not start (it could not bind to the port for example).
 // It also generates the OpenAPI spec and outputs it to a file, the UI, and a handler (if enabled).
 func (s *Server) Run() error {
-	go s.OutputOpenAPISpec()
-
-	s.printStartupMessage()
-
-	s.Server.Handler = s.Mux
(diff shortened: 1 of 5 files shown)


serve.go
```diff
@@ -32,28 +24,35 @@ func (s *Server) Run() error {
...
+		slog.Info("Server running ✅ on "+s.proto()+"://"+s.Server.Addr, "started in", elapsed.String())
+	}
+}
+
+func (s *Server) proto() string {
+	proto := "http"
+	if s.isTLS {
+		return "https"
 	}
+	return proto
```
EwenQuim, line 55:
```go
if s.isTLS {
    return "https"
}

return "http"
```

EwenQuim (commented):
Nice ! Thanks :)

serve.go
dylanhitt: Ah thanks. I new that looked weird.

EwenQuim (approved):

PR #132, opened 2024-06-24
WIP: fix: differentiate between array typres and standard structs

### LINKED ISSUE #130: OpenAPI Spec Generation Doesn't Differentiate Between Single Struct and Slice
**To Reproduce**

I have two handlers:
fuego.Get(group, "/{id}", rh.GetStat)
fuego.Get(group, "/", rh.ListStats)

func (rh RouteHandler) ListStats(c *fuego.ContextNoBody) ([]models.Stat, error) {
	return rh.Service.ListStats()
}

func (rh RouteHandler) GetStat(c *fuego.ContextNoBody) (models.Stat, error) {
	id := c.PathParam("id")
	return rh.Service.GetStat(id)
}

When the openapi spec is generated, the schema doesn't differentiate between passing a single struct versus a slice, and it only creates a schema for the first instance of the struct registered. So in the example above, it generates the following schema. (if the list function was first it would make it an array type)

    Stat:
      properties:
        Description:
          type: string
        ID:
          type: string
        Query:
          type: string
        Result:
          type: integer
      type: object

**Expected behavior**
I would expect it generate two separate schemas, one as shown above, and another one that is an array of the Stats schema.

**Framework version (please check if it happens with the last Fuego version before posting):**
0.13.4

1.22.0

dylanhitt:
Work in progress and I'm not sure if it works for array of array. To be honest I'm not sure what an openapi spec is supposed to look like that in that situation so I will need to do some research. 

close #130

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -164,25 +164,17 @@ func RegisterOpenAPIOperation[T, B any](s *Server, method, path string) (*openap
 	bodyTag := tagFromType(*new(B))
 	if (method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch) && bodyTag != "unknown-interface" && bodyTag != "string" {
 
-		bodySchema, ok := s.OpenApiSpec.Components.Schemas[bodyTag]
-		if !ok {
-			var err error
-			bodySchema, err = generator.NewSchemaRefForValue(new(B), s.OpenApiSpec.Components.Schemas)
-			if err != nil {
-				return operation, err
-			}
-			s.OpenApiSpec.Components.Schemas[bodyTag] = bodySchema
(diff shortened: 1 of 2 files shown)


dylanhitt:
Actually looking at it. This for sure will not handle the input of `[][]string` properly

dylanhitt:
Before I forget. The schema type needs to be built in something similar to our tag type.

dylanhitt:
This actually turned out to be more straightforward than I thought.... Take a look #133  to show the bug. 

I'm gonna refactor this down now as well as add some more robust tests.

dylanhitt:
Basically what was happening was the first component in would win since we were only generating once. I see this was probably done for some optimization purposes, but it appears the lib handles that pretty well already.

PR #134, opened 2024-06-26
fix: differentiate between array types and standard structs

dylanhitt:
This did turn out to be pretty complex. 

The following is by no means finished. More looking for a discussion. If you take a look #133 the bug is fairly clear. We do not respect response nor return types when they are arrays. 

The only way I can think of solving this is that we must create the ref to the schema and link it at the bottom of a recursion tree. If you do not do this you will end up with improper refs, with the original approach the input for the first type wins so if an array of `MyStruct` is passed the schema will be a list of `MyStruct` rather than just the simple Schema of `MyStruct`. See `schemaDive`. Currently this does require us to dive twice on request bodies. Since we seem to check request bodies if they are `unknown-type`/`string`, which I am unsure why (wouldn't we want to have []string be a valid requestBody)? As long as that check exists I'm not sure how to avoid both recursive steps. 

This does however give the proper diff
image

This does need more tests, but I'd like another pair of eyes too look at the approach @EwenQuim if you don't mind. Could you please take a look. I'm sure this can be simplified. I've been staring at it too long at this point. 

@epentland feel free to chime since you brought this to our attention

Thank you

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -164,25 +164,17 @@ func RegisterOpenAPIOperation[T, B any](s *Server, method, path string) (*openap
 	bodyTag := tagFromType(*new(B))
 	if (method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch) && bodyTag != "unknown-interface" && bodyTag != "string" {
 
-		bodySchema, ok := s.OpenApiSpec.Components.Schemas[bodyTag]
-		if !ok {
-			var err error
-			bodySchema, err = generator.NewSchemaRefForValue(new(B), s.OpenApiSpec.Components.Schemas)
-			if err != nil {
-				return operation, err
-			}
-			s.OpenApiSpec.Components.Schemas[bodyTag] = bodySchema
+		bodySchema, err := generator.NewSchemaRefForValue(new(B), s.OpenApiSpec.Components.Schemas)
+		if err != nil {
+			return operation, err
 		}
 
+		content := openapi3.NewContentWithSchema(bodySchema.Value, []string{"application/json"})
+		content["application/json"].Schema.Ref = "#/components/schemas/" + bodyTag
 		requestBody := openapi3.NewRequestBody().
 			WithRequired(true).
-			WithDescription("Request body for " + reflect.TypeOf(*new(B)).String())
-
-		if bodySchema != nil {
-			content := openapi3.NewContentWithSchema(bodySchema.Value, []string{"application/json"})
-			content["application/json"].Schema.Ref = "#/components/schemas/" + bodyTag
-			requestBody.WithContent(content)
-		}
+			WithDescription("Request body for " + reflect.TypeOf(*new(B)).String()).
+			WithContent(content)
 
 		s.OpenApiSpec.Components.RequestBodies[bodyTag] = &openapi3.RequestBodyRef{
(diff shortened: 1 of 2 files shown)


dylanhitt:
Was able to drop the need to recurse twice on request bodies by passing a the `tag` back with a struct of string and schemaRef

openapi_test.go
```diff
@@ -67,6 +161,8 @@ func TestServer_generateOpenAPI(t *testing.T) {
 	require.NotNil(t, document.Paths.Find("/"))
 	require.Nil(t, document.Paths.Find("/unknown"))
 	require.NotNil(t, document.Paths.Find("/post"))
+	require.Equal(t, document.Paths.Find("/post").Post.Responses.Value("200").Value.Content["application/json"].Schema.Value.Type, "array")
+	require.Equal(t, document.Paths.Find("/post").Post.Responses.Value("200").Value.Content["application/json"].Schema.Value.Items.Ref, "#/components/schemas/MyStruct")
```
dylanhitt, line 165: Need to do more to validate the actual schema reference here

openapi.go
```diff
@@ -227,25 +206,69 @@ func RegisterOpenAPIOperation[T, B any](s *Server, method, path string) (*openap
...
+		tag.Value, err = s.getOrCreateSchema(tag.name, new(V))
+		return tag, err
+	}
+}
+
+func (s *Server) getOrCreateSchema(key string, v any) (*openapi3.Schema, error) {
+	var err error
+	schemaRef, ok := s.OpenApiSpec.Components.Schemas[key]
+	if !ok {
+		schemaRef, err = generator.NewSchemaRefForValue(v, s.OpenApiSpec.Components.Schemas)
```
dylanhitt, line 267: need test for this case

dylanhitt: It is not possible for this function error unless a custom schema customizer is provided or `ThrowErrorOnCycle` is set.

dylanhitt:
@EwenQuim when you have a moment to do you mind taking a look.

EwenQuim:
I'll look at it omorrow morning, thanks for this investigation.

It is indeed complex.

dylanhitt:
Sorry for the spam 😓. Messed some things up while getting my fork ready for a build.

dylanhitt:
@EwenQuim any word on this?

EwenQuim:
Ready to review it, I'd like to merge #137 just before this one, so I'll be able to work on Multi Return OpenAPI spec

EwenQuim:
Mmmh maybe I'll merge this before #137 finally...

openapi.go
```diff
@@ -227,25 +200,70 @@ func RegisterOpenAPIOperation[T, B any](s *Server, method, path string) (*openap
...
+
+	case reflect.Slice, reflect.Array:
+		item := dive(s, t.Elem(), tag, maxDepth-1)
+		tag.name = item.name
+		tag.Value = &openapi3.Schema{
+			Type:  "array",
+			Items: &item.SchemaRef,
 		}
-		return dive(t.Elem(), maxDepth-1)
+		return tag
```
EwenQuim, line 252: Perfect, you really dived into the code! 😉

EwenQuim (approved):

PR #135, opened 2024-07-01
fix: revert back to the simple funcName style operationID

dylanhitt:
When using the FullName you end up with some pretty gnarly function name when generating clientSDKs 

```
type PATCHv2devicesNamedeallocateGitlabComdylanhittmyprojpkgapiserverv2handlerHandlerRequesBody
```

I realized it's possible to override this but our default should be a bit more organized.

Let me know what you think.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -123,9 +123,9 @@ func Register[T, B any](s *Server, route Route[T, B], controller http.Handler, m
 		route.FullName = route.Path
 	}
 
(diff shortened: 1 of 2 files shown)


EwenQuim:
Good idea. Can you add some tests please? So users will have some ideas of what they can expect to do.

Thanks again!

dylanhitt:
There,

Let me know if you need anything else. 

Cheers

EwenQuim (approved):

PR #137, opened 2024-07-02
Multi return

EwenQuim:
## Spec

Currently, return type is either JSON or XML for the whole app. This needs to be improved.

The route /recipes should send HTML with a list of recipes if the `Accept` Header is "text/html", or JSON if the `Accept` Header is "application/json", or XML if the `Accept` Header is "application/xml". This allows to build a traditional website in the same time as building an API.

This change is of course backward-compatible and existing controller will not change current behaviour. Performance should also not be degraded.

## Example

A controller might look like this (see `multi_return_test.go` for more examples).

```go
fuego.Get(s, "/data", func(c ContextNoBody) (DataOrTemplate[MyType], error) {
	entity := MyType{Name: "Ewen"}

	return DataOrTemplate[MyType]{
		Data: entity, // Used when asking for JSON or XML
		Template: MyTemplate(entity) // std template, Templ or Gomponent rendered when asking for HTML
	}, nil
})
```


multi_return.go
```diff
@@ -0,0 +1,47 @@
...
+// depending on the asked type.
+type DataOrTemplate[T any] struct {
+	Data     T
+	Template any
+}
+
+var (
+	_ CtxRenderer    = DataOrTemplate[any]{} // Can render HTML
+	_ json.Marshaler = DataOrTemplate[any]{} // Can render JSON
+	_ xml.Marshaler  = DataOrTemplate[any]{} // Can render XML
```
dylanhitt, line 20: No Yaml?

EwenQuim: I'll add it!

EwenQuim:
This is a lot cleaner and I think it is ready to merge. Examples of the new features in `_test.go` files.

Another thing is missing : when using this new `DataOrTemplate` structure, the generated OpenAPI spec is a bit buggy. But I'll fix this after reviewing PR #134, since it is very deeply linked. What do you think about this @dylanhitt ?

dylanhitt:
Will look at it at the end of the week. :smile:

EwenQuim:
Ready for review ✅ 

The generated OpenAPI schema should be correct.

When using `DataOrTemplate[MyType] `, OpenAPI spec will only see `MyType` as it is the type return by JSON/XML APIs, the template is ignored.

This allow to work on classic "view" (from MVC pattern) routes and create an API at the same time. I was playing with this feature the last few weeks and it is very helpful.c

dylanhitt:
Will be looking at this Sunday night. Day job got in the way this week

serialization.go
```diff
@@ -71,13 +75,43 @@ func transformOut[T any](ctx context.Context, ans T) (T, error) {
...
+	switch parseAcceptHeader(r.Header.Get("Accept"), ans) {
+	case "application/xml":
+		return SendXML(w, r, ans)
+	case "text/html":
+		return SendHTML(r.Context(), w, ans)
+	case "text/plain":
+		return SendText(w, ans)
+	case "application/json":
+		SendJSON(w, ans)
+	case "application/yaml":
```
dylanhitt, line 90: We'd want `case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html` here as well right?

serialization.go
```diff
@@ -88,6 +122,28 @@ func SendJSON(w http.ResponseWriter, ans any) {
...
+	switch accept {
+	case "application/xml":
+		SendXMLError(w, r, err)
+	case "text/html":
+		_ = SendHTMLError(r.Context(), w, err)
+	case "text/plain":
+		_ = SendText(w, err)
+	case "application/json":
+		SendJSONError(w, err)
+	default:
```
dylanhitt, line 142: So in the case of yaml we'll just want to respond with json?

dylanhitt (changes requested):
This is super clean and very clear. Just a few comments to address.

serialization.go
```diff
@@ -88,6 +122,28 @@ func SendJSON(w http.ResponseWriter, ans any) {
 	}
 }
 
+// SendError sends an error.
+// Declared as a variable to be able to override it for clients that need to customize serialization.
+var SendError = func(w http.ResponseWriter, r *http.Request, err error) {
+	accept := parseAcceptHeader(r.Header.Get("Accept"), nil)
+	if accept == "" {
```
dylanhitt, line 129: I believe this is redundant? You'll just fall through and the default condition if it's empty

serialization.go
EwenQuim: Good point, I'm adding it!

serialization.go
EwenQuim: Yes exact, Removing this :)

serialization.go
EwenQuim: Adding yaml here too

PR #138, opened 2024-07-03
Ci/markdown lint

dylanhitt:
Closes #121

Let me know if anyone has opinions on the rules.

Thanks

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,9 @@
+MD001: false
+
+MD010: false
+
+MD013:
+  code_blocks: false
+  tables: false
+
+MD033: false
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,6 +3,7 @@ name: Lint
 on:
   pull_request:
   push:
+    branches: ["main"]
 
 jobs:
   markdownlint:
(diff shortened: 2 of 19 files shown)


EwenQuim (approved):

.markdownlint.yaml
```diff
@@ -0,0 +1,9 @@
+MD001: false
+
+MD010: false
+
+MD013:
+  code_blocks: false
+  tables: false
+
+MD033: false
```
ccoVeille, line 9: These disabled rules might sound opinionated, mainly because they come without a comment about what they are about and why they are deactivated.

I know markdown lint, and I'm OK with the rules you chose, but I think comments would help

dylanhitt: Good point, we definitely should add a summary about what they are disabling for each one at the very least. 

Feel free to do it if you like, if not I’ll slide into my day Monday

Cheers

ccoVeille: I could, but I'm not in your head and the reason why you removed might differ from what I think you removed them 😅

ccoVeille: I opened #143

dylanhitt: Awesome thanks, I’ll add you to review Monday.

PR #139, opened 2024-07-03
deps: bump gorilla/scheme to v1.4.1

dylanhitt:


go.mod
```diff
@@ -7,8 +7,9 @@ require (
 	github.com/go-playground/validator/v10 v10.19.0
 	github.com/golang-jwt/jwt/v5 v5.2.1
 	github.com/gomarkdown/markdown v0.0.0-20231222211730-1d6d20845b47
-	github.com/gorilla/schema v1.2.1
+	github.com/gorilla/schema v1.4.1
 	github.com/hashicorp/golang-lru/v2 v2.0.7
+	github.com/rs/cors v1.11.0
```
EwenQuim, line 12: I don't like having this in go.mod, it make it looks like Fuego needs this packagez but it's only used in tests

dylanhitt: I’ll take a look. But all I did was bump scheme and go mod tidy. Maybe I fat fingered something.

dylanhitt: Okay just looked at schema. Not sure how this is needed at all.

dylanhitt: Should be fine now. Thank you for the catch 😄

PR #144, opened 2024-07-07
chore: add justification for disabled markdownlint rules

### LINKED ISSUE #143: Add comments in .markdownlint.yml
These disabled rules might sound opinionated, mainly because they come without a comment about what they are about and why they are deactivated.

I know markdown lint, and I'm OK with the rules you chose, but I think comments would help

_Originally posted by @ccoVeille in https://github.com/go-fuego/fuego/pull/138#discussion_r1667323081_

dylanhitt:
Closes #143

@ccoVeille thank you for the suggestions. The practice actually uncovered that the CI wasn't actually testing anything 😅. I appreciate your time here.

Let me know how this looks!

Cheers

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -9,6 +9,8 @@ jobs:
   markdownlint:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
+
       - name: markdownlint-cli
         uses: nosborn/github-action-markdown-cli@v3.3.0
         with:
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,9 +1,22 @@
-MD001: false
-
+# MD010/no-hard-tabs : Hard tabs : https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md010.md
+#
+# The rendered markdown is viewed in github and
+# the Docusaurus site which render hard-tabs fine.
+# It is annoying to have to deal with so many spaces
+# specifically in golang code blocks.
 MD010: false
 
+# MD013/line-length : Line length : https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md013.md
 MD013:
+  # Applying line length to code blocks
+  # makes the code look worse sometimes.
   code_blocks: false
+  # Applying line length to tables makes
+  # it more difficult to edit.
   tables: false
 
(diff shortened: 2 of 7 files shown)


ccoVeille (approved):

.github/workflows/lint.yaml
```diff
@@ -9,6 +9,8 @@ jobs:
   markdownlint:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
+
```
EwenQuim, line 13: Haha classic mistake :)

.markdownlint.yaml
```diff
@@ -1,9 +1,22 @@
...
 MD010: false
 
+# MD013/line-length : Line length : https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md013.md
 MD013:
+  # Applying line length to code blocks
+  # makes the code look worse sometimes.
   code_blocks: false
+  # Applying line length to tables makes
+  # it more difficult to edit.
   tables: false
```
EwenQuim, line 16: Is it really necessary? Why would you want that?

.markdownlint.yaml
```diff
@@ -1,9 +1,22 @@
...
 
+# MD033/no-inline-html : Inline HTML : https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md033.md
+#
+# It's more convenient to work with HTML in some cases.
+# We use HTML in a lot of our docs.
 MD033: false
```
EwenQuim, line 22: Nice. Thanks for adding doc on top of each rule!

.markdownlint.yaml
dylanhitt: not following. Both these lines are disabling line length checks on `code_blocks` and `tables`

I guess this show I should change my rhetoric too `Disabling line length check on code blocks. Applying against code blocks makes the code look work sometimes`?

EwenQuim: Sorry my comment was unclear.
I would disable line length checker. I don't care if my markdown line is 123 character long. I don't see how it would improve lisibility.

dylanhitt: It doesn't have to do with the rendering of the markdown, more so the developer experience. I use a basic text editor. Viewing long lines is a pain. If you're against I can disable, but it IMO it makes for a much better experience when editing markdown.

EwenQuim: If it helps you, that's OK for me :)

EwenQuim (approved):

PR #148, opened 2024-07-22
Golden file testing

EwenQuim:
Adds golden file testing, and vacuum for schema validation.

This helps tracks the generated OpenAPI schema when making change on the generation


mux.go
```diff
@@ -125,7 +125,7 @@ func Register[T, B any](s *Server, route Route[T, B], controller http.Handler, m
 
 	route.Operation.Summary = route.NameFromNamespace(camelToHuman)
 	route.Operation.Description = "controller: `" + route.FullName + "`\n\n---\n\n"
-	route.Operation.OperationID = route.Method + " " + s.basePath + route.Path + ":" + route.NameFromNamespace()
+	route.Operation.OperationID = route.Method + "_" + s.basePath + strings.ReplaceAll(strings.ReplaceAll(route.Path, "{", ":"), "}", "")
```
EwenQuim, line 128: This is a breaking change fix, will notify it on next release!

openapi.go
```diff
@@ -162,8 +163,8 @@ func RegisterOpenAPIOperation[T, B any](s *Server, method, path string) (*openap
 
 	// Request body
 	bodyTag := schemaTagFromType(s, *new(B))
-	if (method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch) && bodyTag.name != "unknown-interface" && bodyTag.name != "string" {
-		content := openapi3.NewContentWithSchemaRef(&bodyTag.SchemaRef, []string{"application/json"})
+	if bodyTag.name != "unknown-interface" {
+		content := openapi3.NewContentWithSchemaRef(&bodyTag.SchemaRef, []string{"application/json", "application/xml"})
```
EwenQuim, line 167: Thanks to #137 we will have this behaviour accepted as well 🥳

openapi.go
```diff
@@ -192,6 +193,10 @@ func RegisterOpenAPIOperation[T, B any](s *Server, method, path string) (*openap
 	for _, pathParam := range parsePathParams(path) {
 		parameter := openapi3.NewPathParameter(pathParam)
 		parameter.Schema = openapi3.NewStringSchema().NewRef()
+		if strings.HasSuffix(pathParam, "...") {
```
EwenQuim, line 196: Supports `{things...}` syntax https://pkg.go.dev/net/http#ServeMux

mux.go
dylanhitt: Much better 👍

dylanhitt (approved):

dylanhitt:
My only comment is we should probably add a check in the CI to ensure the golden-file is being updated. The CI check shouldn't update the file, but it would just fail until the author of the MR updated it.

EwenQuim:
If the golden file isn't updated and it should, the tests will fail, so no issue here :)

PR #150, opened 2024-07-22
Errors can be declared globally and route per route

### LINKED ISSUE #127: openapi spec for non 200 responses
**To Reproduce**
Not sure if this is a bug, or just not implemented yet but I noticed that when the openapi spec is generated there is no error responses. It just generated 200's responses, but in the case we want to have non 200 responses included in the openapi spec how would this be possible? Is there documentation how to generate error responses for handlers?

**Expected behavior**
For example I have this 

```
"/login": {
      "post": {
        "description": "Authenticates User",
        "operationId": "POST /login:Handler",
        "requestBody": {
          "$ref": "#/components/requestBodies/LoginReq"
        },
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/LoginResp"
                }
              }
            },
            "description": "OK"
          },
          "default": {
            "description": ""
          }
        },
        "summary": "Login",
        "tags": [
          "Login"
        ]
      }
    }
```

I would want to add to the responses spec
```
"responses": {
          "200": {
            ......
            },
           "401": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/BadResp"
                }
              }
            },
          },
```

**Framework version (please check if it happens with the last Fuego version before posting):**
github.com/go-fuego/fuego v0.13.1

1.22.1

EwenQuim:
Fixes #127 

Please see the golden file for OpenAPI changes :)


examples/full-app-gourmet/server/server.go
```diff
@@ -23,13 +23,16 @@ func (rs Ressources) Setup(
 		fuego.WithAutoAuth(controller.LoginFunc),
 		fuego.WithTemplateFS(templates.FS),
 		fuego.WithTemplateGlobs("**/*.html", "**/**/*.html"),
+		fuego.WithGlobalResponseTypes(403, "Forbidden"),
```
ccoVeille, line 26:
```suggestion
		fuego.WithGlobalResponseTypes(http.StatusForbidden, "Forbidden"),
```

examples/full-app-gourmet/views/views.go
```diff
@@ -24,7 +24,7 @@ func (rs Ressource) Routes(s *fuego.Server) {
 
 	// Public Chunks
 	fuego.Get(s, "/recipes-list", rs.showRecipesList)
-	fuego.Get(s, "/search", rs.searchRecipes)
+	fuego.Get(s, "/search", rs.searchRecipes).AddError(401, "Authorization Error").AddError(500, "My Server Error")
```
ccoVeille, line 27:
```suggestion
	fuego.Get(s, "/search", rs.searchRecipes).AddError(http.StatusUnauthorized, "Authorization Error").AddError(http.StatusInternalError, "My Server Error")
```

options.go
```diff
@@ -113,6 +115,8 @@ func NewServer(options ...func(*Server)) *Server {
 		WithSerializer(Send),
 		WithErrorSerializer(SendError),
 		WithErrorHandler(ErrorHandler),
+		WithGlobalResponseTypes(400, "Bad Request _(validation or deserialization error)_", HTTPError{}),
```
ccoVeille, line 118:
```suggestion
		WithGlobalResponseTypes(http.StatusBadRequest, "Bad Request _(validation or deserialization error)_", HTTPError{}),
```

dylanhitt:
Per @ccoVeille suggestions why not just add:

https://golangci-lint.run/usage/linters/#usestdlibvars

to our golangci-lint config?

openapi_operations_test.go
```diff
@@ -61,6 +61,17 @@ func TestHeaderParams(t *testing.T) {
...
+		Message string
+	}
+	s := NewServer()
+	route := Get(s, "/test", testController).
+		AddError(400, "My Validation Error", MyError{})
+
+	require.Equal(t, "My Validation Error", *route.Operation.Responses.Map()["400"].Value.Description)
+}
+
 func TestCookieParams(t *testing.T) {
```
dylanhitt, line 75: Mind adding a test showing the behavior when Errors are set globally and on a route? Also display what should happen when status code is declared globally but then overridden by the route. I think this would be good too make sure we don't make a breaking change in the future

openapi_operations.go
```diff
@@ -82,6 +82,33 @@ func (r Route[ResponseBody, RequestBody]) AddTags(tags ...string) Route[Response
...
+func (r Route[ResponseBody, RequestBody]) AddError(code int, description string, errorType ...any) Route[ResponseBody, RequestBody] {
+	addResponse(r.mainRouter, r.Operation, code, description, errorType...)
+	return r
+}
+
+func addResponse(s *Server, operation *openapi3.Operation, code int, description string, errorType ...any) {
+	responseSchema := schemaTagFromType(s, HTTPError{})
+	if len(errorType) > 0 {
+		responseSchema = schemaTagFromType(s, errorType[0])
+	}
```
dylanhitt, line 95: Small optimization. We can either use an `if...else` or wrap this in a function to avoid the `if...else`, would avoid a call to `schemaTagFromType`. Anywho just a nitpick - this would be pretty quick anyway.

dylanhitt (changes requested):
Small things. Feel free to disregard. 

This is an awesome feature and the interface seems to feel natural

ccoVeille:
> Per @ccoVeille suggestions why not just add:
> 
> https://golangci-lint.run/usage/linters/#usestdlibvars
> 
> to our golangci-lint config?
> 

I'm afraid usestdlibvars won't detect this.

https://golangci-lint.run/usage/linters/#usestdlibvars

Also, there is way to configure it yet to tell him this int in this method is an http.Status

But, adding the linter is indeed a great thing.

openapi_operations_test.go
EwenQuim: Very good comment indeed, the tests should reflect this feature's behaviour and show some examples!

Adding this right now.

openapi_operations.go
EwenQuim: Done! Thank you

options.go
EwenQuim: Fixed!

dylanhitt:
> > Per @ccoVeille suggestions why not just add:
> > https://golangci-lint.run/usage/linters/#usestdlibvars
> > to our golangci-lint config?
> 
> I'm afraid usestdlibvars won't detect this.
> 
> https://golangci-lint.run/usage/linters/#usestdlibvars
> 
> Also, there is way to configure it yet to tell him this int in this method is an http.Status
> 
> But, adding the linter is indeed a great thing.

Oh I see, it only works in the context of the stdlib. It's actually already enabled for this project btw.

ccoVeille:
Exactly, it would have been reported otherwise.

One way to detect them in the context of the methods this project use is to set up forbiddigo rules.

It would work

PR #152, opened 2024-07-23
chore: add test for multidimensional array bodies

dylanhitt:
This is intended to just demonstration for #64. It could be useful getting merged 🤷


EwenQuim:
Thanks again!

PR #155, opened 2024-07-24
feat: add ability to disable just the swagger ui

dylanhitt:
per #154

We should provide a way to disable the UI and also not have the route register. I elected to keep with the `DisableSwagger*` vernacular. Which maybe we want to be more specific in the future, maybe a bit more thought out (my bad 😄). 

Maybe a discussion over the interface is warranted, but for now 🤷 

Cheers


EwenQuim:
Isn't it enough to do this ? 

image

Or do you want to disable the UI without disabling the JSON served?

EwenQuim:
Oops sorry for the force-push, don't worry I'll handle it once we agree on what to do

dylanhitt:
`DsiableSwagger` disables the creation of both the spec.json and the UI.

PR #157, opened 2024-08-16
Uses operation provided in register

### LINKED ISSUE #156: Function 'Register' overwrites the provided Route.Operation
I expected Route.Operation to be applied by 'Register', given its exported visibility scope. Had to look into the implementation to see it is completely ignored.

Use the already existing 'Route.Operation' (when not nil).  This way the default field assignments for 'Route.Operation' (such as 'Summary', 'Description', 'OperationID', or any others to be added in the future) are not forcefully overwritten when the user clearly has its own Operation object already made.

An alternative solution could be to make the 'Route.Operation' field private, if that was the original intention.

**Additional context**
The use case of the preferred solution is that sometimes you know some or all the Operation fields beforehand, and other times you have a dynamic logic that passes the Route struct and conditionally calls the different field setters on it for example:
Route.Summary("mySummary")
Route.Description("myDesc")
Route.OperationID("myID")

I think this proposal would add a new feature without much effort to implement. Let me know if you would like me to do it, and whether my proposal is feasible in your current design/vision.

Thank you for this package @EwenQuim !

EwenQuim:
Fixes #156

It uses the Operation in the Register operation.

See [the tests](https://github.com/go-fuego/fuego/pull/157/files#diff-b39fc442f543ad6df386d0772fdbafafee01095292545626a39251403bec6b5c).


EwenQuim:
@dylanhitt what do you think about it?

README.md
```diff
@@ -56,7 +56,8 @@ to develop APIs and web applications.
   (or not, if you want to do it yourself)
 - **Validation**: Fuego provides a simple and fast validator based on `go-playground/validator`
 - **Transformation**: easily transform your data by implementing the
-  `fuego.InTransform` and `fuego.OutTransform` interfaces - also useful for custom **validation**
+  `fuego.InTransform` and `fuego.OutTransform` interfaces - also useful for
```
dylanhitt, line 59: This should probably be a separate commit at least. What do you think?

dylanhitt (approved):
Looks good!

Feel free to resolve the one comment, if you find it pedantic.

EwenQuim:
> Looks good!
> 
> Feel free to resolve the one comment, if you find it pedantic. 

Not pedantic, but I currently don't have my laptop and it is a minor comment so I'll just merge via the app haha

As always, thanks for your review!

PR #158, opened 2024-08-16
Tests/golden 2d array

dylanhitt:
Attempting to ensure that we don't mess this up again. Not the greatest concept for the example but works. 🤷‍♂️ of course a map would work better here.

If there are other ideas let me know. 

Cheers

### DIFF
--- a/examples/petstore/controllers/pets_test.go
+++ b/examples/petstore/controllers/pets_test.go
@@ -55,6 +55,26 @@ func TestGetPets(t *testing.T) {
 	})
 }
 
+func TestGetAllPestByAge(t *testing.T) {
+	t.Run("can get a pet by id", func(t *testing.T) {
+		s := lib.NewPetStoreServer()
+
+		w := httptest.NewRecorder()
+		r := httptest.NewRequest("POST", "/pets/", strings.NewReader(`{"name": "kitkat"}`))
+		s.Mux.ServeHTTP(w, r)
+		t.Log(w.Body.String())
+		require.Equal(t, 200, w.Code)
+
+		w = httptest.NewRecorder()
+		r = httptest.NewRequest("GET", "/pets/by-age", nil)
+
+		s.Mux.ServeHTTP(w, r)
+
+		t.Log(w.Body.String())
+		require.Equal(t, 200, w.Code)
+	})
+}
+
 func TestGetPetsByName(t *testing.T) {
 	t.Run("can get a pet by name", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


examples/petstore/controllers/pets_test.go
```diff
@@ -55,6 +55,26 @@ func TestGetPets(t *testing.T) {
...
 
+func TestGetAllPestByAge(t *testing.T) {
+	t.Run("can get a pet by id", func(t *testing.T) {
+		s := lib.NewPetStoreServer()
+
+		w := httptest.NewRecorder()
+		r := httptest.NewRequest("POST", "/pets/", strings.NewReader(`{"name": "kitkat"}`))
+		s.Mux.ServeHTTP(w, r)
+		t.Log(w.Body.String())
+		require.Equal(t, 200, w.Code)
```
ccoVeille, line 66:
```suggestion
		require.Equal(t, http.StatusOK, w.Code)
```

examples/petstore/controllers/pets_test.go
```diff
@@ -55,6 +55,26 @@ func TestGetPets(t *testing.T) {
...
+
+		w = httptest.NewRecorder()
+		r = httptest.NewRequest("GET", "/pets/by-age", nil)
+
+		s.Mux.ServeHTTP(w, r)
+
+		t.Log(w.Body.String())
+		require.Equal(t, 200, w.Code)
```
ccoVeille, line 74:
```suggestion
		require.Equal(t, http.StatusOK, w.Code)
```

examples/petstore/services/in_memory_pets_test.go
```diff
@@ -38,7 +41,14 @@ func TestInMemoryPets(t *testing.T) {
...
 		require.NoError(t, err)
-		require.Len(t, pets, 1)
+		require.Len(t, pets, 2)
+	})
+
+	t.Run("can get all pets by age", func(t *testing.T) {
+		pets, err := service.GetAllPetsByAge()
+		require.NoError(t, err)
+		require.Equal(t, "kitkat", pets[1][0].Name)
+		require.Equal(t, "payday", pets[3][0].Name)
```
ccoVeille, line 51: I'm unsure about pets size (4), but this should avoid panic if there is a later refactoring

```suggestion
		require.NoError(t, err)
		require.Len(t, pets, 4)
		require.Equal(t, "kitkat", pets[1][0].Name)
		require.Equal(t, "payday", pets[3][0].Name)
```

PR #164, opened 2024-08-22
BREAKING: all default Send* and Send*Error functions to have the same…

dylanhitt:
Per: https://github.com/go-fuego/fuego/discussions/162#discussioncomment-10367657

Squaring up all these signatures.

### DIFF
--- a/examples/full-app-gourmet/controller/recipe.go
+++ b/examples/full-app-gourmet/controller/recipe.go
@@ -31,11 +31,11 @@ func (rs recipeRessource) MountRoutes(s *fuego.Server) {
 func (rs recipeRessource) getAllRecipesStandardWithHelpers(w http.ResponseWriter, r *http.Request) {
 	recipes, err := rs.RecipeRepository.GetRecipes(r.Context())
 	if err != nil {
-		fuego.SendJSONError(w, err)
+		fuego.SendJSONError(w, r, err)
 		return
 	}
 
-	fuego.SendJSON(w, recipes)
+	fuego.SendJSON(w, r, recipes)
 }
 
 func (rs recipeRessource) getAllRecipes(c fuego.ContextNoBody) ([]store.Recipe, error) {
--- a/options.go
+++ b/options.go
@@ -74,9 +74,9 @@ type Server struct {
 	DisableOpenapi        bool // If true, the routes within the server will not generate an openapi spec.
 	maxBodySize           int64
 
-	Serialize      func(w http.ResponseWriter, r *http.Request, ans any) error // Custom serializer that overrides the default one.
-	SerializeError func(w http.ResponseWriter, r *http.Request, err error)     // Used to serialize the error response. Defaults to [SendError].
-	ErrorHandler   func(err error) error                                       // Used to transform any error into a unified error type structure with status code. Defaults to [ErrorHandler]
+	Serialize      Sender                // Custom serializer that overrides the default one.
+	SerializeError ErrorSender           // Used to serialize the error response. Defaults to [SendError].
+	ErrorHandler   func(err error) error // Used to transform any error into a unified error type structure with status code. Defaults to [ErrorHandler]
 	startTime      time.Time
 
 	OpenAPIConfig OpenAPIConfig
@@ -304,13 +304,15 @@ func WithLogHandler(handler slog.Handler) func(*Server) {
 	}
 }
 
-// WithSerializer sets a custom serializer that overrides the default one.
+// WithSerializer sets a custom serializer of type Sender that overrides the default one.
 // Please send a PR if you think the default serializer should be improved, instead of jumping to this option.
-func WithSerializer(serializer func(w http.ResponseWriter, r *http.Request, ans any) error) func(*Server) {
+func WithSerializer(serializer Sender) func(*Server) {
 	return func(c *Server) { c.Serialize = serializer }
 }
 
-func WithErrorSerializer(serializer func(w http.ResponseWriter, r *http.Request, err error)) func(*Server) {
+// WithErrorSerializer sets a custom serializer of type ErrorSender that overrides the default one.
+// Please send a PR if you think the default serializer should be improved, instead of jumping to this option.
+func WithErrorSerializer(serializer ErrorSender) func(*Server) {
 	return func(c *Server) { c.SerializeError = serializer }
 }
 
--- a/serialization.go
+++ b/serialization.go
@@ -75,75 +75,79 @@ func transformOut[T any](ctx context.Context, ans T) (T, error) {
 	return ans, nil
 }
 
+type Sender func(http.ResponseWriter, *http.Request, any) error
+
 // Send sends a response.
 // The format is determined by the Accept header.
 func Send(w http.ResponseWriter, r *http.Request, ans any) error {
 	switch parseAcceptHeader(r.Header.Get("Accept"), ans) {
 	case "application/xml":
-		return SendXML(w, r, ans)
+		return SendXML(w, nil, ans)
 	case "text/html":
-		return SendHTML(r.Context(), w, ans)
+		return SendHTML(w, r, ans)
 	case "text/plain":
-		return SendText(w, ans)
+		return SendText(w, nil, ans)
 	case "application/json":
-		SendJSON(w, ans)
+		return SendJSON(w, nil, ans)
 	case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
-		SendYAML(w, ans)
-	default:
-		return errors.New("unsupported Accept header")
+		return SendYAML(w, nil, ans)
 	}
-
-	return nil
+	return errors.New("unsupported Accept header")
 }
 
 // SendYAML sends a YAML response.
 // Declared as a variable to be able to override it for clients that need to customize serialization.
-var SendYAML = func(w http.ResponseWriter, ans any) {
+var SendYAML = func(w http.ResponseWriter, _ *http.Request, ans any) error {
 	w.Header().Set("Content-Type", "application/x-yaml")
 	err := yaml.NewEncoder(w).Encode(ans)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
 		slog.Error("Cannot serialize returned response to YAML", "error", err)
 		_, _ = w.Write([]byte(`{"error":"Cannot serialize returned response to YAML"}`))
-		return
 	}
+	return err
 }
 
 // SendJSON sends a JSON response.
 // Declared as a variable to be able to override it for clients that need to customize serialization.
-var SendJSON = func(w http.ResponseWriter, ans any) {
+var SendJSON = func(w http.ResponseWriter, _ *http.Request, ans any) error {
 	w.Header().Set("Content-Type", "application/json")
 	err := json.NewEncoder(w).Encode(ans)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
 		slog.Error("Cannot serialize returned response to JSON", "error", err)
 		_, _ = w.Write([]byte(`{"error":"Cannot serialize returned response to JSON"}`))
-		return
 	}
+	return err
 }
+
+type ErrorSender = func(http.ResponseWriter, *http.Request, error)
 
 // SendError sends an error.
 // Declared as a variable to be able to override it for clients that need to customize serialization.
 var SendError = func(w http.ResponseWriter, r *http.Request, err error) {
 	switch parseAcceptHeader(r.Header.Get("Accept"), nil) {
 	case "application/xml":
-		SendXMLError(w, r, err)
+		SendXMLError(w, nil, err)
 	case "text/html":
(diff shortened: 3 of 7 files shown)


dylanhitt:
seems coveralls is having some problems 🤷

ccoVeille (commented):
Now all method are sharing the same signature

Do you consider adding a interface/type then ?

I'm asking because I don't see a reason to do this change unless you consider to add an interface or a type alias.

Cleaning up in not a reason for a refactoring that includes such breaking changes

please note, I'm not complaining about the change. I try to understand.

Having an interface/alias such as this one, would make sense with your PR

```go
type Sender func(http.ResponseWriter, *http.Request, any) error
```

But I see nothing about it here, is it something you plan to do later?

Having a code like the one currently in Send that choose the approriate Sender based on the type and return it, then apply `sender(w, r, ans)` in `Send` would be understood, but it's not what you did. You added useless parameter to many function. Something that can only be understood, because you want them to share the same signature. But there is no need now.

I have read the discussion that led to opening this PR. It mentions `WithSerializer`.

Shouldn't we see something that should let think the changes are made to match what it needs? So refactor `WithSerializer`to use the Sender interface/alias I mentioned.

I have very little knowledge on the codebase, and the project, but here are my feedbacks

dylanhitt:
I agree with you a type alias for would make good sense here.

dylanhitt:
@ccoVeille thanks for the feedback and great idea.

serialization.go
```diff
@@ -82,44 +82,42 @@ func Send(w http.ResponseWriter, r *http.Request, ans any) error {
...
+		return SendText(w, r, ans)
 	case "application/json":
-		SendJSON(w, ans)
+		return SendJSON(w, r, ans)
 	case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
-		SendYAML(w, ans)
+		return SendYAML(w, r, ans)
 	default:
 		return errors.New("unsupported Accept header")
 	}
```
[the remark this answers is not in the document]
dylanhitt: I assume this is just preference?

options.go
```diff
@@ -304,13 +304,13 @@ func WithLogHandler(handler slog.Handler) func(*Server) {
...
-// WithSerializer sets a custom serializer that overrides the default one.
+// WithSerializer sets a custom serializer of type Sender that overrides the default one.
 // Please send a PR if you think the default serializer should be improved, instead of jumping to this option.
-func WithSerializer(serializer func(w http.ResponseWriter, r *http.Request, ans any) error) func(*Server) {
+func WithSerializer(serializer Sender) func(*Server) {
 	return func(c *Server) { c.Serialize = serializer }
 }
 
-func WithErrorSerializer(serializer func(w http.ResponseWriter, r *http.Request, err error)) func(*Server) {
+func WithErrorSerializer(serializer ErrorSender) func(*Server) {
```
ccoVeille, line 313: Adding a comment to this method, could be like the one you modified for WithSerializer a good thing

serialization.go
```diff
@@ -158,34 +162,35 @@ func SendJSONError(w http.ResponseWriter, err error) {
...
 
 	w.WriteHeader(status)
-	SendJSON(w, err)
+	//nolint:errcheck
+	SendJSON(w, nil, err)
```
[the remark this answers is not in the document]
dylanhitt: going to pass on this for this PR at least

options.go
```diff
@@ -304,13 +304,13 @@ func WithLogHandler(handler slog.Handler) func(*Server) {
 	}
 }
 
-// WithSerializer sets a custom serializer that overrides the default one.
+// WithSerializer sets a custom serializer of type Sender that overrides the default one.
 // Please send a PR if you think the default serializer should be improved, instead of jumping to this option.
-func WithSerializer(serializer func(w http.ResponseWriter, r *http.Request, ans any) error) func(*Server) {
+func WithSerializer(serializer Sender) func(*Server) {
 	return func(c *Server) { c.Serialize = serializer }
```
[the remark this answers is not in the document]
ccoVeille: I'm talking about changing this

https://github.com/go-fuego/fuego/blob/f01dbefd0f6bbb21c5a608479d5f6cd862c5017a/options.go#L77-L78

options.go
ccoVeille: I cannot tell if you fixed, or if I missed, but it's good now :shrug:

serialization.go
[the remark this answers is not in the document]
ccoVeille: I always find better to

it's a simple convention to put the default at the end, but unless I'm wrong, (but [I don't think so](https://goplay.space/#Ctzlw3o958b))  "default" is always evaluated at the end when they are other "cases", but these cases are evaluated in the exact order they are. See my example https://goplay.space/#Ctzlw3o958b

Why am I talking about that ?

Because some people might add a case after the default, making the code hard to read.
Having a return at the end instead of a default help to keep things clear.

But yes, it's a personal preference, only based on code reviews and people ability to surprise me with strange code.

options.go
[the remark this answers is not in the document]
ccoVeille: thanks for doing the changes

ccoVeille (approved):

ccoVeille:
> @ccoVeille thanks for the feedback and great idea.

You are welcome

dylanhitt:
Added one more commit adding function `SendTextError`, `SendHTMLError`, and `SendYAMLError`

serialization.go
```diff
@@ -228,7 +234,23 @@ var SendHTML = func(ctx context.Context, w http.ResponseWriter, ans any) error {
...
+// SendHTMLError sends a HTML response.
+// If the error implements ErrorWithStatus, the status code will be set.
+func SendHTMLError(w http.ResponseWriter, _ *http.Request, err error) {
+	status := http.StatusInternalServerError
+	var errorStatus ErrorWithStatus
+	if errors.As(err, &errorStatus) {
+		status = errorStatus.StatusCode()
+	}
+
+	w.Header().Set("Content-Type", "text/html; charset=utf-8")
```
ccoVeille, line 246: Unless I'm wrong, SendHTML called a few line after already sends this header

serialization.go
```diff
@@ -245,6 +267,20 @@ func SendText(w http.ResponseWriter, ans any) error {
...
+// SendTextError sends a Text response.
+// If the error implements ErrorWithStatus, the status code will be set.
+func SendTextError(w http.ResponseWriter, _ *http.Request, err error) {
+	status := http.StatusInternalServerError
+	var errorStatus ErrorWithStatus
+	if errors.As(err, &errorStatus) {
+		status = errorStatus.StatusCode()
+	}
+
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
```
ccoVeille, line 279: Unless I'm wrong, SendText called a few line after already sends this header

serialization_test.go
```diff
@@ -259,36 +267,54 @@ func (errorWriter) Header() http.Header {
...
 
 	t.Run("error", func(t *testing.T) {
 		errorWriter := &errorWriter{}
-		SendJSON(errorWriter, response{Message: "Hello World", Code: 200})
+		SendJSON(errorWriter, nil, response{Message: "Hello World", Code: http.StatusOK})
 		require.Contains(t, errorWriter.arg, "Cannot serialize returned response to JSON")
 	})
 }
 
+func TestSendHTMLError(t *testing.T) {
```
ccoVeille, line 309: I think  TestSendHTML should be added before  TestSendHTMLError, because I don't see it.

maybe I'm wrong, and then it's only a matter of moving this test aside  TestSendHTML

serialization_test.go
```diff
@@ -239,11 +239,19 @@ func TestJSONError(t *testing.T) {
 
 func TestSend(t *testing.T) {
 	w := httptest.NewRecorder()
-	SendText(w, "Hello World")
+	SendText(w, nil, "Hello World")
 
 	require.Equal(t, "Hello World", w.Body.String())
```
ccoVeille, line 244: I would like to recommend you to test the content type here and in all other tests, such as : 
-  TestSendTextError
- TestSendYAML
- TestSendJSON

Please note  TestSendYAMLError and TestSendHTMLError already test the content type

serialization_test.go
dylanhitt: I didn't affect the logic with SendHTML. However, I agree, but it should be test with an additional PR/commit. That's a bit more focused. 

Just my opinion. I can add it if you prefer here.

serialization.go
```diff
@@ -245,6 +266,19 @@ func SendText(w http.ResponseWriter, ans any) error {
 	return err
 }
 
+// SendTextError sends a Text response.
+// If the error implements ErrorWithStatus, the status code will be set.
+func SendTextError(w http.ResponseWriter, _ *http.Request, err error) {
+	status := http.StatusInternalServerError
+	var errorStatus ErrorWithStatus
+	if errors.As(err, &errorStatus) {
```
dylanhitt, line 274: Need test case

serialization.go
```diff
@@ -75,75 +75,93 @@ func transformOut[T any](ctx context.Context, ans T) (T, error) {
...
+		SendHTMLError(w, nil, err)
 	case "text/plain":
-		_ = SendText(w, err)
+		SendTextError(w, r, err)
 	case "application/json":
-		SendJSONError(w, err)
+		SendJSONError(w, nil, err)
 	case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
-		SendYAML(w, err)
+		SendYAMLError(w, nil, err)
```
dylanhitt, line 156: Need test case

serialization.go
```diff
@@ -75,75 +75,93 @@ func transformOut[T any](ctx context.Context, ans T) (T, error) {
...
 	switch parseAcceptHeader(r.Header.Get("Accept"), nil) {
 	case "application/xml":
-		SendXMLError(w, r, err)
+		SendXMLError(w, nil, err)
 	case "text/html":
-		_ = SendHTMLError(r.Context(), w, err)
+		SendHTMLError(w, nil, err)
 	case "text/plain":
-		_ = SendText(w, err)
+		SendTextError(w, r, err)
```
dylanhitt, line 152: Need test case

serialization.go
```diff
@@ -75,75 +75,93 @@ func transformOut[T any](ctx context.Context, ans T) (T, error) {
...
+	}
+	return err
+}
+
+// SendYAMLError sends a YAML error response.
+// If the error implements ErrorWithStatus, the status code will be set.
+func SendYAMLError(w http.ResponseWriter, _ *http.Request, err error) {
+	status := http.StatusInternalServerError
+	var errorStatus ErrorWithStatus
+	if errors.As(err, &errorStatus) {
```
dylanhitt, line 116: Need test case

serialization.go
```diff
@@ -228,7 +230,22 @@ var SendHTML = func(ctx context.Context, w http.ResponseWriter, ans any) error {
 	return fmt.Errorf("cannot serialize HTML from type %T (not string, fuego.HTML and does not implement fuego.CtxRenderer or fuego.Renderer)", ans)
 }
 
-func SendText(w http.ResponseWriter, ans any) error {
+// SendHTMLError sends a HTML response.
+// If the error implements ErrorWithStatus, the status code will be set.
+func SendHTMLError(w http.ResponseWriter, _ *http.Request, err error) {
+	status := http.StatusInternalServerError
+	var errorStatus ErrorWithStatus
+	if errors.As(err, &errorStatus) {
```
dylanhitt, line 238: Need test case

examples/full-app-gourmet/controller/recipe.go
```diff
@@ -35,7 +35,7 @@ func (rs recipeRessource) getAllRecipesStandardWithHelpers(w http.ResponseWriter
 		return
 	}
 
-	fuego.SendJSON(w, recipes)
+	fuego.SendJSON(w, nil, recipes)
```
EwenQuim, line 38: A good example would be using "r", whether it is used or not.
Users do not know the implementation.

serialization.go
```diff
@@ -158,53 +172,41 @@ func SendJSONError(w http.ResponseWriter, err error) {
...
 // SendHTML sends a HTML response.
 // Declared as a variable to be able to override it for clients that need to customize serialization.
-var SendHTML = func(ctx context.Context, w http.ResponseWriter, ans any) error {
+var SendHTML = func(w http.ResponseWriter, r *http.Request, ans any) error {
 	w.Header().Set("Content-Type", "text/html; charset=utf-8")
 
 	ctxRenderer, ok := any(ans).(CtxRenderer)
 	if ok {
-		return ctxRenderer.Render(ctx, w)
+		return ctxRenderer.Render(r.Context(), w)
```
EwenQuim, line 209: Indeed, sorry that was weird from my part, and moreover fuego's context is equal to r.Context, so I really don't know why I've done this

EwenQuim (commented):
Thank you for this refactoring!

ccoVeille (approved):

EwenQuim (approved):

PR #165, opened 2024-08-23
feat: add group parameters

rizerkrof:


options.go
```diff
@@ -414,3 +416,39 @@ func (s *Server) RemoveTags(tags ...string) *Server {
 	}
 	return s
 }
+
+// Header registers a header parameter for the route.
+func (s *Server) Param(name, description string, params ...OpenAPIParam) *Server {
```
EwenQuim, line 421: Wrong description

options.go
```diff
@@ -414,3 +416,39 @@ func (s *Server) RemoveTags(tags ...string) *Server {
...
+
+func (s *Server) Cookie(name, description string, params ...OpenAPIParam) *Server {
+	s.Param(name, description, append(params, OpenAPIParam{Type: "cookie"})...)
+	return s
+}
+
+func (s *Server) Query(name, description string, params ...OpenAPIParam) *Server {
+	s.Param(name, description, append(params, OpenAPIParam{Type: "query"})...)
+	return s
+}
```
EwenQuim, line 454: Adding coms would be really nice here :) 

Documentation near the usage is amazing

options_test.go
```diff
@@ -395,3 +395,17 @@ func TestCustomSerialization(t *testing.T) {
...
+
+	Get(group, "/test", controller)
+	Get(group, "/test2", controller)
+
+	require.Equal(t, group.params, []GroupParams{{Name: "X-Test-Header", Description: "test-value", OpenAPIParam: OpenAPIParam{Required: false, Example: "", Type: "header"}}})
+
+	document := s.OutputOpenAPISpec()
+	require.Equal(t, document.Paths.Find("/api/test").Get.Parameters[0].Value.Name, "X-Test-Header")
+	require.Equal(t, document.Paths.Find("/api/test2").Get.Parameters[0].Value.Name, "X-Test-Header")
+}
```
EwenQuim, line 411: Thanks for thinking about tests!

I would add more tests

- query parameters
- verify it doesn't impact sibling groups / parent groups / children groups / main server instance
- what happens if I add `Hide()` method on the group? Does it panics?

openapi_operations.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/getkin/kin-openapi/openapi3"
 )
 
+type GroupParams struct {
+	Name        string
+	Description string
+	OpenAPIParam
+}
+
 type OpenAPIParam struct {
```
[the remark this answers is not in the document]
dylanhitt: Being pre `1.0.0`. I'm fine with a break now rather than later.

openapi_operations.go
```diff
@@ -6,6 +6,12 @@ import (
...
 	Required bool
 	Example  string
```
dylanhitt, line 17: If we do elect to break should we type def `Type` here
```go
type ParamType string

const (
  ParamTypeQuery ParamType = "query'
  ... etc
)

type OpenAPIParamOption {
  ...
  Type ParamType
}

```

EwenQuim (approved):
Will release a new 0.15 version for this feature

Issue #166, opened 2024-08-23
Readme's example with error


dhanielsales:
**To Reproduce**

1. Create a empty go project;
2. Copy the **Simple POST** example from the README (https://github.com/go-fuego/fuego/tree/main?tab=readme-ov-file#simple-post) to inside of the empty go project;
3. Run the project and hit the API with no body on the request.

**Expected behavior**
I took as an expected behavior receive a 400 error with the error's details but I received 400 with "cannot decode request body: EOF" on the detail.

**Screenshots**

![Captura de tela de 2024-08-23 20-40-16](https://github.com/user-attachments/assets/7576dc27-5d68-46c6-a61d-948eadb124d9)
![Captura de tela de 2024-08-23 20-36-08](https://github.com/user-attachments/assets/fa486c22-5b84-4403-ba02-2b6a2ef6f8e9)

**Framework version: v0.14.0**

**Go version: v1.22.6**

**Additional context**
I already know how to fix this problem, if you want I can open a PR with the fix :)

PR #168, opened 2024-08-24
Fixing EOF bad request error when consume a route with empty request body

dhanielsales:
Related issue: https://github.com/go-fuego/fuego/issues/166

### DIFF
--- a/deserialization.go
+++ b/deserialization.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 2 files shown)


deserialization.go
```diff
@@ -94,7 +94,7 @@ func read[B any](context context.Context, dec decoder) (B, error) {
 	var body B
 
 	err := dec.Decode(&body)
-	if err != nil {
+	if err != nil && err != io.EOF {
```
ccoVeille, line 97: I would expect this.

```suggestion
	if err != nil && !errors.Is(err, io.EOF) {
```

dhanielsales: Good call, thanks

EwenQuim (approved):
Thank you!

Issue #167, opened 2024-08-24
Can't get all values from query params


dhanielsales:
First of all, I simply loved the idea of ​​the project! Congrats! 

Usually I create a similar structure for my personal projects and it was really fun seen it here, so I'll try to contib with my experiences :)

So, going to the suggestion: 
I noticed that the interfaces `fuego.ContextWithBody[Body]` and of couse, `fuego.ContextNoBody` doesn't have a method to grab all Query params from the request, insted of it, you only have those methods to handle with query params:

```go
QueryParam(name string) string
QueryParamInt(name string, defaultValue int) int // If the query parameter does not exist or is not an int, it returns the default given value. Use [Ctx.QueryParamIntErr] if you want to know if the query parameter is erroneous.
QueryParamIntErr(name string) (int, error)
QueryParamBool(name string, defaultValue bool) bool // If the query parameter does not exist or is not a bool, it returns the default given value. Use [Ctx.QueryParamBoolErr] if you want to know if the query parameter is erroneous.
QueryParamBoolErr(name string) (bool, error)
QueryParams() map[string]string
```

Non of them can handle with multiple values, so my suggestions is to add a new method like `QueryParamArr(name string) []string` to handle with this type of things.

Also, for the `QueryParams() map[string]string` I would not recoment you guys to treat the query params as an `map[string]string` because it will suffer by the same problem which I mentioned just above. So, my seconde suggestion is to change it for `url.Values`. It is an standard type for query params in go and it's a `map[string][]string`.

So, as I said in another issue, I already know how to fix thos problems, if you want I can open a PR with the fix :)

Take care :D

EwenQuim:
Yes very good idea!

Happy to review a PR if you have time. I'll do it next monday otherwise, I'm pretty busy this weekend.

Thank you for reporting this, feedback is always valuable 🙏🏼

drwpls:
@EwenQuim It's should be `fuego.ContextWithParams[T]` there to automatic format incoming GET request parameters, which should be use like:
```Go
fuego.Get(api, "/test", func(c fuego.ContextWithParams[types.TestRequest] (*types.TestResponse, error) {
		req := c.Params()
		// handle this req params
                _ = req
		return &types.TestResponse{
			hello: "world",
		}, nil
	})
```

The parameters parsing can be use inside fuego (from `*http.Request`:
```Go
func ParseFormValues[T any](r *http.Request) (*T, error) {
	if err := r.ParseForm(); err != nil {
[...]

PR #169, opened 2024-08-24
Enhance query parameters to support a list of parameters

dhanielsales:
Related issue: https://github.com/go-fuego/fuego/issues/167

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"log/slog"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


EwenQuim (approved):

dylanhitt:
Instead of QueryParamArr. Why don’t we just just fix the concept of QueryParam to return an array?

EwenQuim:
> Instead of QueryParamArr. Why don’t we just just fix the concept of QueryParam to return an array?

Because 90% of the time only one param is used, so it's more convenient. That's why the std lib uses [0] also by default

image

dhanielsales:
Exactly. The standard type `url.Values` have a Get method to return the first value on the slice.

dylanhitt:
Ah, excellent points. This is great!

dylanhitt (approved):

PR #172, opened 2024-08-26
WIP: attempt to send HTML/JSON when Accept Header: `*/*`

dylanhitt:
Per discussion #162 

Iterate through all values of `Accept` Header when choosing a response type. If the accept header of `*/*` is found fuego will attempt to first respond with HTML and then JSON if it is unable to Render HTML.

Will add tests and such if this is acceptable.


dylanhitt:
Closing as I completely missed what parseAcceptHeader was doing

PR #175, opened 2024-08-28
Chore/send default

dylanhitt:
Per #173 and closes #162

### DIFF
--- a/serialization.go
+++ b/serialization.go
@@ -94,6 +94,10 @@ func Send(w http.ResponseWriter, r *http.Request, ans any) (err error) {
 			err = SendJSON(w, nil, ans)
 		case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
 			err = SendYAML(w, nil, ans)
+		default:
+			// if we don't support the header
+			// keep iterating
+			continue
 		}
 
 		if err != nil {
@@ -159,18 +163,22 @@ var SendError = func(w http.ResponseWriter, r *http.Request, err error) {
 		switch inferAcceptHeader(header, nil) {
 		case "application/xml":
 			SendXMLError(w, nil, err)
+			return
 		case "text/html":
 			SendHTMLError(w, nil, err)
+			return
 		case "text/plain":
 			SendTextError(w, r, err)
+			return
 		case "application/json":
 			SendJSONError(w, nil, err)
+			return
 		case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
 			SendYAMLError(w, nil, err)
-		default:
-			SendJSONError(w, r, err)
+			return
 		}
 	}
+	SendJSONError(w, r, err)
 }
 
 // SendJSONError sends a JSON error response.
--- a/serialization.go
+++ b/serialization.go
@@ -79,19 +79,30 @@ type Sender func(http.ResponseWriter, *http.Request, any) error
 
 // Send sends a response.
 // The format is determined by the Accept header.
+// If Accept header `*/*` is found Send will Attempt to send
+// HTML, and the JSON.
 func Send(w http.ResponseWriter, r *http.Request, ans any) error {
-	switch parseAcceptHeader(r.Header.Get("Accept"), ans) {
-	case "application/xml":
-		return SendXML(w, nil, ans)
-	case "text/html":
-		return SendHTML(w, r, ans)
-	case "text/plain":
-		return SendText(w, nil, ans)
-	case "application/json":
-		return SendJSON(w, nil, ans)
-	case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
-		return SendYAML(w, nil, ans)
+	for _, header := range r.Header.Values("Accept") {
+		switch header {
+		case "application/xml":
+			return SendXML(w, nil, ans)
+		case "text/html":
+			return SendHTML(w, r, ans)
+		case "text/plain":
+			return SendText(w, nil, ans)
+		case "application/json":
+			return SendJSON(w, nil, ans)
+		case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
+			return SendYAML(w, nil, ans)
+		case "*/*":
+			// attempt to return html or default to json
+			if SendHTML(w, r, ans) == nil {
+				return nil
+			}
+			return SendJSON(w, nil, ans)
+		}
 	}
+
 	return errors.New("unsupported Accept header")
 }
 
--- a/serialization.go
+++ b/serialization.go
@@ -80,30 +80,35 @@ type Sender func(http.ResponseWriter, *http.Request, any) error
 // Send sends a response.
 // The format is determined by the Accept header.
 // If Accept header `*/*` is found Send will Attempt to send
-// HTML, and the JSON.
-func Send(w http.ResponseWriter, r *http.Request, ans any) error {
-	for _, header := range r.Header.Values("Accept") {
-		switch header {
+// HTML, and then JSON.
+func Send(w http.ResponseWriter, r *http.Request, ans any) (err error) {
+	for _, header := range parseAcceptHeader(r.Header) {
+		switch inferAcceptHeader(header, ans) {
 		case "application/xml":
-			return SendXML(w, nil, ans)
+			err = SendXML(w, nil, ans)
 		case "text/html":
-			return SendHTML(w, r, ans)
+			err = SendHTML(w, r, ans)
 		case "text/plain":
-			return SendText(w, nil, ans)
+			err = SendText(w, nil, ans)
 		case "application/json":
-			return SendJSON(w, nil, ans)
+			err = SendJSON(w, nil, ans)
 		case "application/x-yaml", "text/yaml; charset=utf-8", "application/yaml": // https://www.rfc-editor.org/rfc/rfc9512.html
-			return SendYAML(w, nil, ans)
-		case "*/*":
-			// attempt to return html or default to json
-			if SendHTML(w, r, ans) == nil {
-				return nil
-			}
-			return SendJSON(w, nil, ans)
+			err = SendYAML(w, nil, ans)
 		}
+
+		if err != nil {
+			continue
+			// should we log and continue?
+			// or
+			// if err == nil we just return?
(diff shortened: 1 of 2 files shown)


dylanhitt:
Maybe we should consider what RFC says for this: https://datatracker.ietf.org/doc/html/rfc2616#section-10.4.7

>   Note: HTTP/1.1 servers are allowed to return responses which are
      not acceptable according to the accept headers sent in the
      request. In some cases, this may even be preferable to sending a
      406 response. User agents are encouraged to inspect the headers of
      an incoming response to determine if it is acceptable.

dylanhitt:
So right now is the current tested case for:

```golang
        r.Header.Set("Accept", "application/json")
	Send(w, r, &StdRenderer{
		templates:         template,
		templateToExecute: templateName,
	})
```

The above will respond with `{}\n`. The above RFC states that we are allowed to respond with HTML or a 406. The above also states (at least how I interpret it), that we should be responding either with a 406 or with our inferred payload regardless if the caller provided any correct `Accept` Header.  So if the caller provided and `Accept` header of `not-a-chance` we should still respond either with our inferred type or a `406`.

serialization_test.go
```diff
@@ -429,43 +446,76 @@ func TestInferAcceptHeaderFromType(t *testing.T) {
...
+		header.Set("Accept", "text/html,text/plain")
+		accept := parseAcceptHeader(header)
+		require.Equal(t, []string{"text/html", "text/plain"}, accept)
+	})
+
+	t.Run("can parse string from real browser", func(t *testing.T) {
+		header := http.Header{}
+		header.Set("Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
+		accept := parseAcceptHeader(header)
+		require.Equal(t, []string{"text/html", "application/xhtml+xml", "application/xml;q=0.9", "*/*;q=0.8"}, accept)
```
EwenQuim, line 515: What can we do about the [Quality values](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values)?

I suggest to discard them, as the browser will always order the list of mime types correctly. We might do this in another PR if it really is needed.

The target array would be `require.Equal(t, []string{"text/html", "application/xhtml+xml", "application/xml", "*/*"}, accept)`

serialization.go
```diff
@@ -314,15 +332,17 @@ func InferAcceptHeaderFromType(ans any) string {
...
+		return InferAcceptHeaderFromType(ans)
 	}
+	return accept
+}
+
+func parseAcceptHeader(header http.Header) []string {
+	accept := header.Get("Accept")
 	if accept == "" {
-		accept = InferAcceptHeaderFromType(ans)
+		return []string{""}
```
EwenQuim, line 345: With this behaviour, if no Accept header is sent, nothing will be sent. I suggest to provide application/json in that case, what do you think?

serialization_test.go
dylanhitt: 👍 

Probably should switch the `*/*` to strings.Contains

serialization.go
dylanhitt: Take a look at `inferAcceptHeader`. Here is the test case: https://github.com/go-fuego/fuego/pull/175/files/60bf483061906c646f99dfb4ac7c25a6c67378a7#diff-95e7bbc519ba50d88c1639d2f79e997f1011f802b8c69605813150752b40c535R470

serialization_test.go
dylanhitt: Ignoring Quality values entirely.

dylanhitt:
I will update the TestSend test to be more robust once we land on how we want to handle the cases mentioned above and the comments from: https://github.com/go-fuego/fuego/pull/173#discussion_r1734957641

Feel free to play with PR as well as writing test cases is easier than explaining through comments sometimes.

serialization.go
```diff
@@ -79,20 +79,36 @@ type Sender func(http.ResponseWriter, *http.Request, any) error
...
+			// if err == nil we just return?
+		}
+		return nil
+	}
+
+	if err != nil {
+		return err
 	}
-	return errors.New("unsupported Accept header")
+	return errors.New("no supported Accept header was not provided from: " + r.Header.Get("Accept"))
```
ccoVeille, line 111: I'm sorry but I'm doubtful about your algorithm

First, if you have a header you don't know

Then outside the loop you are testing if you have an error something that couldn't happen

I would like to suggest you this

```go
func Send(w http.ResponseWriter, r *http.Request, ans any) (err error) { 
         for _, header := range parseAcceptHeader(r.Header) { 
                 switch inferAcceptHeader(header, ans) { 
                 case "application/xml": 
                         return SendXML(w, nil, ans) 
                 case "text/html": 
                         return SendHTML(w, r, ans) 
                 case "text/plain": 
                        return SendText(w, nil, ans) 
                 case "application/json": 
                         return SendJSON(w, nil, ans) 
                         return SendYAML(w, nil, ans) 
                 } 
         } 
 
         return errors.New("no supported Accept header was not provided from: " + r.Header.Get("Accept")) 
  }
```

There is no need to continue:
- if it works
- if there is an error, something that break the JSON is likely to break the HTML. A database error is likely to happen for each content type
- also, if the JSON is broken, you should stop and not try another content type, you would have silent errors that will fill your log, and people asking for a json,yaml,html via header, might be disappointed to receive a yaml if json fails. The error handling will be a nightmare on their side.

For this reason, using return should be preferred.

If you want to log you can log in the caller

ccoVeille (changes requested):

serialization.go
ccoVeille: My comment sounds like what @EwenQuim reported [here](https://github.com/go-fuego/fuego/pull/173#discussion_r1734957641)

I'm only see what was posted in other PR now

dylanhitt: I think you're misunderstanding the original point. 

The case this is trying solve is when a caller passes something like `text/html, */*`, and the route they are returns only JSON. In that case with the algorithm you're proposing. We would just fail to respond with `text/html` and never attempt any other Accept types.

dylanhitt: Your second point 

> if there is an error, something that break the JSON is likely to break the HTML. A database error is likely to happen for each content type

Is handled by a controller returning and error. This code block wouldn't execute see: https://github.com/go-fuego/fuego/blob/main/serve.go#L113

Third point 

> also, if the JSON is broken, you should stop and not try another content type, you would have silent errors that will fill your log, and people asking for a json,yaml,html via header, might be disappointed to receive a yaml if json fails. The error handling will be a nightmare on their side.

if the caller is passing something like: `application/json, application/yaml` then we should try to fulfill whatever we can for the caller. Even if the caller asks for just `application/xml` on a route intending to return json and it cannot be fulfilled the HTTP spec [states](https://datatracker.ietf.org/doc/html/rfc2616#section-10.4.7):

> Note: HTTP/1.1 servers are allowed to return responses which are
      not acceptable according to the accept headers sent in the
      request. In some cases, this may even be preferable to sending a
      406 response. User agents are encouraged to inspect the headers of
      an incoming response to determine if it is acceptable.
      
This code only attempts to fulfill what the caller is asking for. If they are only asking for `application/json` then we will only attempt json. If the caller is asking for `application/json, */*`, the caller is giving permission to try anything that can possibly be returned; this is typically a browser.

ccoVeille: > I think you're misunderstanding the original point.

I notice now. I only read the previous PR after posting my initial comment here.

dylanhitt: I will say I do agree with you. If I as I caller pass `application/xml, */*` and I get a response of json on error. That would be strange, however the caller is telling us we're allowed to do that with that Accept Header.

dylanhitt: Too provide a bit more context from a browser if you're hitting a JSON endpoint you will get this:

image

As we're failing to render any html.

ccoVeille: There is something strange anyway with the logic of handling of */*

I would say the order of the list is important.

If the user sends "application/rss, application/json", I would expect a json to be rendered

If the user sends "application/rss, */*", I would expect one supported format to be tried and an error to be returned if it fails

But then, it means that you need to define a priority between each content type handler.

I would expect */* to always try the exact same order (for example)

- application/json
- application/html

And that's it, not necessarily ALL possible handler. I mean no one should expect a YAML when someone asks */*

It's a "show me what you got" (Rick and Morty reference), the client expects to get interesting things.

dylanhitt: so `*/*` doesn't try all possible handlers. `*/*` is inferred based on the response type of the route. 

```golang
func InferAcceptHeaderFromType(ans any) string {
	_, ok := any(ans).(string)
	if ok {
		return "text/plain"
	}
	_, ok = any(ans).(*string)
	if ok {
		return "text/plain"
	}
	_, ok = any(ans).(HTML)
	if ok {
		return "text/html"
	}
	_, ok = any(ans).(CtxRenderer)
	if ok {
		return "text/html"
	}
	_, ok = any(&ans).(*CtxRenderer)
	if ok {
		return "text/html"
	}
	_, ok = any(ans).(Renderer)
	if ok {
		return "text/html"
	}
	_, ok = any(&ans).(*Renderer)
	if ok {
		return "text/html"
	}
	return "application/json"
}

func inferAcceptHeader(accept string, ans any) string {
	if accept == "" || accept == "*/*" {
		return InferAcceptHeaderFromType(ans)
	}
	return accept
}
```

ccoVeille: OK. It's clearer. I should have looked more carefully at the code

dylanhitt:
Last concern on SendError. The error response from the original Send and SendError would be different and not match. For instance the code now would allow this test to pass

```
func TestSendWhenError(t *testing.T) {
	w := httptest.NewRecorder()
	r := httptest.NewRequest(http.MethodGet, "/", nil)
	r.Header.Set("Accept", "text/junk,application/json,text/html")
	errorWriter := &errorWriter{}
	err := Send(errorWriter, r, response{})
	require.Error(t, err)
	SendError(w, r, err)
	require.Equal(t, "application/json", w.Header().Get("Content-Type"))
	body := w.Body.String()
	require.Equal(t, "cannot write on an errorWriter", body)
}
```

SendError is going to respond with it's first acceptable found content type. Previously before adding the `return`'s to each clause it would respond with the last acceptable types. I personally think the erroring content-type in Send should be the type we always respond with in `SendError`. With the current structure of this code the only way to do this is mutate the Accept header in the Send function

EwenQuim (approved):

EwenQuim:
> Maybe this is a bit more complex and needs to be re thought about more than I anticipated 🤷. Just going to apologize for all the spam. I didn't realize the can of worms I was opening.

No it's OK, a framework is the abstraction that is needed to handle cases like this. I'm merging to make stuff move and deliver things. Thanks for your work.

PR #176, opened 2024-08-28
Variadic route options

EwenQuim:
Implements & closes #174.

See the petstore for an example of the feature (`examples/petstore/controllers/pets.go`)

Still in progress.

- We need to apply this pattern to `Groups` too
- Add checks for typing, nullable/required etc in the controller

## Summary by CodeRabbit

- **New Features**
	- Introduced automatic fixing of linting issues for Go files.
	- Added filtering functionality for pets in the pet store API, allowing retrieval based on criteria like name and age.
	- Enhanced OpenAPI documentation with new query and header parameters, including pagination options.
	- Introduced a structured approach for managing route options and parameters in the web framework.
	- Improved clarity and organization of API endpoint definitions with explicit option parameters.

- **Bug Fixes**
	- Corrected naming inconsistencies in test functions for improved clarity.

- **Documentation**
	- Updated README and OpenAPI specifications to reflect new features and improved parameter handling.

- **Tests**
	- Added comprehensive tests for new functionalities, including filtering and parameter handling in the API.


option/option.go
```diff
@@ -0,0 +1,182 @@
...
+		Param(name, options...)(r)
+	}
+}
+
+// Declare an integer query parameter for the route.
+// This will be added to the OpenAPI spec.
+// The query parameter is transmitted as a string in the URL, but it is parsed as an integer.
+// Example:
+//
+//	Query("age", "Filter by age (in years)", param.Example("3 years old", 3), param.Nullable())
```
dylanhitt, line 53: nit: 

```suggestion
//	QueryInt("age", "Filter by age (in years)", param.Example("3 years old", 3), param.Nullable())
```

EwenQuim: Thanks, indeed!

EwenQuim:
@dylanhitt ready for review

The amount of work is quite huge at the moment on Fuego but once v0.15 will be released, we'll be good for some time :)

This might be one of the last big features.

dylanhitt:
Okay nice, how much of a hurry is this? It’s a 3 day weekend in the states and I’m out of town. 

Is next Tuesday or Wednesday fine?

EwenQuim:
Yes of course it will be OK, it will let me time to update the documentation website and maybe 1/2 additional tests.

Thanks again!

mux.go
```diff
@@ -44,66 +44,51 @@ func Group(s *Server, path string) *Server {
...
 	mainRouter *Server // ref to the main router, used to register the route in the OpenAPI spec
 }
 
 // Capture all methods (GET, POST, PUT, PATCH, DELETE) and register a controller.
-func All[T, B any, Contexted ctx[B]](s *Server, path string, controller func(Contexted) (T, error), middlewares ...func(http.Handler) http.Handler) Route[T, B] {
-	return Register(s, Route[T, B]{
-		Path:     path,
-		FullName: FuncName(controller),
-	}, HTTPHandler(s, controller), middlewares...)
+func All[ReturnType, Body any, Contexted ctx[Body]](s *Server, path string, controller func(Contexted) (ReturnType, error), options ...func(*BaseRoute)) *Route[ReturnType, Body] {
```
dylanhitt, line 63: Just to clarify for myself. The only way to add middlewares to a specific route now is to use `options`

ctx.go
```diff
@@ -322,10 +366,10 @@ func (c ContextNoBody) QueryParamBoolErr(name string) (bool, error) {
 	return b, nil
 }
 
-func (c ContextNoBody) QueryParamBool(name string, defaultValue bool) bool {
+func (c ContextNoBody) QueryParamBool(name string) bool {
```
dylanhitt, line 369: Probbaly want to comment this similar to `QueryParamInt`. For instance what is the default value?

ctx.go
```diff
@@ -307,6 +346,11 @@ func (c ContextNoBody) QueryParamInt(name string, defaultValue int) int {
 func (c ContextNoBody) QueryParamBoolErr(name string) (bool, error) {
```
dylanhitt, line 346: > // If the query parameter does not exist or is not a bool, it returns nil.

It seems this always returns a bool as the first return type? Maybe look at this comment while we're here.

ctx_params_test.go
```diff
@@ -0,0 +1,39 @@
...
+			option.QueryInt("age", "Age", param.Nullable(), param.Default(18), param.Example("example1", 1)),
+			option.QueryBool("is_ok", "Is OK?", param.Default(true), param.Example("example1", true)),
+		)
+
+		t.Run("Default should correctly set parameter in controller", func(t *testing.T) {
+			r := httptest.NewRequest("GET", "/test", nil)
+			w := httptest.NewRecorder()
+			s.Mux.ServeHTTP(w, r)
+
+			require.Equal(t, 200, w.Code)
```
dylanhitt, line 35:
```suggestion
			require.Equal(t, http.StatusOK, w.Code)
```

option/option.go
```diff
@@ -0,0 +1,242 @@
...
+
+// Deprecated marks the route as deprecated.
+func Deprecated() func(*fuego.BaseRoute) {
+	return func(r *fuego.BaseRoute) {
+		r.Operation.Deprecated = true
+	}
+}
+
+// // AddError adds an error to the route.
+// func AddError(code int, description string, errorType ...any) func(*fuego.BaseRoute) {
```
dylanhitt, line 221: What is the plan here?

Also `RequestContent`

mux.go
EwenQuim: For route-scoped middlewares, yes.

Using `option.Middleware(myMiddleware)`

ctx.go
EwenQuim: The one defined in `option.QueryBool("myParam", "myDesc", param.Default(true))`. If not, `false`

It needs a comment yes

mux.go
dylanhitt: 👍

option/option.go
EwenQuim: Yes i have trouble inferring the returned type and we need it for these 2 options

dylanhitt (changes requested):
Just a few things mentioned. 

I'm going to take one more pass through this tomorrow. 

Other than that. This was excellent. Very readable

option/option.go
dylanhitt: Got it. So we will generally keep this the same moving forward?

EwenQuim: We don't really _need_ it but it is done that way in the current code.

But I can refactor it, I just need to find a way to do it

EwenQuim: Mmh I think we'll try to delete the old chaining options, for consistency. I hope it's feasible. The new options pattern do not have access to the ReturnType and BodyType. That's the only difference. It could be both a blessing and a curse, in the future.

dylanhitt: I guess for right now. Leave it as it is. But let's at least get this comment out of here for cleanliness.

dylanhitt:
Looking to look this over one more time tomorrow. 

Been a bit busy lately.

EwenQuim:
The last commit was fixing the last details. Ready for release

EwenQuim:
Btw testing CodeRabbit has been a little disappointing haha

Do you think we should remove it?

dylanhitt:
https://github.com/go-fuego/fuego/pull/176#issuecomment-2344629327

Yes.

option/option.go
```diff
@@ -0,0 +1,250 @@
...
+	response := openapi3.NewResponse().
+		WithDescription(description).
+		WithContent(content)
+
+	operation.AddResponse(code, response)
+}
+
+// RequestContentType sets the accepted content types for the route.
+// By default, the accepted content types is */*.
+func RequestContentType(consumes ...string) func(*fuego.BaseRoute) {
```
dylanhitt, line 246: I don't believe this will produce a proper result. I remember having to do this in specific way.  

I'll branch off and show an example.

dylanhitt: Take a look at https://github.com/go-fuego/fuego/pull/193/files

You will see that I'm declaring two different request content types. For each route. But in this case both routes end up referencing the same requestBody. The route will have to only reference the schema in this case. Hence why the logic for generating this bit of the route occurred here: https://github.com/go-fuego/fuego/pull/177/files#diff-127c9bd6b06d3b470c2b0f10ec1c538b7798d9fca66d4812d554a0f3a15fbf92R97

mux.go
```diff
@@ -44,66 +44,52 @@ func Group(s *Server, path string) *Server {
...
+	Method               string              // HTTP method (GET, POST, PUT, PATCH, DELETE)
+	Path                 string              // URL path. Will be prefixed by the base path of the server and the group path if any
+	Handler              http.Handler        // handler executed for this route
+	FullName             string              // namespace and name of the function to execute
+	Params               map[string]OpenAPIParam
+	Middlewares          []func(http.Handler) http.Handler
+	AcceptedContentTypes []string // Content types accepted for the request body. If nil, all content types (*/*) are accepted.
 
-	mainRouter *Server // ref to the main router, used to register the route in the OpenAPI spec
+	MainRouter *Server // PRIVATE ref to the main router, used to register the route in the OpenAPI spec
```
dylanhitt, line 60:
```suggestion
	MainRouter *Server // ref to the main router, used to register the route in the OpenAPI spec
```

dylanhitt (changes requested):

option/option.go
dylanhitt: As discussed earlier not having access to the `RequestBody` is very unfortunate. But I guess now. The best way to do this will if to check if `if r.AcceptedContentTypes !=nil` and branch into a new code path. Could probably just reuse the code that is already in the old `RequestContentType`.

FYI I do plan on adding the server wide default option once this is done. Hence why I'm so fixated on this 😄

EwenQuim: Mmh I see

ctx.go
```diff
@@ -265,17 +267,44 @@ func (c ContextNoBody) QueryParams() url.Values {
...
+	_, ok := c.params[name]
+	if !ok {
+		slog.Warn("query parameter not expected in OpenAPI spec", "param", name)
+	}
 	return c.Req.URL.Query()[name]
 }
 
 // QueryParam returns the query parameter with the given name.
+// If it does not exist, it returns an empty string.
+// If it does not exist and there is a default value declared in the OpenAPI spec, it returns the default value.
```
ccoVeille, line 279:
```suggestion
// If it does not exist, it returns an empty string, unless there is a default value declared in the OpenAPI spec.
```

openapi.go
```diff
@@ -171,17 +173,13 @@ func RegisterOpenAPIOperation[T, B any](s *Server, route Route[T, B]) (*openapi3
...
-			requestBody := newRequestBody[B](bodyTag, []string{"application/json", "application/xml"})
-			s.OpenApiSpec.Components.RequestBodies[bodyTag.name] = &openapi3.RequestBodyRef{
-				Value: requestBody,
-			}
+		if bodyTag.Name != "unknown-interface" {
+			requestBody := newRequestBody[B](bodyTag, route.AcceptedContentTypes)
 
 			// add request body to operation
 			route.Operation.RequestBody = &openapi3.RequestBodyRef{
-				Ref:   "#/components/requestBodies/" + bodyTag.name,
```
dylanhitt, line 184: So are we just electing to not use the references moving foward?

EwenQuim: Simply not the requestBody references, but still other references.

Note that it's just because I want to ship Variadic route options, it's been a month and I want to move forward. We'll probably reuse requestBodies refs later but I need to think about a good way to do so.

dylanhitt: That is fair.

I think I may give a proof concept after I add the global request type feature. 

Cheers!

dylanhitt (approved):

option/option.go
EwenQuim: > FYI I do plan on adding the server wide default option once this is done. Hence why I'm so fixated on this 😄

Veeery good idea. For the moment, we'll ship the variadic options without this feature because I want to move forward, but I'll raise an issue to track this

EwenQuim:
As always thanks @dylanhitt and @ccoVeille for your reviews !

PR #177, opened 2024-08-29
feat: add route.RequestContentType

### LINKED ISSUE #159: Spec should not assume xml responses
Fuego is generating xml schema for all routes do to: https://github.com/go-fuego/fuego/blob/main/openapi.go#L171.

Assuming XML for every route is not great as to support XML routes in go requires a fair amount of work on `xml:` tags on a struct to ensure XML encoding doesn't error.

We should only ever assume json when not specified, and allow for user input when a user wants the route to accept something different.

**Screenshots**

v0.14.0

v1.22.6

I figure something like

`Route.RequestContent(...string)` and `Route.ResponseContent(...string)`

and we default to json when nothing is provided.

Maybe later we add the ability to default the whole server to XML but want to keep the initial case scoped. How does that sound? I'm open to better naming as well.

dylanhitt:
closes #159

Provides the ability for the user to explicitly
set what content types and request body can be.
Maintains the defaullt of `application/json` and `application/xml`

### DIFF
--- a/examples/petstore/controllers/pets.go
+++ b/examples/petstore/controllers/pets.go
@@ -19,6 +19,8 @@ func (rs PetsRessources) Routes(s *fuego.Server) {
 	fuego.Get(petsGroup, "/{id}", rs.getPets)
 	fuego.Get(petsGroup, "/by-name/{name...}", rs.getPetByName)
 	fuego.Put(petsGroup, "/{id}", rs.putPets)
+	fuego.Put(petsGroup, "/{id}/json", rs.putPets).
+		RequestContentType("application/json")
 	fuego.Delete(petsGroup, "/{id}", rs.deletePets)
 }
 
--- a/openapi_operations.go
+++ b/openapi_operations.go
@@ -92,6 +92,23 @@ func (r Route[ResponseBody, RequestBody]) Tags(tags ...string) Route[ResponseBod
 	return r
 }
 
+// Replace the available request Content-Types for the route.
+// By default, the request Content-Types are `application/json` and `application/xml`
+func (r Route[ResponseBody, RequestBody]) RequestContentType(consumes ...string) Route[ResponseBody, RequestBody] {
+	bodyTag := schemaTagFromType(r.mainRouter, *new(RequestBody))
+
+	if bodyTag.name != "unknown-interface" {
+		requestBody := newRequestBody[RequestBody](bodyTag, consumes)
+
+		// set just Value as we do not want to reference
+		// a global requestBody
+		r.Operation.RequestBody = &openapi3.RequestBodyRef{
+			Value: requestBody,
+		}
+	}
+	return r
+}
+
 // AddTags adds tags to the route.
 func (r Route[ResponseBody, RequestBody]) AddTags(tags ...string) Route[ResponseBody, RequestBody] {
 	r.Operation.Tags = append(r.Operation.Tags, tags...)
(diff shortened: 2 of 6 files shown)


openapi_operations.go
```diff
@@ -92,6 +93,27 @@ func (r Route[ResponseBody, RequestBody]) Tags(tags ...string) Route[ResponseBod
 	return r
 }
 
+// Replace the available request Content-Types for the route.
+// By default, the request Content-Types are `application/json` and `application/xml`
+func (r Route[ResponseBody, RequestBody]) RequestContentType(consumes ...string) Route[ResponseBody, RequestBody] {
+	bodyTag := schemaTagFromType(r.mainRouter, *new(RequestBody))
+
+	if bodyTag.name != "unknown-interface" {
+		content := openapi3.NewContentWithSchemaRef(&bodyTag.SchemaRef, consumes)
```
dylanhitt, line 102: Could maybe save a few line by consolidating these into a helper function 🤷‍♂️

EwenQuim (approved):
Nice, thank you very much!

examples/petstore/controllers/pets.go
```diff
@@ -19,6 +19,8 @@ func (rs PetsRessources) Routes(s *fuego.Server) {
 	fuego.Get(petsGroup, "/{id}", rs.getPets)
 	fuego.Get(petsGroup, "/by-name/{name...}", rs.getPetByName)
 	fuego.Put(petsGroup, "/{id}", rs.putPets)
+	fuego.Put(petsGroup, "/{id}/json", rs.putPets).
+		RequestContentType("application/json")
```
ccoVeille, line 23: maybe 


```suggestion
		ExpectContentType("application/json")
```

or


```suggestion
		RequireContentType("application/json")
```

or


```suggestion
		AcceptContentType("application/json")
```

What do you think?

openapi_operations.go
```diff
@@ -92,6 +92,23 @@ func (r Route[ResponseBody, RequestBody]) Tags(tags ...string) Route[ResponseBod
 	return r
 }
 
+// Replace the available request Content-Types for the route.
+// By default, the request Content-Types are `application/json` and `application/xml`
+func (r Route[ResponseBody, RequestBody]) RequestContentType(consumes ...string) Route[ResponseBody, RequestBody] {
+	bodyTag := schemaTagFromType(r.mainRouter, *new(RequestBody))
+
+	if bodyTag.name != "unknown-interface" {
```
ccoVeille, line 100: I would return r when equal, avoiding a function wide `if` statement

openapi_operations_test.go
```diff
@@ -61,6 +61,35 @@ func TestHeaderParams(t *testing.T) {
...
+		route := Post(s, "/test", testControllerWithBody).
+			RequestContentType("application/json")
+
+		content := route.Operation.RequestBody.Value.Content
+		require.NotNil(t, content.Get("application/json"))
+		require.Nil(t, content.Get("application/xml"))
+		require.Equal(t, "#/components/schemas/TestRequestBody", content.Get("application/json").Schema.Ref)
+		_, ok := s.OpenApiSpec.Components.RequestBodies["TestRequestBody"]
+		require.True(t, ok)
+	})
```
ccoVeille, line 76: require.Contains could be used I think

Something like


```suggestion
		require.Contains(t, s.OpenApiSpec.Components.RequestBodies, "TestRequestBody")
	})
```

examples/petstore/controllers/pets.go
dylanhitt: Not sure I'm following?

We're applying the Content-Type for the openapi RequestBody.

ccoVeille: I assumed the method was expecting the caller to send this.

I might have misunderstood the PR purpose

PR #178, opened 2024-08-30
Fix package name for Generated Controllers

### LINKED ISSUE #163: Controller package generation through the CLI creates a file in a package `controller` instead of `controllers`, in the `controllers` folder
**To Reproduce**

1. Run `go run github.com/go-fuego/fuego/cmd/fuego@latest controller books`
2. Try to follow documentation steps.
3. Get an error because package controllers does not exist.

**Expected behavior**
- Package controllers should exist.
- We should be able to call `controllers.BooksRessources{}.Routes(s)` without having to change the package name inside of the generated file.

**Screenshots**
N/A

Fuego version before posting):** `v0.14.0`

**Go version (please check if it happens with the last Go version before posting):**  `go1.22.6 darwin/amd64`

**Additional context**
The docs are partially correct because of this; we need to make a decision whether to keep controllers vs controller and ensure the docs follow this naming convention. If we decide to keep `controller` as the package name, I'd suggest renaming the folder and we still need to update the example in the docs accordingly.

asahasrabuddhe:
Fixes #163

### DIFF
--- a/cmd/fuego/commands/controller.go
+++ b/cmd/fuego/commands/controller.go
@@ -37,7 +37,7 @@ func Controller() *cli.Command {
 
 // createController creates a new controller file
 func createController(controllerName string) (string, error) {
-	controllerDir := "./controllers/"
+	controllerDir := "./controller/"
 	if _, err := os.Stat(controllerDir); os.IsNotExist(err) {
 		err = os.Mkdir(controllerDir, 0o755)
 		if err != nil {
--- a/cmd/fuego/commands/controller.go
+++ b/cmd/fuego/commands/controller.go
@@ -53,8 +53,7 @@ func createController(controllerName string) (string, error) {
 	t := language.English
 	titler := cases.Title(t)
 
-	newContent := strings.ReplaceAll(string(templateContent), "package controller", "package controllers")
-	newContent = strings.ReplaceAll(newContent, "newController", controllerName)
+	newContent := strings.ReplaceAll(string(templateContent), "newController", controllerName)
 	newContent = strings.ReplaceAll(newContent, "NewController", titler.String(controllerName))
 
 	controllerPath := fmt.Sprintf("%s%s.go", controllerDir, controllerName)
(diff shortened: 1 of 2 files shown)


asahasrabuddhe:
@EwenQuim do you mind reviewing this once you have a moment?

EwenQuim (changes requested):

EwenQuim:
Sorry my comment was left pending

cmd/fuego/commands/controller.go
```diff
@@ -53,7 +53,8 @@ func createController(controllerName string) (string, error) {
 	t := language.English
 	titler := cases.Title(t)
 
-	newContent := strings.ReplaceAll(string(templateContent), "newController", controllerName)
+	newContent := strings.ReplaceAll(string(templateContent), "package controller", "package controllers")
```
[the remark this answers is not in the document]
asahasrabuddhe: Sure

cmd/fuego/templates/controller/controller.go
```diff
@@ -1,4 +1,4 @@
-package controller
+package controllers
```
EwenQuim, line 1: I'm really sorry but could you change the folder name instead of the package name ?

Singular is preffered in Go

EwenQuim (changes requested):

cmd/fuego/templates/controller/controller.go
asahasrabuddhe: Do you mean the name of the generated folder?

asahasrabuddhe: I am happy to do this, but we must consider that this will be a breaking change for anyone who has an existing `fuego` project. The updated CLI would write new controllers in the `controller` folder, while the existing folder would be called `controllers`.

EwenQuim: Yes ! People will rename their folder, it's not costly

The other way around was also a breaking change

asahasrabuddhe: Done!

EwenQuim (approved):

PR #181, opened 2024-08-30
Correctly log status code

### LINKED ISSUE #180: ErrorHandler does not print status correctly
**Actual behavior**
When the `ErrorHandler` logs the error message, the status field is incorrectly outputted as `!ERROR:json: unsupported type: func() int`. See the full log message below:
```
{
    "time": "2024-08-29T13:08:15.93402122Z",
    "level": "ERROR",
    "msg": "Error Internal Server Error",
    "status": "!ERROR:json: unsupported type: func() int",
    "detail": "",
    "error": ""
}
```

**Expected behavior**
See below output
```
{
    "time": "2024-08-29T13:08:15.93402122Z",
    "level": "ERROR",
    "msg": "Error Internal Server Error",
    "status": 500,
    "detail": "",
    "error": ""
}
```

`1.23`

MaikelH:
Closes #180 

Call the `StatusCode` function instead of passing it as a pointer.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -130,7 +130,7 @@ func ErrorHandler(err error) error {
 		errResponse.Title = http.StatusText(errResponse.Status)
 	}
 
-	slog.Error("Error "+errResponse.Title, "status", errResponse.StatusCode, "detail", errResponse.Detail, "error", errResponse.Err)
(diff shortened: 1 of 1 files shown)


EwenQuim:
Thanks ! I'll add a test in your PR, it's not normal that we have this behaviour passing our test suite

EwenQuim (approved):

PR #194, opened 2024-09-13
feat: add 500 error

mugraph:
Awesome project :tada: 

While trying it out I noticed the 500 error type was perhaps forgotten to be implemented. At least per the [default errors section](https://go-fuego.github.io/fuego/docs/guides/errors#default-errors), there should be an `InternalServerError` type as well. If there was a decision to not include it, we should take it out from the docs ofc.


errors.go
```diff
@@ -106,6 +106,17 @@ func (e ConflictError) StatusCode() int { return http.StatusConflict }
...
+// InternalServerError is an error used to return a 500 status code.
+type InternalServerError HTTPError
+
+var _ ErrorWithStatus = InternalServerError{}
+
+func (e InternalServerError) Error() string { return e.Err.Error() }
+
+func (e InternalServerError) StatusCode() int { return http.StatusInternalServerError }
+
+func (e InternalServerError) Unwrap() error { return HTTPError(e) }
```
ccoVeille, line 118: HTTPError supports internal error in StatusCode, so an httperror is directly an internal error

But by looking at the implementation I think the issue the fact HTTPError.Error should be modified to call HTTPError.StatusCode() and not HTTPError.Code

```diff
func (e HTTPError) Error() string {
        title := e.Title
+       code := e.StatusCode()
        if title == "" {
-                title = http.StatusText(e.Status)
+                title = http.StatusText(code)
                if title == "" {
                        title = "HTTP Error"
                }
        }
-        return fmt.Sprintf("%s (%d): %s", title, e.Status, e.Detail)
+        return fmt.Sprintf("%s (%d): %s", title, code, e.Detail)
}
```

EwenQuim:
Merged into a temporary branch. I'm editing the PR with changes proposed by @ccoVeille and merging into main just after.

Thanks for the suggestion!

And yes, by default `HTTPError` is an `InternalServerError`, so I just defined `InternalServerError` as a type alias :)

PR #200, opened 2024-10-12
Better Accept Serialization

EwenQuim:
This PR does 2 things:

- explicitely declare te Accept Header that we use in Fuego, I don't know why I didn't do it earlier
- Treat serialization errors as a 406 Not Acceptable error

### DIFF
--- a/serialization.go
+++ b/serialization.go
@@ -213,9 +213,22 @@ func SendJSONError(w http.ResponseWriter, _ *http.Request, err error) {
 
 // SendXML sends a XML response.
 // Declared as a variable to be able to override it for clients that need to customize serialization.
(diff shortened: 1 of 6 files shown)


serialization.go
```diff
@@ -138,13 +149,19 @@ func SendYAMLError(w http.ResponseWriter, _ *http.Request, err error) {
 
 // SendJSON sends a JSON response.
 // Declared as a variable to be able to override it for clients that need to customize serialization.
+// If serialization fails, it does NOT write to the response writer. It has to be passed to SendJSONError.
 var SendJSON = func(w http.ResponseWriter, _ *http.Request, ans any) error {
```
dylanhitt, line 153: do we want to to xml as well?

EwenQuim: Yes indeed, will be working on it tonight

PR #202, opened 2024-10-14
Fix data race due to global instance of openapi3gen.Generator

### LINKED ISSUE #201: Data race due to global instance of openapi3gen.Generator
**To Reproduce**
Create multiple instance of `fuego.Server` in different goroutines.
I have create a simple repository to showcase the problem over [here](https://github.com/loikg/fuego-datarace-reproduce). But for sake of completeness here the test used:

```go
package main

import (
	"fmt"
	"testing"

	"github.com/go-fuego/fuego"
)

func TestDataRace(t *testing.T) {
	t.Parallel()

	for n := range 2 {
		t.Run(fmt.Sprintf("create server %d", n), func(t *testing.T) {
			t.Parallel()
			createFuegoServer(t)
		})
	}
}

func createFuegoServer(t *testing.T) {
	t.Helper()

	server := fuego.NewServer()
	fuego.Get(server, "/", func(ctx *fuego.ContextNoBody) (any, error) {
		return nil, nil
	})
}
```

**Expected behavior**
The data race should not happen.

**Framework version:**
Tested on v0.14.0 and main at the time of writing ([f386d4f](https://github.com/go-fuego/fuego/commit/f386d4f4842ec3b49eb99e26d5df0b04b8706bd6)).

**Go version:**
v1.23.2

**Additional context**
This prevent running tests that exercise the entire server in parallel, but does not affect unit test of handlers as far as I understand.
The root cause is the package global variable `generator` at line  [151](https://github.com/go-fuego/fuego/blob/f386d4f4842ec3b49eb99e26d5df0b04b8706bd6/openapi.go#L151) in openapi.go being shared and concurrently used by every `fuego.Server` instances.
See #202 for a potential fix.

Huge thanks for your work, I really enjoy using fuego!

loikg:
fixes #201

To prevent data races when creating multiple instances of `fuego.Server` in tests for example. I moved the `openapi3gen.Generator` package global to the `fuego.Server` struct. 
Hopefully that make sense.

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -16,7 +16,6 @@ import (
 	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
-	"github.com/getkin/kin-openapi/openapi3gen"
 )
 
 func NewOpenApiSpec() openapi3.T {
@@ -147,10 +146,6 @@ func validateSwaggerUrl(swaggerUrl string) bool {
 	swaggerUrlRegexp := regexp.MustCompile(`^\/[\/a-zA-Z0-9\-\_]+[a-zA-Z0-9\-\_]$`)
 	return swaggerUrlRegexp.MatchString(swaggerUrl)
 }
-
-var generator = openapi3gen.NewGenerator(
-	openapi3gen.UseAllExportedFields(),
-)
 
 // RegisterOpenAPIOperation registers an OpenAPI operation.
 func RegisterOpenAPIOperation[T, B any](s *Server, route Route[T, B]) (*openapi3.Operation, error) {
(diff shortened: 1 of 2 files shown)


EwenQuim:
That makes sense, thank you for your contribution!

I love to have tests so the change you proposed is resistant to future internal refactorings. If you don't provide some tests, it might break in the future (we'll try not to, of course, but there's less guarantees). If you don't have time or you don't want to bother writing some tests, that's fine, we understand your time is valuable :) Just tell me and I'll review it without tests.

Thank you again for your contribution !

loikg:
Sure, it make sense I will add a test to cover this in the coming days, but it will require adding the `-race` flag to the `go test` command  in the Makefile to actually detect the race condition. I hope it make sense. Thanks!

EwenQuim:
I'm merging to make a new release but please consider adding some tests, it's safer for everyone that way :)

PR #203, opened 2024-10-15
Remove Group Parameters

EwenQuim:
This PR aims to remove Group Parameters in favor of reusable Route Options. While it seems to be a regression, it is more in the spririt of Go:

- Group parameters are a kind of heritance feature
- Route options are more a "composable" approach
- Simpler code to maintain. In the next release, I will focus on making Fuego's code more simple & elegant

@dylanhitt could you please review it?


serialization_test.go
```diff
@@ -64,6 +65,8 @@ func TestRecursiveJSON(t *testing.T) {
 		err := SendJSON(w, nil, value)
 
 		require.Error(t, err)
+		var unsupportedValueError *json.UnsupportedValueError
```
dylanhitt, line 68: Is this left over from #200? Not sure how it relates to this PR

EwenQuim: Yes sorry

EwenQuim:
> If the plan is to eventually add the options style configuration on to the Server/Group. I'd be 100% on board.

Yes I hesitated, we might let the final decision to the users.

We might do this and **keep the same signature** as the route options, which will not be trivial.

EwenQuim:
Closed in favor of #233

PR #206, opened 2024-10-21
fixed wrong schema generation when using embedded struct

### LINKED ISSUE #195: wrong schema generation when using struct embedding
**To Reproduce**

```go
package main

import "github.com/go-fuego/fuego"

type Inner struct {
	InnerField string `validate:"required"`
}

type Outer struct {
	Inner
	OuterField string `validate:"required"`
}

func main() {
	s := fuego.NewServer()

	fuego.Get(s, "/", func(c fuego.ContextNoBody) (Outer, error) {
		return Outer{}, nil
	})

	s.Run()
}
```

**Expected behavior**
Both `InnerField` and `OuterField` are marked as required. Instead only, the `OuterField` keeps its "required" tag.

**Screenshots**
Screenshot 2024-09-29 at 1 04 18

**Framework version:** v0.14.0

**Go version:** 1.23.1

**Additional context**
Huge thanks for your framework, it's really well designed <3

mrguamos:
fixes #195

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -336,6 +336,13 @@ func (s *Server) parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) 
 
 	for i := range t.NumField() {
 		field := t.Field(i)
+		fieldType := field.Type
+
+		if field.Anonymous {
+			s.parseStructTags(fieldType, schemaRef)
+			continue
+		}
+
 		jsonFieldName := field.Tag.Get("json")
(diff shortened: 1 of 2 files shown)


EwenQuim (commented):
Thank you very much ! Is it possible to add a quick test case to make sure it doesn't happen in the future?

mrguamos:
> Thank you very much ! Is it possible to add a quick test case to make sure it doesn't happen in the future?

I added a test case. can you check?

EwenQuim:
Thank you very much!

EwenQuim (approved):

PR #215, opened 2024-10-29
Moved Options & Params to main package

EwenQuim:
Functions in param and option packages are aliases to the ones in the main package.
This allows to have cleaner code and not export variables that should only be used internally


dylanhitt (approved):
This is one of thos PRs that makes a person smile. Nice 👍

PR #220, opened 2024-11-01
Perf/url.values

dylanhitt:
Stores `url.Values` in context to avoid unneeded calls to `URL.Query()`

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -101,6 +101,7 @@ func NewContext[B any](w http.ResponseWriter, r *http.Request, options readOptio
 				DisallowUnknownFields: options.DisallowUnknownFields,
 				MaxBodySize:           options.MaxBodySize,
 			},
+			urlValues: r.URL.Query(),
 		},
 	}
 
@@ -124,7 +125,8 @@ type ContextNoBody struct {
 	fs        fs.FS
 	templates *template.Template
 
-	params map[string]OpenAPIParam // list of expected query parameters (declared in the OpenAPI spec)
+	params    map[string]OpenAPIParam // list of expected query parameters (declared in the OpenAPI spec)
+	urlValues url.Values
 
 	readOptions readOptions
 }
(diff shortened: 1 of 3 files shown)


ctx.go
```diff
@@ -271,7 +273,7 @@ func (c ContextNoBody) QueryParamArr(name string) []string {
 	if !ok {
 		slog.Warn("query parameter not expected in OpenAPI spec", "param", name)
 	}
-	return c.Req.URL.Query()[name]
+	return c.urlValues[name]
```
ccoVeille, line 276: I know the code was there.

But why not calling Get here ?


```suggestion
	return c.urlValues.Get(name)
```


As the code does a few line below?

ctx.go
```diff
@@ -288,12 +290,11 @@ func (c ContextNoBody) QueryParam(name string) string {
 		slog.Warn("query parameter not expected in OpenAPI spec", "param", name, "expected_one_of", c.params)
 	}
 
-	_, found := c.Req.URL.Query()[name]
-	if !found {
+	if !c.urlValues.Has(name) {
 		defaultValue, _ := c.params[name].Default.(string)
```
ccoVeille, line 294: It's not a code you are updating.

But I'm worried about a possible panic if c.params[name] is not defined

ctx.go
dylanhitt: Returning the whole array. This pattern was discussed in a previous PR where I asked the same question xD

ctx.go
dylanhitt: Yeah I agree that this code looks odd for that. I don’t believe it’s actually possible though now with #217. The validation there should handle this.

It would be a bit nice looking to handle it here I guess, but it wouldn’t belong with this PR

ctx.go
dylanhitt: See #169

ccoVeille: Thanks for your reply.

It's clearer

ctx.go
ccoVeille: Yes, it is definitely not related to this PR.

But I raised the point, because it was weird to me

EwenQuim (approved):
LGTM !

ctx.go
ccoVeille: Do you plan to open an issue for this?

dylanhitt: Ehhh, I’m neither here nor there. Again it’s not really possible to panic, but it would make the code look less weird I guess.

ccoVeille: OK. A crime without victim then 😅😁

PR #226, opened 2024-11-16
added securityScheme

simonemastella:
# Add WithSecurity Server Option and Route Security Documentation
## Description
Adds a new `WithSecurity` server option to configure security schemes in the OpenAPI specification and documents the route-level `.Security()` method. This change improves the API by providing a complete security configuration solution at both the server and route levels.

### Server-Level Security Configuration
- New `WithSecurity` function that accepts a map of security schemes
- Supports all OpenAPI security scheme types (HTTP, API Key, OAuth2, etc.)
- Safe initialization of SecuritySchemes map
- Allows multiple security schemes to be configured simultaneously

### Route-Level Security
- Enhanced documentation for `.Security(requirementName string)` method
- Allows per-route security requirements specification
- Links routes to security schemes defined at the server level

### Server Security Configuration
```go
s := NewServer(
    WithSecurity(openapi3.SecuritySchemes{
        "bearerAuth": &openapi3.SecuritySchemeRef{
            Value: openapi3.NewSecurityScheme().
                WithType("http").
                WithScheme("bearer").
                WithBearerFormat("JWT").
                WithDescription("Enter your JWT token in the format: Bearer <token>"),
        },
    }),
)
```
### Route Security Configuration
```go
goCopyGet(s, "/protected", handler).
    Security("bearerAuth").  // Requires JWT authentication
    Summary("Protected endpoint")
```
## Benefits

- Complete security solution: Configure security at both server and route levels
- More maintainable: Security schemes can be defined and updated in one place
- More flexible: Easy to add or modify security requirements
- Better API design: Follows the pattern of other server options
- Improved developer experience: Makes security configuration more explicit and easier to understand
- Fine-grained control: Apply different security requirements to different routes

## Breaking Changes
None. This is a new feature that maintains backward compatibility.

## Example 
image


ccoVeille (approved):

openapi_operations.go
```diff
@@ -86,6 +86,18 @@ func (r Route[ResponseBody, RequestBody]) Param(paramType ParamType, name, descr
 	return r
 }
 
+// Security adds a security requirement to the route.
+// It configures the route to require authentication using the specified security scheme.
+// The requirementName parameter should match a security scheme defined in the OpenAPI components section.
+// For example, if you have a JWT bearer auth scheme named "bearerAuth" in your components,
+// you would call Security("bearerAuth") to require JWT authentication for this route.
+func (r Route[ResponseBody, RequestBody]) Security(requirementName string) Route[ResponseBody, RequestBody] {
```
dylanhitt, line 94: This builder style pattern is ultimately deprecated. I don't think we'd want to add a new feature here. Would you mind implementing this in the style that is `option.go` [here is an example](https://github.com/go-fuego/fuego/blob/main/option.go#L204) and it's subsequent entry into the [`option` package](https://github.com/go-fuego/fuego/blob/main/option/option.go#L79)

dylanhitt (changes requested):
Thanks for the contribution. Would you mind making the requested changes?

openapi_operations.go
EwenQuim: Yes, that's the only comment I have too. Thanks for making this PR!

simonemastella:
> Thanks for the contribution. Would you mind making the requested changes?
> 
> Also a reminder to please add tests! Thank you

Yeah I'll work on it but this is a busy week and it might require a couple of days since I would like to add the possibility to combine multiple requirements: (requirement1 OR requirement2) AND requirement3

It might make sense to add the scope of a requirement as well https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#security-requirement-object

I will also add some test to cover the feature!

openapi_operations.go
simonemastella: Thanks for the hint, sorry for the mistake, I am new to golang and also to fuego.
I explored security schemes and included the ability to combine multiple requirements to get the ability to put them in and / or with each other and also to assign them a scope. 

### Support for Single Security Requirements:
Allows adding a single security scheme, e.g., `basic` authentication:
```go
fuego.OptionSecurity(openapi3.SecurityRequirement{
    "basic": []string{},
})
```
### Support for Scopes:
Handles schemes with specific **scopes**, like `oauth2`:
```go
fuego.OptionSecurity(openapi3.SecurityRequirement{
    "oauth2": []string{"read:users", "write:users"},
})
```
### AND Logic Within a Single Requirement:

Combines multiple schemes in one `SecurityRequirement` for **AND** logic:
```go
fuego.OptionSecurity(openapi3.SecurityRequirement{
    "basic":  []string{},
    "oauth2": []string{"read:users"},
})
```
### OR Logic Across Multiple Requirements:

Adds multiple SecurityRequirement entries for **OR** logic:
```go
fuego.OptionSecurity(
    openapi3.SecurityRequirement{"basic": []string{}},
    openapi3.SecurityRequirement{"oauth2": []string{"read:users"}},
)
```
### Handles Complex Combinations:

Supports both **AND** and **OR** in the same configuration:
```go
fuego.OptionSecurity(
    openapi3.SecurityRequirement{
        "basic":  []string{},
        "oauth2": []string{"read:users"},
    },
    openapi3.SecurityRequirement{
        "apiKey": []string{},
    },
)
```

dylanhitt (commented):
We will 100% need a docs update for this (it doesn't have to exist in this PR if you don't have the time. I thank you for the contribution enough).

That being said, and looking at this pretty late. This looks fantastic. To play devils advocate. What are the thoughts on making the OR logic be a new `OptionSecurity` entry on the route. That might possibly simplify the interface/make it more clear. 

Again I didn't look closely and won't have the time til Friday. But that's my initial feedback. I'll more closely friday unless another team member beats me to it.

dylanhitt:
Was able to get to this today. 

Disregard my comment as I like this. I realized OR logic with separate calls to `OptionSecurity` is implemented with this. 

Final thoughts:
1. Should we add some validation in `OptionSecurity` to ensure the schema being inputted is valid schema. If not we panic/exit?
2. Could you add an example of using this into the `examples/petstore`? This will update our golden file open api spec. It helps us spot unwanted changes to spec outputs, kinda an integration test of sorts.

option/option.go
```diff
@@ -76,6 +76,9 @@ var Summary = fuego.OptionSummary
 // Description adds a description to the route.
 var Description = fuego.OptionDescription
 
+// Security adds a list of openapi3.SecurityRequirement to the route.
+var Security = fuego.OptionSecurity
```
dylanhitt, line 80: Would you mind enhancing this bit. Explaining out the `AND`/`OR` logic it a bit?

option_test.go
```diff
@@ -402,3 +403,151 @@ func TestHide(t *testing.T) {
...
+				},
+			),
+		)
+
+		require.NotNil(t, route.Operation.Security)
+		require.Len(t, *route.Operation.Security, 1)
+		require.Equal(t,
+			[]string{"read:users", "write:users"},
+			(*route.Operation.Security)[0]["oauth2"],
+		)
```
ccoVeille, line 444:
```suggestion
		require.NotNil(t, route.Operation.Security)
		require.Len(t, *route.Operation.Security, 1)
		require.Contains(t, *route.Operation.Security[0], "oauth2")
		require.Equal(t,
			[]string{"read:users", "write:users"},
			(*route.Operation.Security)[0]["oauth2"],
		)
```

option/option.go
simonemastella: I have added this, what do you think?
```
// OptionSecurity configures security requirements to the route.
//
// Single Scheme (AND Logic):
//
//	Add a single security requirement with multiple schemes.
//	All schemes must be satisfied:
//	OptionSecurity(openapi3.SecurityRequirement{
//	  "basic": [],        // Requires basic auth
//	  "oauth2": ["read"]  // AND requires oauth with read scope
//	})
//
// Multiple Schemes (OR Logic):
//
//	Add multiple security requirements.
//	At least one requirement must be satisfied:
//	OptionSecurity(
//	  openapi3.SecurityRequirement{"basic": []},        // First option
//	  openapi3.SecurityRequirement{"oauth2": ["read"]}  // Alternative option
//	})
//
// Mixing Approaches:
//
//	Combine AND logic within requirements and OR logic between requirements:
//	OptionSecurity(
//	  openapi3.SecurityRequirement{
//	    "basic": [],             // Requires basic auth
//	    "oauth2": ["read:user"]  // AND oauth with read:user scope
//	  },
//	  openapi3.SecurityRequirement{"apiKey": []}  // OR alternative with API key
//	})
```

option_test.go
```diff
@@ -402,3 +403,151 @@ func TestHide(t *testing.T) {
...
+				openapi3.SecurityRequirement{
+					"basic":  []string{},
+					"oauth2": []string{"read:users"},
+				},
+			),
+		)
+
+		require.NotNil(t, route.Operation.Security)
+		require.Len(t, *route.Operation.Security, 1)
+		require.Contains(t, (*route.Operation.Security)[0], "basic")
```
ccoVeille, line 461: You could check is empty


```suggestion
		require.Contains(t, (*route.Operation.Security)[0], "basic")
		require.Empty(t, (*route.Operation.Security)[0]["basic"])
```

option_test.go
```diff
@@ -402,3 +403,151 @@ func TestHide(t *testing.T) {
...
+				},
+				openapi3.SecurityRequirement{
+					"oauth2": []string{"read:users"},
+				},
+			),
+		)
+
+		require.NotNil(t, route.Operation.Security)
+		require.Len(t, *route.Operation.Security, 2)
+		require.Contains(t, (*route.Operation.Security)[0], "basic")
```
ccoVeille, line 482: Same, about checking basic is empty

option_test.go
```diff
@@ -402,3 +403,151 @@ func TestHide(t *testing.T) {
...
+					"apiKey": []string{},
+				},
+			),
+		)
+
+		require.NotNil(t, route.Operation.Security)
+		require.Len(t, *route.Operation.Security, 2)
+
+		// Check first requirement (basic AND oauth2)
+		require.Contains(t, (*route.Operation.Security)[0], "basic")
```
ccoVeille, line 506: Same

option_test.go
```diff
@@ -402,3 +403,151 @@ func TestHide(t *testing.T) {
...
+	t.Run("empty security options", func(t *testing.T) {
+		s := fuego.NewServer()
+
+		route := fuego.Get(s, "/test", helloWorld,
+			fuego.OptionSecurity(),
+		)
+
+		require.NotNil(t, route.Operation.Security)
+		require.Len(t, *route.Operation.Security, 0)
+		require.Empty(t, (*route.Operation.Security))
```
ccoVeille, line 524:
```suggestion
		require.Empty(t, (*route.Operation.Security))
```

simonemastella:
> Oh sorry for not being clear. I just meant “ensure the security schema being asked on the route exists in the security schemas on the server.” Just check the map if not panic so we can tell the user early.
> 
> 😅 sorry saying validate something in terms of OpenAPI can be super ambiguous.

ahahah perfect, do you think that something like this could be enough?
p.s. by the end of the week I should be able to modify the `examples/petstore`
```go
                // Validate input is not empty
		if len(securityRequirements) == 0 {
			panic("at least one security requirement must be provided")
		}

		// Validate the security scheme exists in components
		for _, req := range securityRequirements {
			for schemeName := range req {
				if r.OpenAPI != nil && r.OpenAPI.Components != nil {
					if _, exists := r.OpenAPI.Components.SecuritySchemes[schemeName]; !exists {
						panic(fmt.Sprintf("security scheme '%s' not defined in components", schemeName))
					}
				}
			}
		}
```

dylanhitt:
Yes, you can remove the nil checks as well as everything is new'd up when the server is created. 

One thing that is interesting. If we moved away from the varadic input to just a single input we wouldn't need the extra for loop

option.go
```diff
@@ -255,6 +256,18 @@ func OptionHide() func(*BaseRoute) {
 
 func OptionSecurity(securityRequirements ...openapi3.SecurityRequirement) func(*BaseRoute) {
 	return func(r *BaseRoute) {
+
+		// Validate the security scheme exists in components
+		for _, req := range securityRequirements {
+			for schemeName := range req {
+				if r.mainRouter.OpenApiSpec.Components != nil {
```
dylanhitt, line 263: Is it really possible for this to be nil? And in the case that it is would really want the `SecurityRequirement`s to be valid

dylanhitt: Right, so maybe something like this would look a bit neater

```golang
func OptionSecurity(securityRequirements ...openapi3.SecurityRequirement) func(*BaseRoute) {
	return func(r *BaseRoute) {
               if r.mainRouter.OpenApiSpec.Components == nil {
                   panic("zero security schemas have been registered with the server")
               }

		// Validate the security scheme exists in components
		for _, req := range securityRequirements {
			for schemeName := range req {
				if _, exists := r.mainRouter.OpenApiSpec.Components.SecuritySchemes[schemeName]; !exists {
					panic(fmt.Sprintf("security scheme '%s' not defined in components", schemeName))
				}
			}
		}
```

and catch the case we're talking about.

Also I hope you're enjoying go 😄

option/option.go
```diff
@@ -76,6 +76,38 @@ var Summary = fuego.OptionSummary
 // Description adds a description to the route.
 var Description = fuego.OptionDescription
 
+// OptionSecurity configures security requirements to the route.
```
ccoVeille, line 79:
```suggestion
// Security configures security requirements to the route.
```

You renamed it

option/option.go
```diff
@@ -76,6 +76,38 @@ var Summary = fuego.OptionSummary
...
+// Mixing Approaches:
+//
+//	Combine AND logic within requirements and OR logic between requirements:
+//	OptionSecurity(
+//	  openapi3.SecurityRequirement{
+//	    "basic": [],             // Requires basic auth
+//	    "oauth2": ["read:user"]  // AND oauth with read:user scope
+//	  },
+//	  openapi3.SecurityRequirement{"apiKey": []}  // OR alternative with API key
+//	})
```
ccoVeille, line 108: Same

```suggestion
//	Security(openapi3.SecurityRequirement{
//	  "basic": [],        // Requires basic auth
//	  "oauth2": ["read"]  // AND requires oauth with read scope
//	})
//
// Multiple Schemes (OR Logic):
//
//	Add multiple security requirements.
//	At least one requirement must be satisfied:
//	Security(
//	  openapi3.SecurityRequirement{"basic": []},        // First option
//	  openapi3.SecurityRequirement{"oauth2": ["read"]}  // Alternative option
//	})
//
// Mixing Approaches:
//
//	Combine AND logic within requirements and OR logic between requirements:
//	Security(
//	  openapi3.SecurityRequirement{
//	    "basic": [],             // Requires basic auth
//	    "oauth2": ["read:user"]  // AND oauth with read:user scope
//	  },
//	  openapi3.SecurityRequirement{"apiKey": []}  // OR alternative with API key
//	})
```

option.go
```diff
@@ -252,3 +253,27 @@ func OptionHide() func(*BaseRoute) {
 		r.Hidden = true
 	}
 }
+
+func OptionSecurity(securityRequirements ...openapi3.SecurityRequirement) func(*BaseRoute) {
```
ccoVeille, line 257: Maybe the comment could be added here

ccoVeille (approved):
LGTM 👍, but minor remarks anyway

option/option.go
dylanhitt: Please update this. And this should be good to go. 😄

dylanhitt (changes requested):
I'm a fan of @ccoVeille requested changes. Once updated I'll approve

EwenQuim (approved):

EwenQuim:
I've made the asked modifications myself on `main`.

Thank you a lot @simonemastella for this feature, and @dylanhitt & @ccoVeille for proofreading this! 🙏🏼

PR #230, opened 2024-11-20
chore: panic when OptionAddError is supplied more than one type

dylanhitt:
Opened this for two reasons. First, this PR.

Saw this today figured we should panic or `slog.Error` with `os.exit`?

---

That being said I noticed that with the default `ErrorHandler` we provide it is not possible to return custom error types as we are always returning `HTTPError` although the OpenAPI spec generated would describe the custom type being in the response. There seems to be two options:

1. Change the default to only return `HTTPError` when it adheres to one of our interfaces. The issue with this though is that we no longer default to RFC 9457 errors format; honestly I'm not really sure it really matters as users need to implement an HTTPErrror or one of our interfaces to really get an RFC 9457 output.
2. Document that in order to add customer error types the caller must override the default ErrorHandler. This could be checked at start time as well.

### DIFF
--- a/option.go
+++ b/option.go
@@ -215,9 +215,14 @@ func OptionDeprecated() func(*BaseRoute) {
 }
 
 // AddError adds an error to the route.
+// Required: should only supply one type to `errorType`
 func OptionAddError(code int, description string, errorType ...any) func(*BaseRoute) {
 	var responseSchema SchemaTag
 	return func(r *BaseRoute) {
+		if len(errorType) > 1 {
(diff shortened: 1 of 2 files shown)


EwenQuim:
Throwing is fine as it's made at start time and not per request, thanks for this PR! 

---

Yes it's a good change we can make. I personally only use HTTPError on my projects, that's why I don't work on custom error handling, but I agree we can make it better !

EwenQuim:
Merged with #229

PR #233, opened 2024-11-21
Variadic Group options

EwenQuim:


mux.go
```diff
@@ -40,6 +40,23 @@ func Group(s *Server, path string) *Server {
 	}
 	newServer.mainRouter = s
 
+	baseRoute := BaseRoute{
+		Params:    make(map[string]OpenAPIParam),
+		Operation: openapi3.NewOperation(),
+	}
+	for _, option := range options {
+		option(&baseRoute)
+	}
```
dylanhitt, line 49: nitpick: could save a few lines by creating a helper function for `WithRouteOptions` and this. Would be nice if we ever change anything about this.

dylanhitt (approved):
I didn't realize we were gonna get the reuse of route options here and on the server. 

Are you gonna mark the old funcs as deprecated?

PR #234, opened 2024-11-21
BREAKING: only respond with HTTPError on proper type match

dylanhitt:
Per discussion in #230

No longer only return HTTPError. We only return HTTPError when the caller uses on our internal errors, HTTPError, or they adhere to `ErrorWithStatus` interface.


errors.go
```diff
@@ -125,6 +125,15 @@ func (e NotAcceptableError) Unwrap() error { return HTTPError(e) }
...
+		return handleHTTPError(err)
+	}
+
+	return err
+}
+
+func handleHTTPError(err error) HTTPError {
 	errResponse := HTTPError{
 		Err: err,
 	}
```
ccoVeille, line 139: There is something strange here

If err is already an HTTPError when entering ErrorHandler, you will created an HTTPError of an HTTPError

dylanhitt: I’m not sure that’s how’s errors.As works. I’m not passing the reference of the second argument for a reason. I could maybe make my intention more clear by just checking the type, but I’m fairly certain the tests wouldn’t have passed if what you say is true.

dylanhitt: Oh I see the concern. I’ll write a test case for it and fix the issue. Nice eye

dylanhitt: Wrote a test case today. It's fine actually this code covers that case

```
func handleHTTPError(err error) HTTPError {
	errResponse := HTTPError{
		Err: err,
	}

	var errorInfo HTTPError
	if errors.As(err, &errorInfo) {
		errResponse = errorInfo
	}
```

errors_test.go
```diff
@@ -39,6 +37,18 @@ func TestErrorHandler(t *testing.T) {
...
 
+	t.Run("not duplicate HTTPError", func(t *testing.T) {
+		err := HTTPError{
+			Err: errors.New("HTTPError"),
+		}
+		errResponse := ErrorHandler(err)
+
+		var httpError HTTPError
+		require.ErrorAs(t, errResponse, &httpError)
+		require.False(t, errors.As(httpError.Err, &HTTPError{}))
```
ccoVeille, line 48: You can use this if you bump testify to latest version. This missing helper was added

```suggestion
		require.NotErrorAs(t, httpError.Err, &HTTPError{})
```

examples/petstore/controllers/pets.go
```diff
@@ -19,6 +19,15 @@ type PetsResources struct {
 	PetsService PetsService
 }
 
+type PetsError struct {
+	Err     error  `json:"-" xml:"-"`
+	Message string `json:"message"`
```
ccoVeille, line 24: Maybe you could add the XML tag here too

errors_test.go
dylanhitt: Ah thanks. I only saw .NotErrorIs and just assumed they didn’t do it.

errors.go
```diff
@@ -122,9 +122,21 @@ func (e NotAcceptableError) StatusCode() int { return http.StatusNotAcceptable }
...
-// It transforms any error into the unified error type [HTTPError],
-// Using the [ErrorWithStatus] interface.
+// If the error is an [HTTPError] that is error is returned.
+// If the error adheres to the [ErrorWithStatus] interface
+// the error is transformed to a [HTTPError].
+// If the error is not an [HTTPError] nor does it adhere to an
+// interface the error is returned.
 func ErrorHandler(err error) error {
+	var errorStatus ErrorWithStatus
+	if errors.As(err, &HTTPError{}) || errors.As(err, &errorStatus) {
```
ccoVeille, line 132: You are not using error status

So maybe you could use the same syntax for both error types


```suggestion
	if errors.As(err, &HTTPError{}) || errors.As(err, &ErrorWithStatus{}) {
```

Also, ypu could increase readability by using a switch


```suggestion
	switch {
		case errors.As(err, &HTTPError{}),
			errors.As(err, &ErrorWithStatus{}):
```

dylanhitt: `ErrorWithStatus` is an interface. The code above isn't valid. 

image

EwenQuim (approved):
I'm OK with this behaviour!
We'll highlight it in the next release.

EwenQuim:
I'm sorry I pushed something else to main, you can rebase & merge!

PR #240, opened 2024-11-29
Fix/validate cookies

dylanhitt:


EwenQuim (approved):
Thank you!
I moved params Validation to their own file and replaced the controllers by `dummyController` just for lisibility (when doing params validation it really does not matter what's inside the controller)

dylanhitt:
Yeah, I liked both changes. There’s probably a few places in the tests that we could change to use dummyController

PR #245, opened 2024-11-30
Customizable net.Listener for advanced use cases, such as tunneling

Ignaciojeria:
image


ccoVeille:
No test added?

dylanhitt:
Would you mind following the `With*` pattern as well such as [here](https://github.com/go-fuego/fuego/blob/main/server.go#L205)

dylanhitt:
Also we have two entrypoints `Run` and `RunTLS`. Updating both would be best. 

Also maybe a refactor is in order within the `s.setup` method. We should probably create the default listener in the `Server` constructor, and instead of calling `ListenAndServe*` we just call `Serve`.

EwenQuim (approved):
Thanks a lot, good idea!
If you can make the requested changes, it would be nice.

dylanhitt:
Pardon me. I was on a phone so I was short. 

I will say this is a great contribution. I hope you’re enjoying fuego. 😄

Ignaciojeria:
I've addressed some of the suggestions and pushed a new change. I haven't had the chance to fully test it yet, but I'll be running some tests soon. Any feedback is greatly appreciated!

serve.go
```diff
@@ -14,19 +17,23 @@ import (
 // It returns an error if the server could not start (it could not bind to the port for example).
 // It also generates the OpenAPI spec and outputs it to a file, the UI, and a handler (if enabled).
 func (s *Server) Run() error {
+	if err := s.setupDefaultListener(); err != nil {
+		return fmt.Errorf("failed to start server: %w", err)
+	}
 	s.setup()
```
ccoVeille, line 23: The setup listener is now called before setup, at all place setup is called if I'm not wrong

What about calling it in setup, and update setup signature to return an error?

Ignaciojeria: Thank you! Two heads are better than one. I made some adjustments and have been running manual tests (locally only).

serve.go
```diff
@@ -24,19 +29,86 @@ func (s *Server) Run() error {
 // It also generates the OpenAPI spec and outputs it to a file, the UI, and a handler (if enabled).
 func (s *Server) RunTLS(certFile, keyFile string) error {
 	s.isTLS = true
-
-	s.setup()
-	return s.Server.ListenAndServeTLS(certFile, keyFile)
+	if err := s.setup(certFile, keyFile); err != nil {
+		return err
+	}
```
ccoVeille, line 34: Can't you call WithTLSListener here?

It would allow to pass certFile and keyFile to setup, and the check on isTLS

Ignaciojeria: The code was significantly simplified based on the observation. Can you help check if there’s any room for improvement? Thank you very much!

ccoVeille: Great.

I'll open a PR to your fork

serve.go
```diff
@@ -100,7 +100,7 @@ func (s *Server) setupDefaultListener() error {
 	}
 	addr := s.Server.Addr
 	if addr == "" {
-		addr = "127.0.0.1:9999"
+		addr = ":9999"
```
ccoVeille, line 64: please check TestPetstoreOpenAPIGeneration

it seems broken now

Ignaciojeria: i will check now

Ignaciojeria: "I don't see any problem, but I can add host:port by default.
image

image

image

serve.go
ccoVeille: I opened https://github.com/Ignaciojeria/fuego/pull/1

ccoVeille:
I'm not a fuego maintainer

I opened 
 https://github.com/Ignaciojeria/fuego/pull/1

Please consider it.

serve.go
ccoVeille: I did it in https://github.com/Ignaciojeria/fuego/pull/1

But let me ask you the question here.

Why did you remove the local host? To listen on all interfaces?

Was this changes needed in this exact PR?

server.go
```diff
@@ -68,6 +70,8 @@ type Server struct {
 
 	OpenApiSpec openapi3.T // OpenAPI spec generated by the server
 
+	Listener net.Listener
```
ccoVeille, line 73: You are adding a listener

So it should be closed when server stops

So please add this

```go
func (s *Server) Close () {
	if s.Listener != nil }
		s.Listener.Close()
	}
	s.Server.Close()
	return
}
```

ccoVeille, line 73: I don't see a need for the listener to be public, you added WithListener

server.go
```diff
@@ -125,7 +129,6 @@ func NewServer(options ...func(*Server)) *Server {
 	}
 
 	defaultOptions := [...]func(*Server){
-		WithAddr("localhost:9999"),
```
ccoVeille, line 128: There is a problem with this.

Please take a look at the PR I opened in your fork

Ignaciojeria: PR merged. I removed WithTLSListener because it added unnecessary complexity. I believe RunTLS and WithListener already provide the user with the necessary customization options. Please correct me if I’m wrong.

server.go
```diff
@@ -159,6 +162,17 @@ func NewServer(options ...func(*Server)) *Server {
...
+func getServerAddress(s *Server) string {
+	if s.Listener != nil {
+		return s.Listener.Addr().String()
+	}
+	if s.Server.Addr != "" {
+		return s.Server.Addr
+	}
+	// Default address if none is set
+	return ":9999"
+}
```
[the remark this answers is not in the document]
Ignaciojeria: method removed

server.go
Ignaciojeria: you are right now is private

server.go
ccoVeille: I thought the same. 👍

Also, semantically, I was annoyed that WithTLSListener was not allowing to pass a listener

server.go
```diff
@@ -352,6 +363,18 @@ func WithoutLogger() func(*Server) {
 	}
 }
 
+// WithListener configures the server to use a custom listener.
+// If no listener is provided, it creates a default listener using the server's address.
```
ccoVeille, line 367:
```suggestion
// WithListener configures the server to use a custom listener.
```

The sentence is no longer valid, as it  raises a panic

server.go
```diff
@@ -352,6 +363,18 @@ func WithoutLogger() func(*Server) {
...
+func WithListener(listener net.Listener) func(*Server) {
+	return func(s *Server) {
+		if s.listener != nil {
+			panic("a listener is already configured; cannot overwrite it")
+		}
+		s.listener = listener
+		s.Server.Addr = listener.Addr().String()
+	}
+}
```
ccoVeille, line 376: Please check, but I think there is a need for something like this


```suggestion
func WithListener(listener net.Listener) func(*Server) {
	return func(s *Server) {
		if s.listener != nil {
			panic("a listener is already configured; cannot overwrite it")
		}
		s.listener = listener
		if _, ok := listener.(tls.Listener) {
			t.isTLS = true
		}
		s.Server.Addr = listener.Addr().String()
	}
}
```

This is pseudo code made on my phone, you may need to adapt/fix it

server.go
```diff
@@ -138,8 +140,12 @@ func NewServer(options ...func(*Server)) *Server {
 		option(s)
 	}
 
+	if s.Server.Addr == "" {
+		WithAddr("localhost:9999")(s)
+	}
+
 	s.OpenApiSpec.Servers = append(s.OpenApiSpec.Servers, &openapi3.Server{
-		URL:         "http://" + s.Addr,
+		URL:         "http://" + s.Server.Addr,
```
ccoVeille, line 148: Shouldn't it be https:// when isTLS is true?

Ignaciojeria: right, i get protocol from proto() method

server.go
Ignaciojeria: I used reflection to check if the instance is of type tls.listener `// WithListener configures the server to use a custom listener.
func WithListener(listener net.Listener) func(*Server) {
	return func(s *Server) {
		if s.listener != nil {
			panic("a listener is already configured; cannot overwrite it")
		}
		s.isTLS = isTLSListener(listener)
		WithAddr(listener.Addr().String())(s)
		s.listener = listener
	}
}

func isTLSListener(listener net.Listener) bool {
	listenerType := reflect.TypeOf(listener)
	if listenerType != nil && listenerType.String() == "*tls.listener" {
		return true
	}
	return false
}
`

server.go
Ignaciojeria: sentence removed

serve.go
Ignaciojeria: Now I maintained localhost as the default to preserve the original implementation

server.go
```diff
@@ -352,6 +364,26 @@ func WithoutLogger() func(*Server) {
...
+	}
+}
+
+func isTLSListener(listener net.Listener) bool {
+	listenerType := reflect.TypeOf(listener)
+	if listenerType != nil && listenerType.String() == "*tls.listener" {
+		return true
+	}
+	return false
+}
```
ccoVeille, line 385: Reflexion is slow, can't you use a type assertion?

server.go
ccoVeille: The code is way better, but using assertion seems overkill, also it's very slow.

Ignaciojeria:
To handle the overwriting of the listener or address, I have three options in mind:

2. If the listener or address is overwritten, print a warning.
3. Continue throwing a panic.
4. Create an array of errors as a private attribute in the server instance and handle them during setup.

What do you think is the best option?

dylanhitt:
I also don't understand the checking around Addr. The default listener should should be created with whatever value `Server.Server.Addr` is.

server.go
dylanhitt: I'm a bit confused on why this is a problem?

serve.go
```diff
@@ -24,19 +27,46 @@ func (s *Server) Run() error {
...
+
+// setupDefaultListener creates a default (non-TLS) listener if none is already configured.
+// If a listener is already set, this method does nothing.
+// Returns an error if the listener cannot be created (e.g., address binding issues).
+func (s *Server) setupDefaultListener() error {
+	if s.listener != nil {
+		return nil // Listener already exists, no action needed.
+	}
+	addr := s.Server.Addr
+	if addr == "" {
```
dylanhitt, line 58: This check isn't really needed. We always default when it is not provided. Also this would only bind the loopback address which would make it confusing for users on why they can't reach the service. The .Listen below should should just take whatever `s.Server.Addr` is.

server.go
```diff
@@ -138,8 +141,12 @@ func NewServer(options ...func(*Server)) *Server {
 		option(s)
 	}
 
+	if s.Server.Addr == "" {
```
dylanhitt, line 144: I think my other comment leads to this just being set where it originally was.

server.go
```diff
@@ -138,8 +141,12 @@ func NewServer(options ...func(*Server)) *Server {
...
+		WithAddr("localhost:9999")(s)
+	}
+
 	s.OpenApiSpec.Servers = append(s.OpenApiSpec.Servers, &openapi3.Server{
-		URL:         "http://" + s.Addr,
+		URL:         fmt.Sprintf("%s://%s", s.proto(), s.Server.Addr),
```
dylanhitt, line 149: nice

server.go
```diff
@@ -296,7 +303,12 @@ func WithPort(port int) func(*Server) {
 // WithAddr optionally specifies the TCP address for the server to listen on, in the form "host:port".
 // If not specified addr ':9999' will be used.
 func WithAddr(addr string) func(*Server) {
-	return func(c *Server) { c.Server.Addr = addr }
+	return func(c *Server) {
+		if c.listener != nil {
+			panic("cannot set addr when a listener is already configured")
```
dylanhitt, line 308: I think we're better off documenting the behavior rather than handling it. Most users will never need to reach the level of passing their own listener. When you reach that point you tend to really know what you're getting yourself into. Would add a note stating that if `WithListener` is used that `WithAddr` is not respected.

serve.go
```diff
@@ -24,19 +27,46 @@ func (s *Server) Run() error {
...
+		return err
+	}
+	return s.Server.ServeTLS(s.listener, certFile, keyFile)
 }
 
-func (s *Server) setup() {
+func (s *Server) setup() error {
+	if err := s.setupDefaultListener(); err != nil {
+		return err
+	}
```
dylanhitt, line 39: I think you should set the `Addr` here.

Like where you do this in the `WithListener` function

```golang
WithAddr(listener.Addr().String())(s)
```

That way we ensure the port being printed in the startup message is for sure correct. At this point there is no way the user can potentially overwrite it. At least our final feedback would be correct.

server.go
```diff
@@ -352,6 +364,26 @@ func WithoutLogger() func(*Server) {
 	}
 }
 
+// WithListener configures the server to use a custom listener.
```
dylanhitt, line 367: Same maybe a mention of `WithAddr` is not respected when a listener is passed.

dylanhitt (changes requested):

server.go
```diff
@@ -352,6 +364,26 @@ func WithoutLogger() func(*Server) {
...
+func WithListener(listener net.Listener) func(*Server) {
+	return func(s *Server) {
+		if s.listener != nil {
+			panic("a listener is already configured; cannot overwrite it")
+		}
+		s.isTLS = isTLSListener(listener)
```
dylanhitt, line 373: It’s not really apparent to me why we care necessarily if the listener is TLS or not?

server.go
dylanhitt: I believe this would always be just http at this point except in the case where a Listener is passed. The current behavior is that we don’t know until the user calls RunTLS which is fine. 

I think we’d need even an even larger refactor to address this well. This falls out of scope of this PR

dylanhitt: Hmmm. Actually probably the proper places is at the end of setup(). 

I could be wrong it’s late 😅

ccoVeille: I agree, I thought it could be a problem too.

So does it have to be merged like that? With somehow a bug?

Or the refactoring should gf done before? Either here or in another PR?

dylanhitt: Doing it in setup now is fine.

server.go
```diff
@@ -352,6 +364,26 @@ func WithoutLogger() func(*Server) {
...
+		WithAddr(listener.Addr().String())(s)
```
dylanhitt, line 374: The one check that may be worth adding. Is the check if the listener `Addr` and the `Server.Server.Addr` match. If they don't `slog.Warn("WithAddr and provided listener addr do match. Default to listen address")` or something along those lines.

Ignaciojeria:
I removed the method that checks if the listener is TLS. Sorry if my commits are chaotic; I’m still learning. What do you think about canceling this pull request? I could create a new fork and consolidate all the changes into a single commit.

ccoVeille:
> I removed the method that checks if the listener is TLS. Sorry if my commits are chaotic; I’m still learning.

It's OK, really.

> What do you think about canceling this pull request? I could create a new fork and consolidate all the changes into a single commit.

I would say keep the PR, it can be squashed in a single commit once merged.

If you still want to do it by your own, you can use `git rebase -i` and merge your commit into one.

But, from my perspective. It's better to have the history of changes in a PR than having it left behind, then another "clean but useless" one

dylanhitt (approved):
Yes, `git rebase -i HEAD~<number>` is a great tool. I use it pretty much everyday. 

Anywho, great work. It's awesome that you're learning by contributing to open source. I found that helped me more than anything. 

Cheers!

serve.go
```diff
@@ -24,26 +29,52 @@ func (s *Server) Run() error {
...
 	go s.OutputOpenAPISpec()
 	s.printStartupMessage()
 
 	s.Server.Handler = s.Mux
 	if s.corsMiddleware != nil {
 		s.Server.Handler = s.corsMiddleware(s.Server.Handler)
 	}
+
+	s.OpenApiSpec.Servers = append(s.OpenApiSpec.Servers, &openapi3.Server{
```
dylanhitt, line 50: Ah changing this causes some problems.

The issue is that we allow users to output the spec by itself. Would you mind putting it back to where it was before the PR. I'll open an issue to refactor it. 

My only thought is that how to do we know if the user is intending to be tls when just outputting the spec 🤷 we don't know that til start time. It probably doesn't matter honestly.

dylanhitt (approved):

Ignaciojeria:
I still have pending unit tests as adding a couple of lines caused the coverage to drop. I'll review it more thoroughly this week after work.

serve.go
```diff
@@ -24,26 +27,45 @@ func (s *Server) Run() error {
...
+func (s *Server) setupDefaultListener() error {
+	if s.listener != nil {
+		WithAddr(s.listener.Addr().String())(s)
+		return nil
+	}
+	listener, err := net.Listen("tcp", s.Addr)
+	if err != nil {
+		return err
+	}
+	WithListener(listener)(s)
```
EwenQuim, line 60: Why?
WithListener sets s.listener, but setting a s.listener will also trigger WithListener that will re-set s.listener with the same value? The flow seems a bit weird to me but I might miss something, please tell me :)

EwenQuim (changes requested):

serve.go
dylanhitt: This is for the case where a user has provided a listener. It would probably be more clear if it was

s.listener = listener than using the With function.

server.go
```diff
@@ -138,8 +141,12 @@ func NewServer(options ...func(*Server)) *Server {
 		option(s)
 	}
 
+	addr := s.Addr
+	if s.listener != nil {
+		addr = s.listener.Addr().String()
+	}
```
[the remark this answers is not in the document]
dylanhitt: It is possible for the user to do something poor here say the user adds their own listener and then their own withaddr. The value could be off since listener trumps everything. I plan on getting rid of this possibility with #248. Basically pushing the openapi server setup as close to startup as we can.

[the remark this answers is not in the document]
EwenQuim: Good idea. Good thing mux and server are decorrelated in the std lib

dylanhitt:
Hey @Ignaciojeria,

I actually have a use case for this myself. If you're low on time I'd be happy to submit commit on to your branch wrapping this up!

Ignaciojeria:
> Hey @Ignaciojeria,
> 
> I actually have a use case for this myself. If you're low on time I'd be happy to submit commit on to your branch wrapping this up!

Of course, that would be great, thanks! I've been a bit busy with work and tried to take a break from coding. If you send a commit, I'll merge it. :D

dylanhitt:
Hey @Ignaciojeria, 

I added the tests and did come cleanup. I am gonna need you to rebase or you main branch though. (The rebase isn't the most fun either I just went through 😅). Here is the PR: https://github.com/Ignaciojeria/fuego/pull/2

the two commits i added are: 
- https://github.com/Ignaciojeria/fuego/pull/2/commits/625a60bc250461d4ccb2f65e5d1548846b86ac78
- https://github.com/Ignaciojeria/fuego/pull/2/commits/401b5e4779cccbd03aa6a1cf6519578b1c5e4260

I also could just open the PR myself. Idk whatever you want to do.

dylanhitt:
hmm. It seems all commits are duplicated. Did you rebase locally?

Ignaciojeria:
> hmm. It seems all commits are duplicated. Did you rebase locally?
Let me clean this up and push the updated branch

dylanhitt:
Hey any update on this? 😄

examples/with-listener/go.mod
```diff
@@ -0,0 +1,28 @@
+module github.com/go-fuego/fuego/examples/with-listener
+
+go 1.22.2
```
dylanhitt, line 3: Been a bit on this. Can we bump to 1.23 and go work sync?

dylanhitt:
@EwenQuim when we go on this flurry of merges I think we should pull this over the line first.

dylanhitt:
@EwenQuim bump on this. I’ve added commits so I don’t like to approve my own work.

EwenQuim:
I'm reading this first thing tonight. Sorry to have taken so long.

EwenQuim (approved):

PR #250, opened 2024-12-05
Adds Path parameter declaration with start-time safety checks

EwenQuim:


dylanhitt (changes requested):
Probably want to use this in `petstore` so we can get the golden to update.

Other than lint and that looks good

openapi_operations.go
```diff
@@ -9,6 +9,7 @@ import (
 type ParamType string // Query, Header, Cookie
 
 const (
+	PathParamType   ParamType = "path"
```
dylanhitt, line 12: Oh we need to update validation as well. `https://github.com/go-fuego/fuego/blob/main/validate_params.go#L5` 

Maybe we should leave a comment here to remember 😅

EwenQuim: Mmh I don't think so, Go router does not route to the right controller if left blank.

Example:
- I declare A: `/user/` and B: `/user/{userid}`
- I request for `/user/`, Go router will just route to A. If `/user/` does not exist, it will simply send a 404

dylanhitt: Ah you're right

dylanhitt (approved):

Issue #247, opened 2024-12-02
WithGlobalResponseTypes should also return XML


DimShadoWWW:
I would like the response type to be configurable.

schema:
```
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/Example"
                }
              },
              "application/xml": {
                "schema": {
                  "$ref": "#/components/schemas/Example"
                }
              }
            },
            "description": "OK"
          },
          "400": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/HTTPError"
                }
              }
            },
            "description": "Bad Request _(validation or deserialization error)_"
          },
          "500": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/HTTPError"
                }
              }
            },
            "description": "Internal Server Error _(panics)_"
          },
```

It inserts "application/json" and it is before "application/xml", this caused it to be selected as default Content-type on documentation.

Also, all errors use "application/json" because there is no way to specify other content-type for them.
(`fuego.WithGlobalResponseTypes` do not change it)

dylanhitt:
`option.RequestContentType("application/xml; charset=UTF-8")` isn't for Bodies it's' for Requests. The code above is only showing requests. 

The response types are hardcoded currently

```
responseSchema := SchemaTagFromType(s, *new(T))
	content := openapi3.NewContentWithSchemaRef(&responseSchema.SchemaRef, []string{"application/json", "application/xml"})
	response := openapi3.NewResponse().WithDescription("OK").WithContent(content)
	route.Operation.AddResponse(200, response)
```

This issue wouldn't be a bug. More a feature request I believe.

dylanhitt:
Ah here it is #159

DimShadoWWW:
I am unable to change the tag to "Feature-request."

Currently, I am working on updating the OpenAPI structure, but I hope that "fuego" can eventually provide a way to specify it directly.

Thank you!

dylanhitt:
Would you mind editing your issue to better reflecting the request. 

Basically stating you want response type to be configurable. At the very least we should update errors to also have an xml response. 

I assume the API you're working on here doesn't support json at all?

DimShadoWWW:
Sure, I updated the the first comment.

I am working on an API that defaults to responding with XML. However, it will optionally support responding in JSON if explicitly requested via the `Accept` header.

[...]

Issue #222, opened 2024-11-10
ability to overwrite how a type is going to be generated in the spec


phenpessoa:
**Is your feature request related to a problem? Please describe.**
Yes, currently there is no way to change how a type is going to be described in the OpenAPI spec, even though we might be using custom Marshalers/Unmarshalers.

**Describe the solution you'd like**
I'd like to have an option to manually overwrite how a type will be described in the OpenAPI spec. I was thinking that maybe it could be an interface that the type implements and returns some sort of data of how it should be described. But open to suggestions.

**Describe alternatives you've considered**
Currently the only alternative is editing the spec file manually.

**Additional context**
To give a concrete example, I have this type:
```go
type Foo struct {
    Bar Option[string] `json:"bar"`
}

type Option[T any] struct {
    Valid bool
    Val   T
}
```
And in the spec it will show up as
```yml
Foo:
  description: Foo schema
  properties:
    bar:
      properties:
        Val:
          type: string
        Valid:
          type: boolean
      type: object
  type: object
```
But, the option field there basically is just a way to make the field nullable, so I'd like it to be encoded as:
```yml
Foo:
  description: Foo schema
  properties:
    bar:
      type: [string, 'null']
  type: object
```

dylanhitt:
This is a great idea! 

We also appreciate your thoughtfulness when describing the feature. Im fairly certain you can just overwrite a route completely if needed, but   I’m away from a computer at least til end of the week.

That being said, providing an idiomatic way to just overwrite an entires route spec is a great idea.

dylanhitt:
I guess the question is do we want the input for this override to be Go type or do we want to allow the user to construct the type themselves? Or maybe both?

phenpessoa:
> To walk through your idea some so I understand. You're describing that in the case above `Foo` would implement say a function called `FuegoAPIOverride()` and the fuego route registration would call that somewhere?

Yup! That was my original idea. And that method would return an object (maybe an object and an error?) which is an OpenAPI Schema/Object/Model. (Not sure if I'm using the right terminology here, sorry). 

So it would be something like this:
```go
// This type would be specified by Fuego, not by the user.
type OpenAPISchema struct {
	Type       []string            `yaml:"type"       json:"type"`
	Required   []string            `yaml:"required"   json:"required"`
	Properties map[string]Property `yaml:"properties" json:"properties"`
	// ... any other fields
}

type Property struct {
	Type    []string `yaml:"type"    json:"type"`
[...]

PR #251, opened 2024-12-05
Declare custom 200 response

EwenQuim:


dylanhitt:
Is the intention for this to cover #222 ?

EwenQuim:
It's not the initial intention but yeah it adds more flexibility, useful for use cases where the user want to declare its own content-types, types, etc...

Adding the ability to override the framework is a must-have: the framework **must be an enabler, not a blocker**!

dylanhitt (approved):
Ah so it relates to #222 and #247. I am gonna take a crack at updating `WithGlobalResponseTypes` for the error part of this.

Cheers

PR #254, opened 2024-12-06
Remove deprecated params

EwenQuim:
Removes deprecated params.

Also fixes the behaviour of `OptionAddError` that was replaced by the global error. We didn't see it when switching to variadic options because the test wasn't made.

### DIFF
--- a/server.go
+++ b/server.go
@@ -124,17 +124,23 @@ func NewServer(options ...func(*Server)) *Server {
 		Security: NewSecurity(),
 	}
 
+	// Default options that can be overridden
 	defaultOptions := [...]func(*Server){
 		WithAddr("localhost:9999"),
 		WithDisallowUnknownFields(true),
 		WithSerializer(Send),
 		WithErrorSerializer(SendError),
 		WithErrorHandler(ErrorHandler),
+	}
+	options = append(defaultOptions[:], options...)
+
+	// Options set if not provided
+	options = append(options,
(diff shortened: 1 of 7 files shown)


server.go
```diff
@@ -124,17 +124,23 @@ func NewServer(options ...func(*Server)) *Server {
...
+	options = append(options,
 		WithGlobalResponseTypes(http.StatusBadRequest, "Bad Request _(validation or deserialization error)_", HTTPError{}),
 		WithGlobalResponseTypes(http.StatusInternalServerError, "Internal Server Error _(panics)_", HTTPError{}),
-	}
+	)
 
-	for _, option := range append(defaultOptions[:], options...) {
+	for _, option := range options {
 		option(s)
 	}
```
EwenQuim, line 145: Yeah this is kind of ugly, but:

Global Response options are set in RegisterOperation AFTER the route Response option, so global response option need to be "soft" (ignore if already exists).
The route options need to be hard (override if already exists).

This is because we're creating a field `(Server).globalOpenAPIResponses`. 

Ideally, as this is a route option, we need a way to register all route options in a queue and use this queue instead of a Server field. For a single route, the options would look like this:

1. Server Global Option A
2. Server Global Option B
3. Group Option A
4. Group Option B
5. SubGroup Option A
6. Subgroup Option B
7. Route option A
8. Route option B

dylanhitt: > A fix could be to set Global Response just before (meaning, before but in the same loop) than route options

Yes

dylanhitt (approved):

PR #256, opened 2024-12-07
Cli/add service generation

rizerkrof:
I added on the `fuego controller` cli command a flag (`--with-service`) to generate a simple in memory map based service that satisfies the controller service interface.

The generated service can be pass into the controller resources to have a fully functional CRUD.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ coverage.out
 recipe.db
 bin
 go.work.sum
-fuego
+/cmd/fuego/fuego
--- a/cmd/fuego/commands/controller.go
+++ b/cmd/fuego/commands/controller.go
@@ -25,7 +25,12 @@ func Controller() *cli.Command {
 				fmt.Println("Note: You can add a controller name as an argument. Example: `fuego controller books`")
 			}
 
-			_, err := createController(entityName)
+			_, err := createControllerFile(entityName, "entity.go", entityName+".go")
+			if err != nil {
+				return err
+			}
+
+			_, err = createControllerFile(entityName, "controller.go", entityName+"Controller.go")
 			if err != nil {
 				return err
 			}
@@ -37,7 +42,7 @@ func Controller() *cli.Command {
 }
 
 // createController creates a new controller file
-func createController(entityName string) (string, error) {
+func createControllerFile(entityName, controllerTemplateFileName, outputFileName string) (string, error) {
 	controllerDir := "./controller/"
 	if _, err := os.Stat(controllerDir); os.IsNotExist(err) {
 		err = os.Mkdir(controllerDir, 0o755)
@@ -46,7 +51,7 @@ func createController(entityName string) (string, error) {
 		}
 	}
 
-	templateContent, err := templates.FS.ReadFile("controller/controller.go")
+	templateContent, err := templates.FS.ReadFile("controller/" + controllerTemplateFileName)
 	if err != nil {
 		return "", err
 	}
@@ -57,7 +62,7 @@ func createController(entityName string) (string, error) {
 	newContent := strings.ReplaceAll(string(templateContent), "newEntity", entityName)
 	newContent = strings.ReplaceAll(newContent, "NewEntity", titler.String(entityName))
 
-	controllerPath := fmt.Sprintf("%s%s.go", controllerDir, entityName)
+	controllerPath := fmt.Sprintf("%s%s", controllerDir, outputFileName)
 
 	err = os.WriteFile(controllerPath, []byte(newContent), 0o644)
 	if err != nil {
--- a/cmd/fuego/commands/controller.go
+++ b/cmd/fuego/commands/controller.go
@@ -38,6 +38,20 @@ func Controller() *cli.Command {
 			fmt.Printf("🔥 Controller %s created successfully\n", entityName)
 			return nil
 		},
+		Flags: []cli.Flag{
+			&cli.BoolFlag{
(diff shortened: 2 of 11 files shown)


cmd/fuego/templates/controller/service.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/go-fuego/fuego"
+)
+
+type NewControllerServiceImpl struct {
```
EwenQuim, line 11: Why `NewController` ? Kind of confusing for future contributors. You can name it `NewEntity` maybe?

cmd/fuego/commands/controller.go
```diff
@@ -25,19 +25,40 @@ func Controller() *cli.Command {
...
-			_, err := createController(controllerName)
+			_, err := createControllerFile(controllerName, "controller.go", controllerName+".go")
 			if err != nil {
 				return err
 			}
 
 			fmt.Printf("🔥 Controller %s created successfully\n", controllerName)
 			return nil
 		},
+		Flags: []cli.Flag{
```
EwenQuim, line 36: I prefer if you make a function `createService` that is either called by `fuego service` or additionally by `fuego controller --with-service`.

You can access the flag within cli context by calling `cCtx.String("with-service")` in the Action.

.gitignore
```diff
@@ -6,5 +6,4 @@ coverage.out
 .env.local
 recipe.db
 bin
-go.work.sum
-fuego
\ No newline at end of file
+go.work.sum
```
EwenQuim, line 9: I'm afraid changing this might be problematic if you try to build fuego cli, I think it's ok to have this ignore. Unless it prevents something else to be tracked?

documentation/docs/tutorials/02-crud.md
```diff
@@ -14,7 +14,13 @@ fuego controller books
...
 ```
 
-This generates a controller and a service for the `books` resource.
+This generates a controller for the `books` resource.
+
+:::tip
+
+Use `fuego controller --with-service books` to generate a simple map based in memory service so you can pass it to the controller resources to quickly interact with your new book entity!
+
+:::
```
EwenQuim, line 24: Thanks a lot for thinking about updating the docs 🙏🏼

cmd/fuego/templates/controller/service.go
```diff
@@ -0,0 +1,90 @@
...
+	newControllerRepository map[string]NewController
+	mu                      sync.RWMutex
```
EwenQuim, line 13: Nice, you could have used [`sync.Map`](https://pkg.go.dev/sync#Map) to not be bothered with Mutexes, but it works!

EwenQuim (changes requested):
Thank you very much for this PR, it'll greatly help new Fuego users try the framework !

Next step, a proposal for using SQL 😉  I'm joking, this can be a real challenge and there are so many different ways of implementing SQL...

A few minor changes need to be addressed in order to ensure the maintainability of this feature, even if there is no doubt that you made it work nicely.

.gitignore
rizerkrof: it prevents any files in cmd/fuego to be tracked

cmd/fuego/templates/controller/service.go
rizerkrof: `sync.Map` does not work with generics as far as I know. I didn't want to add noisy code to wrap `sync.Map` around a generic compliant interface.

EwenQuim:
Btw fixes #256

cmd/fuego/commands/controller.go
```diff
@@ -25,19 +25,40 @@ func Controller() *cli.Command {
...
+					}
+
+					controllerName := cCtx.Args().First()
+
+					_, err := createControllerFile(controllerName, "service.go", controllerName+"Service.go")
+					if err != nil {
+						return err
+					}
+
+					return nil
```
dylanhitt, line 53: nitpick

```suggestion
					return err
```

dylanhitt (approved):

cmd/fuego/templates/newEntity/controller.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (rs NewEntityResources) putNewEntity(c *fuego.ContextWithBody[NewEntityUpdate]) (NewEntity, error) {
+	id := c.PathParam("id")
+
+	body, err := c.Body()
+	if err != nil {
+		return NewEntity{}, err
+	}
+
+	new, err := rs.NewEntityService.UpdateNewEntity(id, body)
```
dylanhitt, line 55: is this generated?

rizerkrof: What do you mean?

EwenQuim (approved):
Nice!

cmd/fuego/templates/newEntity/controller.go
dylanhitt: I'm asking if it was generated from like codegen or something 

For instance this can be `return rs.NewEntityService.UpdateNewEntity(id, body)`

EwenQuim: This will be replaced by `return rs.BookService.UpdateBook(id, body)`

EwenQuim: This file is just a template.

EwenQuim (approved):

PR #258, opened 2024-12-09
Sets response headers in the OpenAPI spec

EwenQuim:
As per the title.

A few questions:

- How do we handle different response types? Currently, it only sets the response headers for 200 response, which is nice but do not cover all existing cases. 
- Should we have some kind of validation? I'm currently trying to raise a warning if the header is marked as required but do not exist in the HTTP response. Any other ideas?


ccoVeille:
I'm unsure about the need here, also I'm not a guru in fuego, just a random Gopher reviewing some PRs.

I played with OpenAPI enough to know it can be fun.

OpenAPI allows you to define multiple response for the same http code. Think about different validation error messages.
These are often displayed in documentation with something called as "examples".

Also, the format could be totally different between a 403, 404, 500, 503, 429, 200, 201, 204

So is your request about dealing with the non-200 replies?

Does it takes into account the fact the OpenAPI reply can be totally different?

If not, it means it has to include an allOf or anyOf in the definition (I'm unsure about the exact term sorry).

Because for me, if you plan to support different code, the logic of handling different replies should be also considered

ccoVeille:
> I'm not sure that's true, that's what OpenAPI 4.0 "moonwalk" is mostly about.

That's why I talked about the examples sections.

The examples will be simple, but the spec would have to use anyOf and define each type.

option.go
```diff
@@ -149,8 +149,49 @@ func panicsIfNotCorrectType(openapiParam *openapi3.Parameter, exampleValue any)
...
+	openapiParam.In = ""
+
+	if len(apiParam.StatusCodes) == 0 {
+		apiParam.StatusCodes = []int{200}
+	}
+
+	return func(r *BaseRoute) {
+		for _, code := range apiParam.StatusCodes {
+			codeString := strconv.Itoa(code)
+			response200 := r.Operation.Responses.Value(codeString)
```
dylanhitt, line 173: The variable name is confusing for me here. IS the only available for 20x error codes?

If so maybe name this `response2XX`. Also maybe we should validate that the user is only providing 200s?

EwenQuim: The naming will be changed after merging the other PR

EwenQuim: Ready for review again :)

dylanhitt (approved):
One thought about this MR though. Whenever i get around to working on #247.

In the struct 

```golang
type Response struct{
  ContentTypes []string
  Type any
}
```
Do we want to add Headers to that input struct?

option.go
```diff
@@ -149,8 +149,49 @@ func panicsIfNotCorrectType(openapiParam *openapi3.Parameter, exampleValue any)
 	return exampleValue
 }
 
-// Registers a parameter for the route. Prefer using the [Query], [QueryInt], [Header], [Cookie] shortcuts.
-func OptionParam(name string, options ...func(*OpenAPIParam)) func(*BaseRoute) {
+// Declare a response header for the route.
+// This will be added to the OpenAPI spec, under the 200 response.
```
dylanhitt, line 153: This a bit confused here. Is this only for 200 responses. Is this sentence only suppose to be on context of the example below?

EwenQuim: Yes I'll update it

EwenQuim:
> Do we want to add Headers to that input struct?

Yeah probably, depends on how you'll be structuring that

option.go
EwenQuim: Updated

dylanhitt:
Yeah, I'm not sure either. I'm sure I'll find reason to do it or not once i'm in the code.

dylanhitt (approved):

EwenQuim:
Thanks again for proofreading this @dylanhitt !

We passed the 500 tests milestone 🥳 

image

PR #259, opened 2024-12-09
Default status code

EwenQuim:
Adds a default status code in the server code and in the OpenAPI Description

### DIFF
--- a/examples/petstore/controllers/pets.go
+++ b/examples/petstore/controllers/pets.go
@@ -46,6 +46,7 @@ func (rs PetsResources) Routes(s *fuego.Server) {
 
 	fuego.Get(petsGroup, "/by-age", rs.getAllPetsByAge, option.Description("Returns an array of pets grouped by age"))
 	fuego.Post(petsGroup, "/", rs.postPets,
+		option.DefaultStatusCode(201),
 		option.AddError(409, "Conflict: Pet with the same name already exists", PetsError{}),
 	)
 
--- a/openapi.go
+++ b/openapi.go
@@ -188,19 +188,23 @@ func RegisterOpenAPIOperation[T, B any](s *Server, route Route[T, B]) (*openapi3
 		addResponseIfNotSet(s, route.Operation, openAPIGlobalResponse.Code, openAPIGlobalResponse.Description, openAPIGlobalResponse.ErrorType)
 	}
 
-	// Automatically add non-declared 200 Response
-	response200 := route.Operation.Responses.Value("200")
-	if response200 == nil {
-		response := openapi3.NewResponse().WithDescription("OK")
-		route.Operation.AddResponse(200, response)
-		response200 = route.Operation.Responses.Value("200")
+	// Automatically add non-declared 200 (or other) Response
+	if route.DefaultStatusCode == 0 {
+		route.DefaultStatusCode = 200
+	}
+	defaultStatusCode := strconv.Itoa(route.DefaultStatusCode)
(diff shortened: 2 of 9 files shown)


openapi.go
```diff
@@ -188,19 +188,23 @@ func RegisterOpenAPIOperation[T, B any](s *Server, route Route[T, B]) (*openapi3
 		addResponseIfNotSet(s, route.Operation, openAPIGlobalResponse.Code, openAPIGlobalResponse.Description, openAPIGlobalResponse.ErrorType)
 	}
 
-	// Automatically add non-declared 200 Response
-	response200 := route.Operation.Responses.Value("200")
-	if response200 == nil {
+	// Automatically add non-declared 200 (or other) Response
+	if route.DefaultStatusCode == 0 {
+		route.DefaultStatusCode = 200
```
ccoVeille, line 193:
```suggestion
		route.DefaultStatusCode = http.StatusOK
```

examples/petstore/controllers/pets.go
```diff
@@ -46,6 +46,7 @@ func (rs PetsResources) Routes(s *fuego.Server) {
 
 	fuego.Get(petsGroup, "/by-age", rs.getAllPetsByAge, option.Description("Returns an array of pets grouped by age"))
 	fuego.Post(petsGroup, "/", rs.postPets,
+		option.DefaultStatusCode(201),
```
ccoVeille, line 49:
```suggestion
		option.DefaultStatusCode(http.StatusCreated),
```

openapi.go
```diff
@@ -188,19 +188,23 @@ func RegisterOpenAPIOperation[T, B any](s *Server, route Route[T, B]) (*openapi3
...
+	}
+	defaultStatusCode := strconv.Itoa(route.DefaultStatusCode)
+	responseDefault := route.Operation.Responses.Value(defaultStatusCode)
+	if responseDefault == nil {
 		response := openapi3.NewResponse().WithDescription("OK")
```
ccoVeille, line 198: Shouldn't it be ?


```suggestion
		response := openapi3.NewResponse().WithDescription(http.StatusText(route.DefaultStatusCode))
```

option.go
```diff
@@ -287,6 +287,13 @@ func OptionHide() func(*BaseRoute) {
 	}
 }
 
+// Hide hides the route from the OpenAPI spec.
+func OptionDefaultStatusCode(defaultStatusCode int) func(*BaseRoute) {
+	return func(r *BaseRoute) {
+		r.DefaultStatusCode = defaultStatusCode
+	}
+}
```
ccoVeille, line 295: I don't get why you keep adding new things at the root, people should use the ones in options.

Is it a dependency cycle issue?

openapi.go
EwenQuim: Good one!

option.go
EwenQuim: Yes it is

ccoVeille: I think you could sort this by adding a new package like `route` that will contain only BaseRoute. You may have issues with OpenAPIParam at some point...

Anyway dependency cycle is never fun, so I can get you don't want to invest on fixing them.

Anyway, I think you should add the `// Deprecated: use option.DefaultStatusCode` here and for all existing code.

It should be OK, I don't think IDE will report the fact option.DefaultStatusCode is deprecated because it uses OptionDefaultStatusCode

EwenQuim: Yes why not

ccoVeille (approved):

dylanhitt (approved):

Issue #241, opened 2024-11-29
Gin compatibility


EwenQuim:
**Is your feature request related to a problem? Please describe.**

Gin is very popular and **a lot of projects and companies are tied with Gin**. Supporting Fuego controllers in Gin or Gin controllers in Fuego might be a good eventual transition for companies wanting to migrate to the new powerful Go1.22 router + Fuego OpenAPI capabilities 

**Describe the solution you'd like**
I'm still thinking about what would it look like. Is there any Gin <-> net/http adaptor to begin with?

dylanhitt:
@EwenQuim I believe we can close this as well.

EwenQuim:
We still have to support Gin middlewares. I don't know how we'll do that

EwenQuim:
I mean, we are supporting gin middlewares through `gin.Engine.Use`, not the route handlers chain

EwenQuim:
we can write an other issue, this is symbolic to close this one ;) 🎉

PR #262, opened 2024-12-11
Decorrelates openapi registration from server

EwenQuim:
The goal here is to make `RegisterOpenAPIOperation` independent from `fuego.Server`. Doing so, we might reuse this function for other routers, like Gin #241 or Echo.

`RegisterOpenAPIOperation` only needed OpenAPI details, not the whole router anyway! What we had before was a demonstration of implementation leak. With a `Server` struct that have many fields, we are locking ourselves where it wasn't necessary.

This PR is backward-compatible! No test have change except for semantics. OMG I love Go for making this refactoring so easy...


mux.go
```diff
@@ -112,7 +112,7 @@ func Register[T, B any](s *Server, route Route[T, B], controller http.Handler, o
 	route.Path = s.basePath + route.Path
 
 	var err error
-	route.Operation, err = RegisterOpenAPIOperation(s, route)
+	route.Operation, err = RegisterOpenAPIOperation(s.OpenAPI, route)
 	if err != nil {
```
EwenQuim, line 116: That's what this PR is mostly about

dylanhitt: yeah this is a good idea.

openapi.go
```diff
@@ -164,7 +193,7 @@ func validateSwaggerUrl(swaggerUrl string) bool {
 }
 
 // RegisterOpenAPIOperation registers an OpenAPI operation.
-func RegisterOpenAPIOperation[T, B any](s *Server, route Route[T, B]) (*openapi3.Operation, error) {
+func RegisterOpenAPIOperation[T, B any](s *OpenAPI, route Route[T, B]) (*openapi3.Operation, error) {
 	if route.Operation == nil {
```
EwenQuim, line 197: Enables the possibility to do `RegisterOpenAPIOperation[T, B](fuegoGinCompat.OpenAPI, route)`

mux.go
```diff
@@ -112,7 +112,7 @@ func Register[T, B any](s *Server, route Route[T, B], controller http.Handler, o
 	route.Path = s.basePath + route.Path
 
 	var err error
-	route.Operation, err = RegisterOpenAPIOperation(s, route)
+	route.Operation, err = RegisterOpenAPIOperation(s.OpenAPI, route)
```
ccoVeille, line 115: If you are passing route, to set a field of route

Why is it the signature like this?


```suggestion
	err = route.RegisterOpenAPIOperation(s.OpenAPI)
```

I know I'm suggesting a code change, but it can be retro compatible

ccoVeille: Ob maybe because you are relying g on the magic of generics

dylanhitt (approved):

PR #269, opened 2024-12-13
Added Security Policy

EwenQuim:
Do you all agree?


Makefile
```diff
@@ -39,6 +39,9 @@ lint:
 lint-markdown:
 	markdownlint --ignore documentation/node_modules --dot .
 
+lint-markdown-fix:
```
dylanhitt, line 42: I've always like used the pattern above this with lint. In this case `lint-markdown` is what the user uses and the CI becomes

```
golangci-lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - uses: actions/setup-go@v5
        with:
          go-version: ${{ env.GO_VERSION }}

      - run: make lint FIX=""
```

something like the above

SECURITY.md
```diff
@@ -0,0 +1,25 @@
+# Security Policy
+
+## Supported Versions
+
+The following table outlines which versions of Fuego are actively supported with security updates. Please ensure that you are using a supported version to benefit from the latest patches and improvements.
+
+| Version | Supported            |
+| ------- | -------------------- |
+| 0.x.y (x being the latest version released)     | :white_check_mark: |
+| 0.x.y (x being NOT the latest version released)     | :x: |
```
dylanhitt, line 10: Eh, I don't really like making the promise of maintaining older versions until at least `v1`, but considering the we are changing a lot all the time. I could be convinced.

dylanhitt (changes requested):

SECURITY.md
EwenQuim: The `:x:` is a red cross, meaning we won't support them ;)

Makefile
EwenQuim: mmmh I see, I'm removing it then

SECURITY.md
dylanhitt: AH. Reading raw markup is hard

Makefile
```diff
@@ -37,7 +37,7 @@ lint:
 	golangci-lint run ${FIX} ./...
 
 lint-markdown:
```
dylanhitt, line 39: Oh. I'm sorry i just noticed we just use the action for our markdown lint ci job. Yeah if you will just remove the `${FiX}` now. My apologies.

SECURITY.md
EwenQuim: Yes, and it is also important to understand with raw text. Fixed by also adding raw text.

PR #272, opened 2024-12-13
Refactor/set openapi server url

### LINKED ISSUE #248: refactor setting of - Server.OpenApiSpec.Servers
Right now we do it in the constructor of [Server function](https://github.com/go-fuego/fuego/blob/main/server.go#L14). We should push this closer to startup of the server as we can interpret if the user is running an http or https server as well as get the proper address since #245 we do not know the final address we use until `s.setup()` is called nor if the server is https or http. 

In regards to just printing. We can just assume `http` for now. If we find we have a reason to allow the user to ensure the server setup is https we can update the `s.OutputOpenAPISpec()` to be `s.OutputOpenAPISpec(isTLS bool)`. We really don't have another way of doing this since we truly don't know if the server is an http or https webserver until startup.

dylanhitt:
closes #248 

Will conflict with #245 but it does solve the issue of creating the server too early so that code could be simplified I believe. This also adds native option to configure the openapi.Description() server.

Thanks


openapi_test.go
```diff
@@ -328,6 +330,29 @@ func TestServer_OutputOpenApiSpec(t *testing.T) {
...
+					Server: &openapi3.Server{
+						URL:         "foo",
+						Description: "bar",
+					},
+				},
+			),
+		)
+		Get(s, "/", func(*ContextNoBody) (MyStruct, error) {
+			return MyStruct{}, nil
+		})
```
EwenQuim, line 348: Nitpick: You can use `dummyController` for lisibility

server.go
```diff
@@ -17,14 +17,24 @@ import (
...
-	DisableLocalSave bool                              // If true, the server will not save the OpenAPI JSON spec locally
-	SwaggerUrl       string                            // URL to serve the swagger UI
-	UIHandler        func(specURL string) http.Handler // Handler to serve the OpenAPI UI from spec URL
-	JsonUrl          string                            // URL to serve the OpenAPI JSON spec
-	JsonFilePath     string                            // Local path to save the OpenAPI JSON spec
-	PrettyFormatJson bool                              // Pretty prints the OpenAPI spec with proper JSON indentation
+	// Overrides the server description generated by fuego
+	Server *openapi3.Server
+	// If true, the server will not serve the Swagger UI nor the OpenAPI JSON spec
+	DisableSwagger bool
```
EwenQuim, line 23: Not sure that a chunk (`*openapi3.Server`) of the OpenAPI Description should be in the `OpenAPIConfig`.

`OpenAPIConfig` is more about "How will Fuego handle the output of the OpenAPI Description". But as there's no comment, you wouldn't have guessed it, sorry about that!

server.go
```diff
@@ -135,11 +145,6 @@ func NewServer(options ...func(*Server)) *Server {
 		option(s)
 	}
 
-	s.OpenAPI.Description().Servers = append(s.OpenAPI.Description().Servers, &openapi3.Server{
-		URL:         "http://" + s.Addr,
-		Description: "local server",
-	})
-
 	s.startTime = time.Now()
```
EwenQuim, line 149: Good thing to get rid of this here, thanks!

server_test.go
```diff
@@ -83,6 +83,10 @@ func TestWithOpenAPIConfig(t *testing.T) {
 	t.Run("with custom values", func(t *testing.T) {
 		s := NewServer(
 			WithOpenAPIConfig(OpenAPIConfig{
+				Server: &openapi3.Server{
+					URL:         "foo",
+					Description: "bar",
+				},
 				SwaggerUrl:       "/api",
 				JsonUrl:          "/api/openapi.json",
```
EwenQuim, line 91: Mmh I prefer our users to code themselves:

```go
s.OpenAPI.Description().Servers = append(...)
```

Rather than giving them a shortcut that does exactly the same. I mean, we have no added value of doing this

openapi.go
```diff
@@ -95,6 +95,14 @@ func declareAllTagsFromOperations(s *Server) {
 // Also serves a Swagger UI.
 // To modify its behavior, use the [WithOpenAPIConfig] option.
 func (s *Server) OutputOpenAPISpec() openapi3.T {
+	if s.OpenAPIConfig.Server == nil {
+		s.OpenAPIConfig.Server = &openapi3.Server{
+			URL:         s.proto() + "://" + s.Addr,
```
EwenQuim, line 100: So to sum up, the strong point of this PR is mostly this s.proto am I right?

Nice!

EwenQuim (changes requested):
Thank you very much! I'm suggesting a much more simple PR by not taking care of `Server` in the config but keeping the late initialisation with `s.proto()`, but don't hesitate to challenge!

server.go
dylanhitt: Yeah I figured as much. I personally think this does have to do with the OpenApi description. That's all this change is really focusing on.

openapi.go
dylanhitt: No, it's to provide a way to configure a value that we've been setting for our users. This leads into #248. 

If you don't think the configurability is worth it. Please let me just revert the single commit that adds the feature. Thanks

dylanhitt:
Closing per suggestions and #273

PR #273, opened 2024-12-13
refactor: set openapi server url as close to startup as possible

### LINKED ISSUE #248: refactor setting of - Server.OpenApiSpec.Servers

dylanhitt:
Closes #248 

Will conflict with https://github.com/go-fuego/fuego/pull/245 but it does solve the issue of creating the server too early so that code could be simplified now I believe. 

Thanks

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -96,7 +96,7 @@ func declareAllTagsFromOperations(s *Server) {
 // To modify its behavior, use the [WithOpenAPIConfig] option.
 func (s *Server) OutputOpenAPISpec() openapi3.T {
 	s.OpenAPI.Description().Servers = append(s.OpenAPI.Description().Servers, &openapi3.Server{
-		URL:         s.proto() + "://" + s.Addr,
+		URL:         s.url(),
 		Description: "local server",
 	})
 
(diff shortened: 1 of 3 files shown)


openapi.go
```diff
@@ -95,6 +95,11 @@ func declareAllTagsFromOperations(s *Server) {
 // Also serves a Swagger UI.
 // To modify its behavior, use the [WithOpenAPIConfig] option.
 func (s *Server) OutputOpenAPISpec() openapi3.T {
+	s.OpenAPI.Description().Servers = append(s.OpenAPI.Description().Servers, &openapi3.Server{
+		URL:         s.proto() + "://" + s.Addr,
```
ccoVeille, line 99: there should be a method on `Server` for computing `s.proto() + "://" + s.Addr`

ccoVeille (approved):

EwenQuim (approved):
Nice, thanks!

openapi.go
ccoVeille: It's also used in

https://github.com/go-fuego/fuego/blob/main/serve.go

dylanhitt: OIC in the startup messages. Will update in a second

dylanhitt:
Updated with `s.url()` method.

openapi.go
ccoVeille: Adding this would have been appreciated
```
    Co-authored-by: ccoVeille <3875889+ccoVeille@users.noreply.github.com>
```

dylanhitt: I don't have push rights to main. If @EwenQuim wouldn't mind doing it that would be great.

PR #275, opened 2024-12-13
Description option : override + default description contains the list of middlewares

EwenQuim:


openapi.go
```diff
@@ -258,7 +270,7 @@ func RegisterOpenAPIOperation[T, B any](openapi *OpenAPI, route Route[T, B]) (*o
 	for _, params := range route.Operation.Parameters {
 		if params.Value.In == "path" {
 			if !strings.Contains(route.Path, "{"+params.Value.Name) {
-				return nil, fmt.Errorf("path parameter '%s' is not declared in the path", params.Value.Name)
+				panic(fmt.Errorf("path parameter '%s' is not declared in the path", params.Value.Name))
 			}
```
EwenQuim, line 274: The tests were previously passing because of a panic later in the code but it wasn't intentional, here it is, as it should.

examples/petstore/lib/testdata/doc/openapi.golden.json
```diff
@@ -748,8 +748,8 @@
 				]
 			},
 			"get": {
-				"description": "controller: `github.com/go-fuego/fuego/examples/petstore/controllers.PetsResources.getPets`\n\n---\n\n",
-				"operationId": "GET_/pets/:id",
+				"description": "Replace description with this sentence.",
```
dylanhitt, line 751: What's the reasoning behind `AddDescription` over `Description`. This looks more like it is setting it.

EwenQuim:
The end goal is to have this default description : 

image

A user might want the ability to disable or override it.  

Perhaps providing a route-level option to disable it would be a better solution?

examples/petstore/lib/testdata/doc/openapi.golden.json
EwenQuim: No, `AddDescription` really is a string concatenation, whereas `Description` is simply setting it.

dylanhitt:
I was more asking about the naming of the function. Why change to Add vs just `Description`. You seem to be overriding.

option/option.go
```diff
@@ -105,8 +105,15 @@ var Description = fuego.OptionDescription
 // AddDescription adds a description to the route.
 // By default, the description is set by Fuego with some info,
 // like the controller function name and the package name.
+//
+// Deprecated: Use [Description] instead.
```
dylanhitt, line 109: OHHH. Yeah I missed this call out

mux.go
```diff
@@ -58,10 +58,18 @@ type BaseRoute struct {
 	Middlewares          []func(http.Handler) http.Handler
 	AcceptedContentTypes []string // Content types accepted for the request body. If nil, all content types (*/*) are accepted.
 	Hidden               bool     // If true, the route will not be documented in the OpenAPI spec
+	OverrideDescription  bool     // Override the default description
```
dylanhitt, line 61: Only thing I can say is. Does this really need to be exported?

option.go
```diff
@@ -273,7 +273,17 @@ func OptionDescription(description string) func(*BaseRoute) {
 // like the controller function name and the package name.
```
dylanhitt, line 273: Mind calling out deprecation here as well?

dylanhitt (approved):

openapi.go
```diff
@@ -209,6 +214,18 @@ func RegisterOpenAPIOperation[T, B any](openapi *OpenAPI, route Route[T, B]) (*o
...
+	if route.FullName == "" {
+		route.FullName = route.Path
+	}
+
+	if route.Operation.Summary == "" {
+		route.Operation.Summary = route.NameFromNamespace(camelToHuman)
+	}
+
+	if route.Operation.OperationID == "" {
+		route.Operation.OperationID = route.Method + "_" + strings.ReplaceAll(strings.ReplaceAll(route.Path, "{", ":"), "}", "")
```
ccoVeille, line 226: This should be moved to a method, here is looks like an hack to clean things, while it's very important

openapi.go
```diff
@@ -263,7 +280,7 @@ func RegisterOpenAPIOperation[T, B any](openapi *OpenAPI, route Route[T, B]) (*o
 	for _, params := range route.Operation.Parameters {
 		if params.Value.In == "path" {
 			if !strings.Contains(route.Path, "{"+params.Value.Name) {
-				return nil, fmt.Errorf("path parameter '%s' is not declared in the path", params.Value.Name)
+				panic(fmt.Errorf("path parameter '%s' is not declared in the path", params.Value.Name))
```
ccoVeille, line 283: Why a panic?

option.go
```diff
@@ -273,7 +273,17 @@ func OptionDescription(description string) func(*BaseRoute) {
...
 	return func(r *BaseRoute) {
-		r.Operation.Description += "\n\n" + description
+		r.Operation.Description += description
+	}
+}
+
+// Description appends a description to the route.
+// By default, the description is set by Fuego with some info,
+// like the controller function name and the package name.
+func OptionOverrideDescription(description string) func(*BaseRoute) {
```
ccoVeille, line 283:
```suggestion
// OptionOverrideDescription overrides the default description set by Fuego. 
// By default, the description is set by Fuego with some info, 
// like the controller function name and the package name.
func OptionOverrideDescription(description string) func(*BaseRoute) {
```

mux.go
ccoVeille: I agree

option.go
ccoVeille: If you don't want users  to use it from the root, you can then deprecate it. Here and suggest people to use the one in option package

openapi.go
EwenQuim: We don't want to declare non-existing path params. Do you think a warning/an error is enough?

openapi.go
EwenQuim: Done. Thanks!

EwenQuim: Done, thanks!

openapi.go
ccoVeille: I would say if the app crash on boot, I'm OK with a panic, otherwise it could be a real problem if it occurs only on some condition like having a call that will cause a panic.

I think it's a first case, so why not

dylanhitt: Yeah I think a general we should is anything before start is fine too panic where it's genuinely pretty difficult to get the error to the user. Another case where I like panic is when the users configuration just should be possible. I.E providing nil as a Type etc

PR #274, opened 2024-12-13
feat: AddResponse 

dylanhitt:
to allow for setting or overriding of any routes status code

Could really decide whether I wanted to provide a constructor for `fuego.Response` pretty open on that. I did elect to not allow for variadic default of type for 400+ error codes as we discussed in some previous issues/MRs that we shouldn't assuming that all errors are `fuego.HTTPError`'s especially with the recent updates with Default Error Handler.

Also deprecation of AddError in favor of AddResponse.

PR #276, opened 2024-12-14
BREAKING: update WithGlobalResponseType to take

### LINKED ISSUE #247: WithGlobalResponseTypes should also return XML

dylanhitt:
Inspired by #274. 

- Changes `WithGlobalResponseType` to use `fuego.Response` as an input
- Defaults `WithGlobalResponseType` inputs to contain `application/json` and `application/xml` closes #247 
- Refactors `WithGlobalResponseType` to use the same helper as `OptionAddResponse` so we can stay in lock step with `fuego.Response` struct

Some notes: 

- I'm not the biggest fan of `openAPIResponse` for a struct name but it's just to hold the global response types until we get to generating the spec. Maybe `GlobalReponse` open to thoughts.
- thoughts on making `addResponseIfNotSet` a method

This was a fun PR. Looking forward to the feedback.

Cheers

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -24,7 +24,7 @@ func NewOpenAPI() *OpenAPI {
 	return &OpenAPI{
 		description:            &desc,
 		generator:              openapi3gen.NewGenerator(),
-		globalOpenAPIResponses: []openAPIError{},
+		globalOpenAPIResponses: []openAPIResponse{},
 	}
 }
 
@@ -32,7 +32,7 @@ func NewOpenAPI() *OpenAPI {
 type OpenAPI struct {
 	description            *openapi3.T
 	generator              *openapi3gen.Generator
-	globalOpenAPIResponses []openAPIError
+	globalOpenAPIResponses []openAPIResponse
 }
 
 func (d *OpenAPI) Description() *openapi3.T {
@@ -244,7 +244,13 @@ func RegisterOpenAPIOperation[T, B any](openapi *OpenAPI, route Route[T, B]) (*o
 
 	// Response - globals
 	for _, openAPIGlobalResponse := range openapi.globalOpenAPIResponses {
-		addResponseIfNotSet(openapi, route.Operation, openAPIGlobalResponse.Code, openAPIGlobalResponse.Description, openAPIGlobalResponse.ErrorType)
+		addResponseIfNotSet(
+			openapi,
+			route.Operation,
+			openAPIGlobalResponse.Code,
+			openAPIGlobalResponse.Description,
+			openAPIGlobalResponse.Response,
+		)
 	}
 
 	// Automatically add non-declared 200 (or other) Response
--- a/openapi_operations_test.go
+++ b/openapi_operations_test.go
@@ -98,7 +98,7 @@ func TestWithGlobalResponseType(t *testing.T) {
 		routeGlobal := Get(s, "/test-global", testController)
 		routeCustom := Get(s, "/test-custom", testController,
 			OptionAddResponse(http.StatusBadRequest, "My Local Error", Response{Type: MyLocalResponse{}}),
-			OptionAddError(http.StatusTeapot, "My Local Teapot", Response{Type: HTTPError{}}),
+			OptionAddResponse(http.StatusTeapot, "My Local Teapot", Response{Type: HTTPError{}}),
 		)
 
 		require.Equal(t, "My Global Error", *routeGlobal.Operation.Responses.Value("400").Value.Description, "Overrides Fuego's default 400 error")
--- a/openapi_operations_test.go
+++ b/openapi_operations_test.go
@@ -1,6 +1,7 @@
 package fuego
 
 import (
+	"net/http"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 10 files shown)


openapi_operations_test.go
```diff
@@ -61,32 +62,110 @@ func TestCustomError(t *testing.T) {
...
+		s := NewServer(
+			WithGlobalResponseTypes(http.StatusBadRequest, "My Global Error", Response{Type: MyGlobalResponse{}}),
+			WithGlobalResponseTypes(http.StatusNotImplemented, "Another Global Error", Response{Type: MyGlobalResponse{}}),
+		)
+
+		routeGlobal := Get(s, "/test-global", testController)
+		routeCustom := Get(s, "/test-custom", testController,
+			OptionAddResponse(http.StatusBadRequest, "My Local Error", Response{Type: MyLocalResponse{}}),
+			OptionAddError(http.StatusTeapot, "My Local Teapot", Response{Type: HTTPError{}}),
+		)
```
EwenQuim, line 102: OptionAddResponse missing here!

openapi.go
```diff
@@ -24,15 +24,15 @@ func NewOpenAPI() *OpenAPI {
...
 	}
 }
 
 // Holds the OpenAPI OpenAPIDescription (OAD) and OpenAPI capabilities.
 type OpenAPI struct {
 	description            *openapi3.T
 	generator              *openapi3gen.Generator
-	globalOpenAPIResponses []openAPIError
+	globalOpenAPIResponses []openAPIResponse
 }
```
EwenQuim, line 37: No need for it now. We can send it as a global `routeOption` that will be set at the Route creation time!

But maybe for another PR.

server.go
```diff
@@ -195,18 +195,20 @@ func WithCorsMiddleware(corsMiddleware func(http.Handler) http.Handler) func(*Se
...
-	errorType = append(errorType, HTTPError{})
+func WithGlobalResponseTypes(code int, description string, response Response) func(*Server) {
 	return func(c *Server) {
-		c.OpenAPI.globalOpenAPIResponses = append(c.OpenAPI.globalOpenAPIResponses, openAPIError{code, description, errorType[0]})
+		c.OpenAPI.globalOpenAPIResponses = append(
+			c.OpenAPI.globalOpenAPIResponses,
+			openAPIResponse{Code: code, Description: description, Response: response},
+		)
 	}
 }
```
EwenQuim, line 214: Now it is the same as this :

```go
func WithGlobalResponseTypes(code int, description string, response Response) func(*Server) {
	return func(c *Server) {
		WithRouteOptions(
			OptionAddResponse(code, description, response),
		)(c)
	}
}
```


Or, can we remove it completely? I think it's better if we remove redundant configuration items. I think that everything that will be applied to a route must be applied as a route option.

EwenQuim (changes requested):
This was a huge PR. I think it's really good! 

I think we can do further refacto but maybe you can do it in another PR.

Sorry for the 3 days delay!

I mark as "request changes" for the missing `OptionAddError` used in tests.

server.go
dylanhitt: Hmm let met test this real quick. 

I personally really like being allowed to set global response types for servers/groups.

dylanhitt: Yeah these two things aren't the same.

```
--- FAIL: TestWithGlobalResponseType (0.00s)
    --- FAIL: TestWithGlobalResponseType/errors_with_route_overrides (0.00s)
        openapi_operations_test.go:104: 
                Error Trace:    /Users/dhitt/github/go-fuego/fuego/openapi_operations_test.go:104
                Error:          Not equal: 
                                expected: "My Global Error"
                                actual  : "Bad Request _(validation or deserialization error)_"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -My Global Error
                                +Bad Request _(validation or deserialization error)_
                Test:           TestWithGlobalResponseType/errors_with_route_overrides
                Messages:       Overrides Fuego's default 400 error
```

dylanhitt: I guess we can just use WithRouteOptions here as well. I forget that exists always. Let me figure out the reason for the error above and I'll submit another PR

dylanhitt: At least I tested this well enough 😅

dylanhitt: Okay, nice catch. Just needed to change up this: https://github.com/go-fuego/fuego/pull/276/commits/81a7afcda321dc40268a9f1b0d08a7d7d7dc2feb

So couple things. Yes we should deprecate `WithGlobalResponseType`. I think we should get this change merge with a deprecation notice and best way to migrate to `WithRouteOptions`. Emphasis on maintaining the two default api responses declare as well. I'll probably move those over to use `WithRouteOptions` in this MR as well so we don't forget. 

How does that sound?

openapi.go
dylanhitt: Assuming you agree with https://github.com/go-fuego/fuego/pull/276#discussion_r1887167481

This will just be removed after deprecation after the deprecation notice of this thing on the next release.

server.go
EwenQuim: I think it looks good!

And yes moving back the default server options is really nice as well, that's something we wanted to do anyway.

EwenQuim:
I'll wait for your PR to be finished before merging mine, because there will be conflicts I think ;) 

Only the deprecation notice for `WithGlobalResponseType` is missing. And I think you can use WithRouteOptions in  `NewServer` in `server.go` file.

server.go
dylanhitt: Okay cool, I’ll get this don’t either tonight or tomorrow night then.

dylanhitt:
Updated

EwenQuim (approved):

ccoVeille (commented):
LGTM :+1: 

But I won't approve as the changes are out of my knowlege in term of regression.

PR #277, opened 2024-12-14
chore: cleanup some status codes and godoc

dylanhitt:
Some cleanup suggest by @ccoVeille in #274

### DIFF
--- a/serve_test.go
+++ b/serve_test.go
@@ -514,7 +514,7 @@ func TestServer_RunTLS(t *testing.T) {
 			}()
 
 			// wait for the server to start
-			conn, err := net.DialTimeout("tcp", s.Server.Addr, 2*time.Second)
+			conn, err := net.DialTimeout("tcp", s.Server.Addr, 5*time.Second)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


serve_test.go
```diff
@@ -514,7 +514,7 @@ func TestServer_RunTLS(t *testing.T) {
 			}()
 
 			// wait for the server to start
-			conn, err := net.DialTimeout("tcp", s.Server.Addr, 2*time.Second)
+			conn, err := net.DialTimeout("tcp", s.Server.Addr, 5*time.Second)
 			if err != nil {
```
EwenQuim, line 518: Yeah why not

EwenQuim (approved):

serve_test.go
dylanhitt: I admittedly did not look through the entirety of of `Dial` to see if refused meant timeout. But 🤷. Feels like the server isn't starting up quick enough to me personally.

ccoVeille:
Thanks for the co-authored-by

PR #279, opened 2024-12-16
Refactor, mutualize and document Route object

EwenQuim:
Allow to centralize the route object creation : less code duplication, usable from the Gin compat package, and also documented the Route object.

Not the most important PR to be honest, but it cleans the code and make it cleaner.

### DIFF
--- a/route.go
+++ b/route.go
@@ -1,0 +1,85 @@
+package fuego
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+func NewRoute[T, B any](method, path string, handler any, openapi *OpenAPI, options ...func(*BaseRoute)) Route[T, B] {
+	return Route[T, B]{
+		BaseRoute: NewBaseRoute(method, path, handler, openapi, options...),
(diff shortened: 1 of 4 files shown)


route.go
```diff
@@ -0,0 +1,67 @@
...
+		o(&baseRoute)
+	}
+
+	return baseRoute
+}
+
+// BaseRoute is the base struct for all routes in Fuego.
+// It contains the OpenAPI operation and other metadata.
+type BaseRoute struct {
+	Operation            *openapi3.Operation // GENERATED OpenAPI operation, do not set manually in Register function. You can change it after the route is registered.
```
dylanhitt, line 43: Nitpick: I've always been a fan of the comment above the field

dylanhitt (approved):

route.go
EwenQuim: Yeah it allows easier multiline read. I've just copy-pasted this one but I'll try to migrate all comments above the field in the next PRs

EwenQuim: Fixed. Finally I'm merging, as your PR does not collide with this one :)

PR #282, opened 2024-12-18
Set status code before returning when no data

### LINKED ISSUE #281: Could not return NoContent status code with empty body response
**To Reproduce**

1. set up route with option.DefaultStatusCode(204)
2. define route as returning (any, error) and return nil, nil
3. use this route

**Expected behavior**
This route should return code 204

**Actual behavior**
200 is returned

**Screenshots**
image

0.16.2

1.23

**Additional context**
Seems like this code (serve.go:138) is executed in the wrong order:
```
if reflect.TypeOf(ans) == nil {
    return
}

if route.DefaultStatusCode != 0 {
    w.WriteHeader(route.DefaultStatusCode)
}
```

EwenQuim:
Fixes #281

### DIFF
--- a/option_test.go
+++ b/option_test.go
@@ -1,6 +1,7 @@
 package fuego_test
 
 import (
+	"errors"
 	"log/slog"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


EwenQuim:
Trivial & simple bugfix, auto-merging! Don't hesitate to leave post-merge comments

PR #287, opened 2024-12-18
[Go 1.24] [Fuego 1.0] Params in context

EwenQuim:
Not at all ready.

#286 is nice so we can have easily extendable interfaces.

This PR is backward compatible! We just need to wait for Go1.24 to be out to enjoy it.


EwenQuim:
Replaced by #406

PR #288, opened 2024-12-19
Mutualise a part of Context between net/http and gin

EwenQuim:
Tried to mutualize the Context so we will have less work to do for other integrations.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -12,6 +12,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/go-fuego/fuego/internal"
 )
 
 const (
@@ -77,6 +79,12 @@ type ContextWithBody[B any] interface {
 	Header(key string) string                 // Get request header
 	SetHeader(key, value string)              // Sets response header
 
+	// Returns the underlying net/http, gin or echo context.
+	//
+	// Usage:
+	//  ctx := c.Context() // net/http: the [context.Context] of the *http.Request
+	//  ctx := c.Context().(*gin.Context) // gin: Safe because the underlying context is always a [gin.Context]
+	//  ctx := c.Context().(echo.Context) // echo: Safe because the underlying context is always a [echo.Context]
 	Context() context.Context
 
 	Request() *http.Request        // Request returns the underlying HTTP request.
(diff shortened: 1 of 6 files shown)


ctx.go
```diff
@@ -77,6 +77,12 @@ type ContextWithBody[B any] interface {
 	Header(key string) string                 // Get request header
 	SetHeader(key, value string)              // Sets response header
 
+	// Returns the underlying net/http, gin or echo context.
+	//
+	// Usage:
+	//  ctx := c.Context() // net/http: the [context.Context] of the *http.Request
+	//  ctx := c.Context().(*gin.Context) // gin: Safe because the underlying context is always a [gin.Context]
+	//  ctx := c.Context().(echo.Context) // echo: Safe because the underlying context is always a [echo.Context]
```
dylanhitt, line 85:
```suggestion
	//  ctx := c.Context().(*echo.Context) // echo: Safe because the underlying context is always a [echo.Context]
```

Maybe? Can't tell honestly.

internal/common_context.go
```diff
@@ -0,0 +1,216 @@
...
+// If the query parameter does not exist or is not a bool, it returns the default value declared in the OpenAPI spec.
+// For example, if the query parameter is declared as:
+//
+//	fuego.Get(s, "/test", myController,
+//	  option.QueryBool("is_ok", "Is OK?", param.Default(true))
+//	)
+//
+// and the query parameter does not exist in the HTTP request, it will return true.
+// Accepted values are defined as [strconv.ParseBool]
+func (c CommonContext[B]) QueryParamBoolErr(name string) (bool, error) {
```
dylanhitt, line 176: I don't think these can be `internal` our users use these?

dylanhitt (changes requested):

ctx.go
EwenQuim: No, it's an interface, like Fuego now

internal/common_context.go
EwenQuim: They're embedded in a struct that is hidden behind the Context interface!

dylanhitt: Oh. I see now. Thanks for the clarification. I didn't realize those would be accessible though. TIL

dylanhitt (approved):

PR #290, opened 2024-12-20
Moved all non-net/http related code to the new Engine struct

EwenQuim:
As the title says.

It's not a lot haha, and we could have expected this, because we're really close to `net/http`. The next step will be to fit our Content-Negociation into this Engine. But we'll release the Gin compatibility package before as it's not the end if it only supports `application/json` for a few weeks.

### DIFF
--- a/server.go
+++ b/server.go
@@ -56,8 +56,7 @@ type Server struct {
 	disableAutoGroupTags   bool
 	basePath               string // Base path of the group
 
-	// Points to the server OpenAPI struct.
-	OpenAPI *OpenAPI
+	*Engine
 
 	Security Security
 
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -56,25 +56,22 @@ type Server struct {
...
 	DisallowUnknownFields bool // If true, the server will return an error if the request body contains unknown fields. Useful for quick debugging in development.
 	DisableOpenapi        bool // If true, the routes within the server will not generate an OpenAPI spec.
 	maxBodySize           int64
 
-	Serialize      Sender                // Custom serializer that overrides the default one.
-	SerializeError ErrorSender           // Used to serialize the error response. Defaults to [SendError].
-	ErrorHandler   func(err error) error // Used to transform any error into a unified error type structure with status code. Defaults to [ErrorHandler]
-	startTime      time.Time
+	Serialize      Sender      // Custom serializer that overrides the default one.
+	SerializeError ErrorSender // Used to serialize the error response. Defaults to [SendError].
```
dylanhitt, line 72: I guess while we're here. Would you mind sending all the comments above the fields? If not approved

dylanhitt (approved):

server.go
EwenQuim: Done! I did not do it immediately for readability but I thought about it ;)

PR #291, opened 2024-12-20
Optionshow

### LINKED ISSUE #270: Move `(Server).DisableOpenapi` to a route option
The `(Server).DisableOpenapi` and the two functions below must be removed in favor of a route option named `fuego.OptionShowInOpenAPI(bool)` and its alias `option.ShowInOpenAPI(bool)`. Can be useful in some cases (the user is serving HTML for example...). 

```go
// Hide prevents the routes in this server or group from being included in the OpenAPI spec.
func (s *Server) Hide() *Server {
	s.DisableOpenapi = true
	return s
}

// Show allows displaying the routes. Activated by default so useless in most cases,
// but this can be useful if you deactivated the parent group.
func (s *Server) Show() *Server {
	s.DisableOpenapi = false
	return s
}
```

dylanhitt:
Closes #270

This doesn't follow the theme of what #270 was saying but I believe it covers it. If we were to do `fuego.OptionShowInOpenAPI(bool)`. I say we should deprecate `OptionHide` in favor of use `fuego.OptionShowInOpenAPI(bool)`. 

Either way one hidden feature with this pull request is that it now allows for users to hide at the group level but also show a route in that group

### DIFF
--- a/option/option.go
+++ b/option/option.go
@@ -170,5 +170,8 @@ var RequestContentType = fuego.OptionRequestContentType
 // Hide hides the route from the OpenAPI spec.
 var Hide = fuego.OptionHide
 
+// Hide hides the route from the OpenAPI spec.
+var Show = fuego.OptionShow
+
 // DefaultStatusCode sets the default status code for the route.
 var DefaultStatusCode = fuego.OptionDefaultStatusCode
--- a/option/option.go
+++ b/option/option.go
@@ -170,7 +170,7 @@ var RequestContentType = fuego.OptionRequestContentType
 // Hide hides the route from the OpenAPI spec.
 var Hide = fuego.OptionHide
 
-// Hide hides the route from the OpenAPI spec.
+// Show shows the route from the OpenAPI spec.
 var Show = fuego.OptionShow
 
 // DefaultStatusCode sets the default status code for the route.
--- a/server.go
+++ b/server.go
@@ -64,8 +64,6 @@ type Server struct {
 	fs       fs.FS
 	template *template.Template // TODO: use preparsed templates
 
-	acceptedContentTypes []string
-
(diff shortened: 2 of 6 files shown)


option/option.go
```diff
@@ -170,5 +170,8 @@ var RequestContentType = fuego.OptionRequestContentType
 // Hide hides the route from the OpenAPI spec.
 var Hide = fuego.OptionHide
 
+// Hide hides the route from the OpenAPI spec.
```
EwenQuim, line 173: Beware of baming here, copy=pasted too fast ;)

engine.go
```diff
@@ -12,5 +12,5 @@ type Engine struct {
 	OpenAPI      *OpenAPI
 	ErrorHandler func(error) error
 
-	acceptedContentTypes []string
+	acceptedContentTypes []string //nolint:unused
```
EwenQuim, line 15: If it is not used anymore, you can delete it.

server.go
```diff
@@ -64,8 +64,10 @@ type Server struct {
 	fs       fs.FS
 	template *template.Template // TODO: use preparsed templates
 
-	DisallowUnknownFields bool // If true, the server will return an error if the request body contains unknown fields. Useful for quick debugging in development.
-	DisableOpenapi        bool // If true, the routes within the server will not generate an OpenAPI spec.
+	acceptedContentTypes []string
```
EwenQuim, line 67: Why is it added back on the Server struct?

EwenQuim (changes requested):

server.go
dylanhitt: Ah I had a conflict on a rebase.

EwenQuim (approved):

PR #293, opened 2024-12-20
[skip ci] - WIP req resp logger

dylanhitt:
Looking for some feedback on this idea for the UI of setting/disabling a request response logger. 

`WithRequestResponseLog(nil)` - Would disable middleware logging completely and not register the middleware

`WithRequestResponseLog(NewRequestResponseLog(RequestLog, nil))` - Would set the default log on request time

`WithRequestResponseLog(NewRequestResponseLog(nil, ResponseLog))` - Would set the default log at response. 

I believe this also should be move to `Server.Engine`

### DIFF
--- a/default_middlewares.go
+++ b/default_middlewares.go
@@ -1,0 +1,22 @@
+package fuego
+
+import (
+	"log/slog"
+	"net/http"
+)
+
+func (l *RequestResponseLogger) Log(next http.Handler) http.Handler {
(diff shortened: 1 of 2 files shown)


dylanhitt:
My other thought is breaking it our entirely so something like `WithRequestLog` and `WithResponseLog`. If both are nil don't register the middleware. If either is nil zero it with `func(w http.ResponseWriter, r *http.Request) {}`, and then assume default is wanted unless provided.

PR #294, opened 2024-12-20
Chore/req res logger v2

dylanhitt:


dylanhitt:
Would probably get rid of the `RequestResponseLogger` struct and add the fields directly to server or just add it directly without named field. 🤷

PR #295, opened 2024-12-20
chore: comment Deprecatd -> OptionDeprecated

dylanhitt:
My developer tools complain about this being deprecated when it isn't. Not sure if there is a better way to avoid this

### DIFF
--- a/option.go
+++ b/option.go
@@ -293,7 +293,7 @@ func OptionOperationID(operationID string) func(*BaseRoute) {
 	}
 }
 
-// Deprecated marks the route as deprecated.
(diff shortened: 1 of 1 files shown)


EwenQuim:
Lol, that's unfortunate and a bit ironic 😆


Deprecated must obey some rules in Go standard : beginning after a paragraph and be followed by a colon, so your IDE is wrong.

But if you get this behaviour, it means you're probably not alone, and I don't want you or any other user to be bothered by this, so yeah I'm ok with this change!

EwenQuim (approved):

Issue #285, opened 2024-12-18
Continuous benchmark


EwenQuim:
**Is your feature request related to a problem? Please describe.**

To avoid performance regressions.

**Describe the solution you'd like**

Something like this: https://gofiber.github.io/fiber/benchmarks/

**Describe alternatives you've considered**

**Additional context**

Jeremie-Chauvel:
Hey!
Might be interested in looking into it over holidays. Will keep you posted!

EwenQuim:
Nice! 

I spoke with Adrien Cacciaguerra, he was interested, but Codspeed does not support Go yet...

Jeremie-Chauvel:
It was ofc my idea 😁

ekumamatthew:
Could I be assigned to this?

EwenQuim:
I thought @Jeremie-Chauvel wanted to work on it. Ask him!

But you already have 2 first good issues to work on :)

Jeremie-Chauvel:
Hey @ekumamatthew, you can start, I will try to add a codspeed integration but I need to add go support to codspeed first. Having some benchmarks to upgrade to codspeed ones will help me anyway. Worst case scenario, I don't do the code speed integration and you still have benchmarks for Fuego.

TLDR: go for it

EwenQuim:
Using https://github.com/benchmark-action/github-action-benchmark or native go benchmark with [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) can be a good idea. Way easier than implementing Go support to Codspeed...

EwenQuim:
As @ekumamatthew is already working on other issues, I suggest that you can work on this one if you'd like!

EwenQuim:
Maybe try to update/reuse the Makefile commands.

But here it's a special issue because it's CI, so no guidelines!

aidenwong812:
Understand. Thanks.

adriencaccia:
Support for go on CodSpeed is in the works at https://github.com/CodSpeedHQ/codspeed-go 😉
We would love for you to try it out once it is ready

PR #299, opened 2024-12-21
Optimize memory usage by reordering fields

EwenQuim:
Just a dumb PR to reduce memory usage and improve performance by simply realigning struct fields.  
I can't believe this has to be done manually—obviously, the compiler should handle it.  

This also enables all govet linters in the CI (including the `fieldalignment` check).  

We'll wait for #285 to see the improvement or lack thereof.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,6 +24,12 @@ linters-settings:
     locale: "US"
     error: true
 
+  govet:
+    enable-all: true
+    disable:
+      - shadow
+
+
   gofumpt:
     extra-rules: true
   staticcheck:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,7 +28,7 @@ linters-settings:
     enable-all: true
     disable:
       - shadow
-
+      - fieldalignment
 
   gofumpt:
     extra-rules: true
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ dependencies-analyze:
 	govulncheck ./...
 
 fmt:
+	which betteralign || go install github.com/dkorunic/betteralign/cmd/betteralign@latest
+	betteralign -fix -apply .
 	which gofumpt || go install mvdan.cc/gofumpt@latest
 	gofumpt -l -w -extra .
 
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,6 @@ dependencies-analyze:
 	govulncheck ./...
 
 fmt:
-	which betteralign || go install github.com/dkorunic/betteralign/cmd/betteralign@latest
-	betteralign -fix -apply .
 	which gofumpt || go install mvdan.cc/gofumpt@latest
 	gofumpt -l -w -extra .
 
--- a/ctx.go
+++ b/ctx.go
@@ -124,19 +124,21 @@ func NewNetHTTPContext[B any](route BaseRoute, w http.ResponseWriter, r *http.Re
 // has a Body. The Body type parameter represents the expected data type
 // from http.Request.Body. Please do not use a pointer as a type parameter.
 type netHttpContext[Body any] struct {
-	internal.CommonContext[Body]
+	Res http.ResponseWriter
+
+	fs fs.FS
 
 	body *Body // Cache the body in request context, because it is not possible to read an HTTP request body multiple times.
 
-	Req *http.Request
-	Res http.ResponseWriter
-
-	fs        fs.FS
+	Req       *http.Request
 	templates *template.Template
 
-	readOptions     readOptions
 	serializer      Sender
 	errorSerializer ErrorSender
+
+	internal.CommonContext[Body]
+
+	readOptions readOptions
 }
 
 var (
@@ -153,8 +155,8 @@ func (c netHttpContext[B]) SetStatus(code int) {
 
 // readOptions are options for reading the request body.
 type readOptions struct {
-	DisallowUnknownFields bool
 	MaxBodySize           int64
+	DisallowUnknownFields bool
 	LogBody               bool
 }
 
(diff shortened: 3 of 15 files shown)


ccoVeille (approved):

dylanhitt (approved):
TIL

.golangci.yml
```diff
@@ -24,6 +24,12 @@ linters-settings:
     locale: "US"
     error: true
 
+  govet:
+    enable-all: true
+    disable:
+      - shadow
+
+
```
dylanhitt, line 32: nit pick extra space.

dylanhitt:
> What do you think about the Makefile change @ccoVeille @dylanhitt ? It makes make fmt and then make optimize struct memory usage automatically!

Yes I like it. Will say maybe we shouldn't be installing things on people machines so implicitly. I tend to use like a `make init` or `make deps`. For instance I would `which` like how you're doing and then if fail i would print a statement to the user saying x needs to be installed and `make init` will do it for you.

Makefile
```diff
@@ -28,6 +28,8 @@ dependencies-analyze:
 	govulncheck ./...
 
 fmt:
+	which betteralign || go install github.com/dkorunic/betteralign/cmd/betteralign@latest
```
ccoVeille, line 31: Is there so much differences with govet fieldalignment that you want to use this fork?

ccoVeille (changes requested):
I don't think anything like this is needed.

You are building an API service, with OpenAPI features.

Here it's about optimizing bytes of memory on a framework, that stay still between 2 API calls.

If someone needs memory efficient solution, I'm afraid to say they will never use a framework like fuego, I mean any framework.

I would like to discourage you to go in this direction that is making the struct less readable.

For me it's pure "oh it's good feature we should try"

So unless you have a benchmark that shows a change of at least 20% I'm term of speed or memory consumption, I would say NOGO

dylanhitt:
wait a minute what?

> Here it's about optimizing bytes of memory on a framework, that stay still between 2 API calls.

A fuego context and fuego response is gonna be created on every API call to some using this framwork? If i have an API with 1000 concurrent calls (which I do). You're allocating a context for all 1000 api calls.

dylanhitt:
Like this struct alone could be created `10,000` times: https://github.com/go-fuego/fuego/pull/299/files#diff-c4dd82978a1a49bcd620aff72208bbd8114f09341d701e188c247ce3e3511117R126

ccoVeille:
I'm sorry if my reply 

Is fuego highly optimized for ultra fast performance? I cannot tell.

The recent PR I see are about adding feature, not about looking for insane performance

I'm not saying the changes in this PR have no consequences at all. It's proven, known, documented, there are linters about it.

My point is "are they noticeable?"

But will it has a consequence of server bills? A noticeable consequence on performance?

What I can see is the consequence on readability

You talked about an allocatio on every http calls, but what is it compared the cost in memory of receiving and returning the HTTP messages ?

For me such optimization should be on compiler side. But it isn't unfortunately.

Also, such optimization provides an easy fix for small memory optimization. But this is far from being a way to improve a framework performance.

It's very good optimization when dealing with millions of data to sort in memory, yes.

But here, I'm doubtful. That's why I'm asking for benchmarks.

If you think the readability is still OK after such "optimization", why not.

dylanhitt:
> You talked about an allocatio on every http calls, but what is it compared the cost in memory of receiving and returning the HTTP messages ?

Again I'm not really understanding your point here either. This would be entirely on the user using fuego. We won't know their body or requestbody it's up to them to optimize their own structs. The best we can do is optimize what we can control when someone wants to squeak out some performance where they can.

dylanhitt:
And to answer readability. No I don't really see an impact/maintenance over grouping struct fields. If we really cared so much about this I would much rather focus on better commits/concise PRs than I would the grouping of the struct. Git can at least tell a story.

ccoVeille:
I apparently fail to explain why I think (hypothesis) it will have quite no consequences on the end user.

At the same time, I think that I'm getting you upset or disappointed, because of what I said, I'm sorry about that.

Also, you think it could (here another hypothesis) have noticeable consequences for the performance.

I conceeded there will be improvement, my point is about the fact I don't think the could be noticed.

I can be wrong, and it won't be a problem to me.

I asked for a benchmark, because I would like to get something that can be checked/compared.

Maybe @EwenQuim can tell what he thinks about it. Maybe he made a benchmark

dylanhitt:
I'm not upset or disappointed. I'm just not understanding the point being made. I don't know what

> Here it's about optimizing bytes of memory on a framework, that stay still between 2 API calls.

When I read that it sounds like you're saying is that "the code in fuego really only is responsible for 2 API calls so why optimize"? 

I do agree that the majority of the optimization is really dependent on our users. 

Fuego specifically `fuego.Server` really doesn't do much past creating a `netHttpContext` for every api call in regards to memory. However that is struct is allocated every time a fuego server route is used. So let's take our current benchmark on `main`. Specifically `BenchmarkContext_Body/valid_JSON_body-10` and `BenchmarkRequest/fuego_server_and_fuego_post-10` main gives us

```
BenchmarkContext_Body/valid_JSON_body-10                  389785              3055 ns/op            7258 B/op         32 allocs/op
BenchmarkRequest/fuego_server_and_fuego_post-10           207608              5718 ns/op            8273 B/op         69 allocs/op
BenchmarkRequest/fuego_server_and_std_post-10             307184              3850 ns/op            7489 B/op         46 allocs/op
BenchmarkRequest/std_server_and_std_post-10               391288              3054 ns/op            7137 B/op         29 allocs/op
``` 

vs this branch

```
BenchmarkContext_Body/valid_JSON_body-10                  400556              3012 ns/op            7242 B/op         32 allocs/op
BenchmarkRequest/fuego_server_and_fuego_post-10           207830              5660 ns/op            8257 B/op         69 allocs/op
BenchmarkRequest/fuego_server_and_std_post-10             300900              3814 ns/op            7489 B/op         46 allocs/op
BenchmarkRequest/std_server_and_std_post-10               371756              3028 ns/op            7137 B/op         29 allocs/op
```

So 16 Bytes per request.

ccoVeille:
Thanks for doing the benchmark.

EwenQuim:
Don’t worry @ccoVeille, everything is all right!

I understand this approach might be cumbersome for readability in most cases, so I’ll remove the PR and the linter.

However, it still makes sense to optimize memory usage for certain structs, especially `netHttpContext`, since it’s used on every request and remains hidden from the user’s perspective.

Makefile
EwenQuim: Yes, there was the -fix option that worked!

But I'm removing this from the CI

PR #301, opened 2024-12-22
Updated all dependencies and requires Go 1.23

EwenQuim:
Needed because I want to use the new `maps` and `slices` packages. 

We will move to 1.24 in a few months anyway so this is not horrible. I made a PR so I won't forget it in the Release Notes as a breaking change


EwenQuim:
Merging because I have too much PRs in parallel and I want to continue, but don't hesitate to send post-merging comments

PR #260, opened 2024-12-09
Gin compat

EwenQuim:
Fixes #241 : it works!!! 🎉 

🚧 Ultra-WIP as it is not ready for production.

Enables the way for Echo & other frameworks.

The (challengeable) goal here is not to support both frameworks but to allow smooth transition towards stdlib.

PR #302, opened 2024-12-23
BREAKING: create generic engine.OutputOpenAPISpec for use in all engines

dylanhitt:
This pulls out as much functionality as possible in regards to generating the assets when calling `engine.OutputOpenApiSpec`.

This change is breaking in so far as the `OpenAPIConfig` will be composed of `EngineOpenAPIConfig`. I personally don't like these names but I want to ask about changing it here first. I would prefer that we change `OpenAPIConfig` to `ServerOpenAPIConfig` and `EngineOpenAPIConfig` becomes the `OpenAPIConfig`. The current `OpenAPIConfig` does have some uniqueness in regards to the `fuego.Server`. We probably could consolidate nearly all of it though if in #260 we created an interface that dealt with route/url registration, but I'm not sure how far we want to take this. Or if someone can think of a better name for `EngineOpenAPIConfig`.

For instance every framework we bring in would need to be able to handle something like this to make it work

```golang
// OutputOpenAPISpec takes the OpenAPI spec and outputs it to a JSON file and/or serves it on a URL.
// Also serves a Swagger UI.
	s.OpenAPI.Description().Servers = append(s.OpenAPI.Description().Servers, &openapi3.Server{
		URL:         s.url(),
		Description: "local server",
	})

	if !s.OpenAPIConfig.DisableSwagger {
		s.registerOpenAPIRoutes(s.Engine.OutputOpenApiSpec())
	}

	return *s.OpenAPI.Description()
}

// Registers the routes to serve the OpenAPI spec and Swagger UI.
func (s *Server) registerOpenAPIRoutes(jsonSpec []byte) {
	GetStd(s, s.OpenAPIConfig.JsonUrl, func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "application/json")
		_, _ = w.Write(jsonSpec)
	})
	s.printOpenAPIMessage(fmt.Sprintf("JSON spec: %s%s", s.url(), s.OpenAPIConfig.JsonUrl))

	if !s.OpenAPIConfig.DisableSwaggerUI {
		Register(s, Route[any, any]{
			BaseRoute: BaseRoute{
				Method: http.MethodGet,
				Path:   s.OpenAPIConfig.SwaggerUrl + "/",
			},
		}, s.OpenAPIConfig.UIHandler(s.OpenAPIConfig.JsonUrl))
		s.printOpenAPIMessage(fmt.Sprintf("OpenAPI UI: %s%s/index.html", s.url(), s.OpenAPIConfig.SwaggerUrl))
	}
}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -82,6 +82,8 @@ type Server struct {
 }
 
 // NewServer creates a new server with the given options.
+// Fuego's [Server] is built on top of the standard library's [http.Server].
+// The OpenAPI and data flow is handled by the [Engine], a lightweight abstraction available for all kind of routers (net/http, Gin, Echo).
 // For example:
 //
 //	app := fuego.NewServer(
@@ -90,6 +92,7 @@ type Server struct {
 //	)
 //
 // Options all begin with `With`.
+// Some options are at engine level, and can be set with [WithEngineOptions].
 // Some default options are set in the function body.
 func NewServer(options ...func(*Server)) *Server {
 	s := &Server{
@@ -405,7 +408,7 @@ func WithOpenAPIServerConfig(config OpenAPIServerConfig) func(*Server) {
 //		),
 //	)
 //
-// Options all begin with `With`.
+// Engine Options all begin with `With`.
 func WithEngineOptions(options ...func(*Engine)) func(*Server) {
 	return func(s *Server) {
 		for _, option := range options {


engine.go
```diff
@@ -1,16 +1,139 @@
...
+	// Pretty prints the OpenAPI spec with proper JSON indentation
+	PrettyFormatJson bool
+}
+
+// OutputOpenAPISpec takes the OpenAPI spec and outputs it to a JSON file
+func (e *Engine) OutputOpenApiSpec() []byte {
+	e.OpenAPI.computeTags()
+
+	// Validate
+	err := e.OpenAPI.Description().Validate(context.Background())
```
ccoVeille, line 52: Either the context is useless, either the function should take it as argument

engine.go
```diff
@@ -1,16 +1,139 @@
...
+		return errors.New("error writing file ")
+	}
+
+	e.printOpenAPIMessage("JSON file: " + jsonSpecLocalPath)
+	return nil
+}
+
+func (s *Engine) marshalSpec() ([]byte, error) {
+	if s.OpenAPIConfig.PrettyFormatJson {
+		return json.MarshalIndent(s.OpenAPI.Description(), "", "	")
```
ccoVeille, line 97: using strings.Repeat with a number would be less awkward

server.go
```diff
@@ -15,21 +15,26 @@ import (
...
+	// If true, the server will not serve the Swagger UI nor the OpenAPI JSON spec
+	DisableSwagger bool
+	// If true, the server will not serve the Swagger UI
+	DisableSwaggerUI bool
+	// URL to serve the swagger UI
+	SwaggerUrl string
+	// Handler to serve the OpenAPI UI from spec URL
+	UIHandler func(specURL string) http.Handler
+	// URL to serve the OpenAPI JSON spec
+	JsonUrl string
```
ccoVeille, line 28:
```suggestion
	SwaggerURL string
	// Handler to serve the OpenAPI UI from spec URL
	UIHandler func(specURL string) http.Handler
	// URL to serve the OpenAPI JSON spec
	openAPISpecURL string
```

dylanhitt:
Assuming this is acceptable. I believe really the decisions regarding how far we take this will be based on how far we want to take #260.

engine.go
```diff
@@ -1,16 +1,139 @@
...
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"log/slog"
+	"os"
+	"path/filepath"
+)
+
+func NewEngine(config ...OpenAPIConfig) *Engine {
```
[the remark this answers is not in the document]
dylanhitt: Same.

engine.go
dylanhitt: So you're suggesting what exactly? I'm not really a fan of `OutputOpenApiSpec(context.Context)`. We could create ones ourselves in line I guess.

engine.go
dylanhitt: Yeah we could. Would maybe do that in a separate PR as it's gonna regen the entire golden file which would be some noise. (that large space is a tab on a space).

engine.go
```diff
@@ -1,16 +1,139 @@
...
+	// If true, the engine will not save the OpenAPI JSON spec locally
+	DisableLocalSave bool
+	// Local path to save the OpenAPI JSON spec
+	JsonFilePath string
+	// Pretty prints the OpenAPI spec with proper JSON indentation
+	PrettyFormatJson bool
+}
+
+// OutputOpenAPISpec takes the OpenAPI spec and outputs it to a JSON file
+func (e *Engine) OutputOpenApiSpec() []byte {
```
ccoVeille, line 48:
```suggestion
func (e *Engine) OutputOpenAPISpec() []byte {
```

engine.go
ccoVeille: Yes,i was suggesting that.

My problem here is that the context created here means the validate code cannot use a real context.

So my point was to either use a context in the openapispec signature, or consider the context in validate is useless

engine.go
```diff
@@ -1,16 +1,139 @@
...
+	f, err := os.Create(jsonSpecLocalPath) // #nosec G304 (file path provided by developer, not by user)
+	if err != nil {
+		return errors.New("error creating file")
+	}
+	defer f.Close()
+
+	_, err = f.Write(jsonSpec)
+	if err != nil {
+		return errors.New("error writing file ")
+	}
```
ccoVeille, line 89: Two issues here

- a strange trailing space
- you create an error, whole you have already one: wrap it


```suggestion
	_, err = f.Write(jsonSpec)
	if err != nil {
		return fmt.Errorf("error writing file: %w", err)
	}
```

engine.go
```diff
@@ -1,16 +1,139 @@
...
+	return jsonSpec
+}
+
+func (e *Engine) saveOpenAPIToFile(jsonSpecLocalPath string, jsonSpec []byte) error {
+	jsonFolder := filepath.Dir(jsonSpecLocalPath)
+
+	err := os.MkdirAll(jsonFolder, 0o750)
+	if err != nil {
+		return errors.New("error creating docs directory")
```
ccoVeille, line 77: See ,https://github.com/go-fuego/fuego/pull/302#discussion_r1896079774

engine.go
```diff
@@ -1,16 +1,139 @@
...
+	}
+
+	f, err := os.Create(jsonSpecLocalPath) // #nosec G304 (file path provided by developer, not by user)
+	if err != nil {
+		return errors.New("error creating file")
```
ccoVeille, line 82: See

https://github.com/go-fuego/fuego/pull/302#discussion_r1896079774

engine.go
EwenQuim: No need to. It is a tab, isn't it? or just `` "\t" ``

engine.go
EwenQuim: It's not critical. And which context OpenAPI Validation should be related to?

EwenQuim (changes requested):

engine.go
```diff
@@ -1,16 +1,139 @@
...
+	"path/filepath"
+)
+
+func NewEngine(config ...OpenAPIConfig) *Engine {
+	if len(config) > 1 {
+		panic("config should not be more than one")
+	}
+	engine := &Engine{
+		OpenAPI:       NewOpenAPI(),
+		OpenAPIConfig: defaultOpenAPIConfig,
```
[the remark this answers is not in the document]
dylanhitt: Yeah @ccoVeille were pretty in agreement above on this.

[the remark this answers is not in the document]
dylanhitt: Yeah @ccoVeille and I were pretty much in agreement on this above. We can just call through til the server.

engine.go
dylanhitt: So I'm happy to do this. This PR was more just moving this code to this file and changing the target of method to be the `Engine`. I'd like to handle this in a separate PR or you're welcome too.

engine.go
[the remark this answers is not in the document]
dylanhitt: Yeah @ccoVeille and I were pretty much in agreement on this above. We can just call through til the server.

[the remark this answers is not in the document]
dylanhitt: One thing though. Not being able to overload Methods in go always makes this unfortunate. So we can't have 

```golang
func WithOpenAPIConfig(openapiConfig OpenAPIConfig) func(*Engine)
func WithOpenAPIConfig(openapiConfig OpenAPIConfig) func(*Server)
```

Any suggestions on naming or we can 

```golang
func WithEngineOptions(options ...func(*Engine)) func(*Server) {
	return func(s *Server) {
		for _, option := range options {
			option(s.Engine)
		}
	}
}
```

engine.go
dylanhitt: Yeah I think I'lll open another PR with `\t`

engine.go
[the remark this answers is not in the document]
dylanhitt: Went ahead and added a commit doing this.

engine.go
dylanhitt: Resolving with #307

engine.go
ccoVeille: https://github.com/getkin/kin-openapi/blob/v0.128.0/openapi3%2Fopenapi3.go#L142

I thought the validate thing was in fuego but it is not.

I'll have to dig further

ccoVeille: https://github.com/search?q=repo%3Agetkin%2Fkin-openapi%20%22context.With%22&type=code

I don't think there is a need for fuego to pass a context, I mean something different than context.Background

engine.go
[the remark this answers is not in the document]
dylanhitt: The last option here would be to a `WithEngine`  option on the server

server.go
```diff
@@ -15,21 +15,26 @@ import (
...
 type OpenAPIConfig struct {
-	DisableSwagger   bool                              // If true, the server will not serve the Swagger UI nor the OpenAPI JSON spec
-	DisableSwaggerUI bool                              // If true, the server will not serve the Swagger UI
-	DisableLocalSave bool                              // If true, the server will not save the OpenAPI JSON spec locally
-	SwaggerUrl       string                            // URL to serve the swagger UI
-	UIHandler        func(specURL string) http.Handler // Handler to serve the OpenAPI UI from spec URL
-	JsonUrl          string                            // URL to serve the OpenAPI JSON spec
-	JsonFilePath     string                            // Local path to save the OpenAPI JSON spec
-	PrettyFormatJson bool                              // Pretty prints the OpenAPI spec with proper JSON indentation
+	EngineOpenAPIConfig
```
dylanhitt, line 18: @EwenQuim 

One issue I have with this structure is that it will place `OpenAPIConfig` options that will have no affect on Engines for other frameworks. I personally think we need a configuration object for our net/http server for such things (so everything that isn't in the current `EngineOpenAPIConfig`) should solely exist on `Server` structure or some other structure that we define that holds these values.

engine.go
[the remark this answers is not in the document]
dylanhitt: Resolving in favor of this comment for discussion as an option concept ahas been implemented:

https://github.com/go-fuego/fuego/pull/302#discussion_r1895996637

engine.go
dylanhitt: Resolving to be dealt with in another PR

engine.go
```diff
@@ -1,16 +1,138 @@
...
+	}
+	return jsonSpec
+}
+
+func (e *Engine) saveOpenAPIToFile(jsonSpecLocalPath string, jsonSpec []byte) error {
+	jsonFolder := filepath.Dir(jsonSpecLocalPath)
+
+	err := os.MkdirAll(jsonFolder, 0o750)
+	if err != nil {
+		return errors.New("error creating docs directory")
```
ccoVeille, line 109:
```suggestion
		return fmt.Errorf("error creating docs directory: %w", err)
```

engine.go
```diff
@@ -1,16 +1,138 @@
...
+	}
+
+	f, err := os.Create(jsonSpecLocalPath) // #nosec G304 (file path provided by developer, not by user)
+	if err != nil {
+		return errors.New("error creating file")
```
ccoVeille, line 114:
```suggestion
		return fmt.Errorf("error creating file: %w", err)
```

engine.go
```diff
@@ -1,16 +1,138 @@
...
+	}
+	defer f.Close()
+
+	_, err = f.Write(jsonSpec)
+	if err != nil {
+		return errors.New("error writing file ")
```
ccoVeille, line 120:
```suggestion
		return fmt.Errorf("error writing file: %w", err)
```

server.go
```diff
@@ -15,21 +15,26 @@ import (
...
+	// If true, the server will not serve the Swagger UI nor the OpenAPI JSON spec
+	DisableSwagger bool
+	// If true, the server will not serve the Swagger UI
+	DisableSwaggerUI bool
+	// URL to serve the swagger UI
+	SwaggerURL string
+	// Handler to serve the OpenAPI UI from spec URL
+	UIHandler func(specURL string) http.Handler
+	// URL to serve the OpenAPI JSON spec
+	JsonURL string
```
ccoVeille, line 28: This one is uneasy because JSONURL would be ugly

This maybe

```suggestion
	OpenAPISpecURL string
```

Or

```suggestion
	SpecURL string
```

Also, do you need exported variable?

server.go
```diff
@@ -15,21 +15,26 @@ import (
...
-	DisableSwaggerUI bool                              // If true, the server will not serve the Swagger UI
-	DisableLocalSave bool                              // If true, the server will not save the OpenAPI JSON spec locally
-	SwaggerUrl       string                            // URL to serve the swagger UI
-	UIHandler        func(specURL string) http.Handler // Handler to serve the OpenAPI UI from spec URL
-	JsonUrl          string                            // URL to serve the OpenAPI JSON spec
-	JsonFilePath     string                            // Local path to save the OpenAPI JSON spec
-	PrettyFormatJson bool                              // Pretty prints the OpenAPI spec with proper JSON indentation
+	EngineOpenAPIConfig
+	// If true, the server will not serve the Swagger UI nor the OpenAPI JSON spec
+	DisableSwagger bool
```
ccoVeille, line 20: The fact DisableSwagger disables OpenAPI is strange to me

What about this?

```suggestion
	Disabled bool
```

engine.go
dylanhitt: Will be implemented in another PR

engine.go
dylanhitt: Will be implemented in another PR

server.go
dylanhitt: Yeah I'll update.

engine.go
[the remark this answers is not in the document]
dylanhitt: WithEngine shown with #310. As @ccoVeille needing a new engine kinda feels bad. This options pattern does feel better in my opinion and again Engine is public on the server a user can always just set it themselves.

One thing I think #310 does well is splitting the OpenApiConfig between the Server and Engine to where it’s pertinent.

engine.go
```diff
@@ -1,16 +1,121 @@
...
-	OpenAPI      *OpenAPI
-	ErrorHandler func(error) error
+	OpenAPI       *OpenAPI
+	ErrorHandler  func(error) error
+	OpenAPIConfig OpenAPIConfig
 
 	acceptedContentTypes []string
 }
+
+type OpenAPIConfig struct {
```
ccoVeille, line 33: One commit has a typo

495aa54a47354777f7bcb026398d8dc518fbea60

ccoVeille (approved):

engine.go
dylanhitt: Ah, thanks I’ll get it in the morning. Really need to stop pushing so late

engine.go
```diff
@@ -1,16 +1,121 @@
...
+	// If true, the server will not serve nor generate any OpenAPI resources
+	Disabled bool
```
EwenQuim, line 35: We will also disable generation at the route registration level (and keep the checks) for faster server start-time (even if it'll save only a few microseconds)

engine.go
```diff
@@ -1,16 +1,121 @@
...
+	e := &Engine{
+		OpenAPI:       NewOpenAPI(),
+		OpenAPIConfig: defaultOpenAPIConfig,
+		ErrorHandler:  ErrorHandler,
+	}
+	for _, option := range options {
+		option(e)
 	}
+	return e
 }
```
EwenQuim, line 22: Good!

server.go
```diff
@@ -75,7 +74,7 @@ type Server struct {
 
 	startTime time.Time
 
-	OpenAPIConfig OpenAPIConfig
+	OpenAPIServerConfig
```
EwenQuim, line 77: Not a fan of embedding if it's not to inherit methods (in order to satisfy an interface) or for backward-compatibility reasons.

By doing useless embedding, users will be able to access via `s.<field>` instead of `s.OpenAPIServerConfig.<field>` and I think it's not a good thing if ther isn't a good reason to do so (conciseness isn't a good one IMHO).

openapi.go
```diff
@@ -112,97 +107,45 @@ func (s *Server) OutputOpenAPISpec() openapi3.T {
...
 		_, _ = w.Write(jsonSpec)
 	})
-	s.printOpenAPIMessage(fmt.Sprintf("JSON spec: %s%s", s.url(), s.OpenAPIConfig.JsonUrl))
+	s.printOpenAPIMessage(fmt.Sprintf("JSON spec: %s%s", s.url(), s.SpecURL))
 
-	if !s.OpenAPIConfig.DisableSwaggerUI {
-		Register(s, Route[any, any]{
+	if s.DisableSwaggerUI {
+		return
+	}
```
EwenQuim, line 127: I love early returns, nice

EwenQuim (changes requested):
Nice PR! We just need to de-embed the OpenAPIServerConfig field.

server.go
dylanhitt: Done

engine.go
dylanhitt: see #312

openapi.go
ccoVeille: I have a bunch of golangci-lint linter rules to enable if you like these

ccoVeille:
> Nice PR! We just need to de-embed the OpenAPIServerConfig field.

I thought 84ee2ff566fa8faa148e8ff10196956c0afc109b would come with a change on the struct to de-embed it really. Here you de-embedded the way the embedded struct is called.

But maybe I'm missing something

dylanhitt:
> I thought https://github.com/go-fuego/fuego/commit/84ee2ff566fa8faa148e8ff10196956c0afc109b would come with a change on the struct to de-embed it really. Here you de-embedded the way the embedded struct is called.

Maybe I'm just misunderstanding the ask. I assumed @EwenQuim just wanted it to be structured the way `OpenAPIConfig` originally was.

ccoVeille:
No, you are fine. I'm sorry. I missed the fact you de-embedded at the struct level

engine.go
dylanhitt: Yeah if you don’t mind.

ccoVeille (approved):

engine.go
ccoVeille: I opened

https://github.com/go-fuego/fuego/issues/321

engine.go
ccoVeille: I opened https://github.com/go-fuego/fuego/issues/321

engine.go
```diff
@@ -1,16 +1,144 @@
...
+	}
+	return jsonSpec
+}
+
+func (e *Engine) saveOpenAPIToFile(jsonSpecLocalPath string, jsonSpec []byte) error {
+	jsonFolder := filepath.Dir(jsonSpecLocalPath)
+
+	err := os.MkdirAll(jsonFolder, 0o750)
+	if err != nil {
+		return errors.New("error creating docs directory")
```
EwenQuim, line 115: Please use `fmt.Errorf("error creating docs directory: %w", err)`

engine.go
```diff
@@ -1,16 +1,144 @@
...
+	}
+
+	f, err := os.Create(jsonSpecLocalPath) // #nosec G304 (file path provided by developer, not by user)
+	if err != nil {
+		return errors.New("error creating file")
```
EwenQuim, line 120: Same

engine.go
```diff
@@ -1,16 +1,144 @@
...
+	}
+	defer f.Close()
+
+	_, err = f.Write(jsonSpec)
+	if err != nil {
+		return errors.New("error writing file ")
```
EwenQuim, line 126: Same

EwenQuim:
I'd like to release this this week. Good work @dylanhitt!

engine.go
dylanhitt: Updated

EwenQuim (approved):

PR #303, opened 2024-12-23
basic api for middleware req/res logging

jwtly10:


server.go
```diff
@@ -423,3 +433,33 @@ func WithRouteOptions(options ...func(*BaseRoute)) func(*Server) {
 		s.routeOptions = append(s.routeOptions, options...)
 	}
 }
+
```
dylanhitt, line 436: Curious about other thoughts on if all these were combined similar to how `WithOpenAPIConfig`

dylanhitt (changes requested):

default_middlewares.go
```diff
@@ -0,0 +1,46 @@
...
+			if !s.loggingConfig.Enabled {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			// Placeholders (will flesh this out later)
+			if !s.loggingConfig.DisableRequest {
+				if s.loggingConfig.RequestLogger != nil {
+					s.loggingConfig.RequestLogger(w, r)
+				} else {
```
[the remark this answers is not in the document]
dylanhitt: Or set all the values in the NewServer constructor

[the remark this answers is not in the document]
jwtly10: Good point, I have addressed these

server.go
jwtly10: Yeah I think this may work. I've implemented it so you guy can review if thats something you prefer.

Its cleaner than having a bunch of WithX

default_middlewares.go
```diff
@@ -0,0 +1,44 @@
+package fuego
+
+import (
+	"log/slog"
+	"net/http"
+)
+
+type LoggingConfig struct {
+	Disabled        bool
```
ccoVeille, line 9: I would like to suggest this


```suggestion
	DisableLogging        bool
```

The Disabled would be nice if there were no other Disable without "d" aside.

I think it would be more readable

default_middlewares.go
```diff
@@ -0,0 +1,44 @@
...
+	DisableRequest  bool
+	DisableResponse bool
+
+	RequestLogger  func(w http.ResponseWriter, r *http.Request)
+	ResponseLogger func(w http.ResponseWriter, r *http.Request)
+}
+
+var defaultLoggingConfig = LoggingConfig{
+	Disabled:       false,
```
ccoVeille, line 18: You can simply put this

```suggestion
```

Or you add the other Disable* one

ccoVeille: But maybe, the Disable should be a method checking if DisableRequest and DisableResponse leads to avoid using the middleware

server.go
```diff
@@ -144,6 +148,10 @@ func NewServer(options ...func(*Server)) *Server {
 		)
 	}
 
+	if !s.loggingConfig.Disabled {
```
ccoVeille, line 151: Please consider my suggestion

https://github.com/go-fuego/fuego/pull/303#discussion_r1896585731

server.go
```diff
@@ -423,3 +431,23 @@ func WithRouteOptions(options ...func(*BaseRoute)) func(*Server) {
 		s.routeOptions = append(s.routeOptions, options...)
 	}
 }
+
+func WithLoggingConfig(loggingConfig LoggingConfig) func(*Server) {
+	return func(s *Server) {
+		if loggingConfig.Disabled {
+			s.loggingConfig.Disabled = true
+			return
+		}
```
dylanhitt, line 440:
```suggestion
		s.loggingConfig.Disabled = loggingConfig.Disabled
```

dylanhitt (changes requested):

default_middlewares.go
jwtly10: @ccoVeille good point, thank you. Disabled does feel a bit redundant.. it can just be inferred. have made some changes

ccoVeille (approved):

dylanhitt:
Looks good. Now time for the part we're all very opinionated about what the logs like 😄

default_middlewares.go
```diff
@@ -0,0 +1,47 @@
...
+			if !s.loggingConfig.DisableResponse {
+				s.loggingConfig.ResponseLogger(w, r)
+			}
+		})
+	}
+}
+
+func logRequest(w http.ResponseWriter, r *http.Request) {
+	slog.Info("<- request")
+}
```
EwenQuim, line 43: Really? This seems a bit light! I would love to see at least the basics

- Method
- Path
- Timestamp

jwtly10: Hey sorry this was just a placeholder for the draft pr while trying out the api for the logging... i plan on cleaning this up tommorow!

default_middlewares.go
```diff
@@ -0,0 +1,47 @@
...
+
+func logResponse(w http.ResponseWriter, r *http.Request) {
+	slog.Info("response ->")
+}
```
EwenQuim, line 47: More complicated but I'd like to see

- Method
- Path
- Timestamp

For another PR:
- Corresponding request ID
- Returned status code
- Timer with the request duration

default_middlewares.go
dylanhitt: Yeah hence my comment

>  Looks good. Now time for the part we're all very opinionated about what the logs like 😄

EwenQuim (changes requested):
Thank you for the PR, it looks promising!

I'd love to see at least the method, path, timestamp & duration of the request!

default_middlewares.go
EwenQuim: Ah my bad the PR is indeed in draft, sorry.

EwenQuim: > Now time for the part we're all very opinionated about what the logs like 😄

Indeed. For the moment, I think, let's just use basic structured logging proposed by the standard library!

If we want to be opiniated on the logging, we'll have to implement a logging frontend (the "displayed", cf [this example](https://github.com/lmittmann/tint)) and keep the backend (the "generated log" structured)

EwenQuim: Something like this

```golang
slog.Info("<- request", "method", r.Method, "url", r.URL.String(), "timestamp", time.Now().Format(time.RFC3339))

```

dylanhitt: I think if we ever want to be opinionated, we should probably just shoot for a full fledged library in middlewares/.

also upvote for a requestid in the default as well

default_middlewares.go
dylanhitt: You’ll probably need to wrap request or response as well to get the duration as well

dylanhitt: This is also where I stopped my original implementation. I’m not really sure it’s worth it to allow users to set the functions themselves. By that point they’re being forced to implement a specific function signature where it’s probably easier to just disable the middleware and bring their own.

EwenQuim: For duration you can just have a timer in the middleware. But I'm not sure that current implementation allows to share data between the first and the second function

ccoVeille: The usual way is indeed a time.Now before the call and a time.Since after the call.

The start time doesn't need to be in the context

jwtly10:
@EwenQuim @dylanhitt 

> This is also where I stopped my original implementation. I’m not really sure it’s worth it to allow users to set the functions themselves. By that point they’re being forced to implement a specific function signature where it’s probably easier to just disable the middleware and bring their own.

Yeah i agree with this and couldn't think of a nice way to get this done. I've created a new pr #313 with addresses all placeholder logs and removes the options to configuring logging funcs. I'll close this one in favour of that

PR #305, opened 2024-12-24
chore: fix examples/gin-compat module pathing

dylanhitt:


dylanhitt:
Yeah ran a go work sync when updating some deps and noticed this, bringing a indirect dependency to `github.com/go-fuego/fuego`.

EwenQuim (approved):

PR #311, opened 2024-12-25
Transform and Validate Gin input

EwenQuim:
Adds the Transform and Validate capabilities of Fuego into the Gin adaptor

### DIFF
--- a/deserialization.go
+++ b/deserialization.go
@@ -104,21 +104,7 @@ func read[B any](context context.Context, dec decoder) (B, error) {
 	}
 	slog.Debug("Decoded body", "body", body)
 
-	body, err = transform(context, body)
-	if err != nil {
-		return body, BadRequestError{
-			Title:  "Transformation Failed",
-			Err:    err,
-			Detail: "cannot transform request body: " + err.Error(),
-		}
-	}
-
-	err = validate(body)
-	if err != nil {
-		return body, err
-	}
-
-	return body, nil
+	return TransformAndValidate(context, body)
 }
 
 // ReadString reads the request body as string.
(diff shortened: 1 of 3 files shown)


dylanhitt (approved):

deserialization.go
```diff
@@ -225,8 +194,12 @@ func transform[B any](ctx context.Context, body B) (B, error) {
...
 		if err != nil {
 			return body, BadRequestError{
+				Title:  "Transformation Failed",
 				Err:    err,
 				Detail: "cannot transform request body: " + err.Error(),
+				Errors: []ErrorItem{
+					{Name: "transformation", Reason: "transformation failed"},
+				},
 			}
 		}
```
ccoVeille, line 204: No tests added for this?

EwenQuim: You're right, I'll add some

EwenQuim: I'll work on another PR specially for error handling in the Transform, I think current implementation isn't ideal anyway

extra/fuegogin/context.go
```diff
@@ -99,12 +103,27 @@ func (c ginContext[B]) SetStatus(code int) {
...
+	}
+	status := c.ginCtx.Writer.Status()
+	if status == 0 {
+		status = c.DefaultStatusCode
+	}
+	if status == 0 {
+		status = 200
+	}
+	c.ginCtx.JSON(status, data)
 	return nil
```
ccoVeille, line 117:
```suggestion
	status := c.ginCtx.Writer.Status()
	if status == 0 {
		status = c.DefaultStatusCode
	}
	if status == 0 {
		status = http.StatusOK
	}
	c.ginCtx.JSON(status, data)
 	return nil
```

Or


```suggestion
	if c.DefaultStatusCode == 0 {
		c.DefaultStatusCode = http.StatusOK
	}
	status := c.ginCtx.Writer.Status()
	if status == 0 {
		status = c.DefaultStatusCode
	}
	c.ginCtx.JSON(status, data)
 	return nil
```

Because otherwise you are doing twice

PR #312, opened 2024-12-26
perf: disable route registration when Engine.OpenAPIConfig.Disabled i…

dylanhitt:
Per comment https://github.com/go-fuego/fuego/pull/302#discussion_r1897496730 in #302. 

Decided to do this in another MR do it's bit more visible.

### DIFF
--- a/openapi_test.go
+++ b/openapi_test.go
@@ -286,7 +286,7 @@ func TestServer_OutputOpenApiSpec(t *testing.T) {
 		require.Error(t, err)
 		require.Nil(t, file)
 	})
-	t.Run("swagger disabled", func(t *testing.T) {
+	t.Run("OpenAPI disabled", func(t *testing.T) {
 		s := NewServer(
 			WithEngineOptions(
 				WithOpenAPIConfig(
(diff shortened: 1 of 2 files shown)


openapi_test.go
```diff
@@ -303,7 +303,7 @@ func TestServer_OutputOpenApiSpec(t *testing.T) {
 		})
 
 		document := s.OutputOpenAPISpec()
-		require.Len(t, document.Paths.Map(), 1)
+		require.Len(t, document.Paths.Map(), 0)
```
ccoVeille, line 306: Require Len 0 is about using Empty


```suggestion
		require.Empty(t, document.Paths.Map())
```

openapi_test.go
```diff
@@ -286,7 +286,7 @@ func TestServer_OutputOpenApiSpec(t *testing.T) {
 		require.Error(t, err)
 		require.Nil(t, file)
 	})
-	t.Run("swagger disabled", func(t *testing.T) {
+	t.Run("openapi disabled", func(t *testing.T) {
```
ccoVeille, line 289: Nit

```suggestion
	t.Run("OpenAPI disabled", func(t *testing.T) {
```

ccoVeille (approved):
LGTM 👍 with minor remarks

ccoVeille (approved):

Issue #265, opened 2024-12-12
Request & Response Logger


EwenQuim:
**Is your feature request related to a problem? Please describe.**

A middleware, enabled by default but desactivable via a server option, that logs:
- when a request enters into the server, showing method, path, params (but not body) and the request id
- when a response goes out of the server, showing the response code and answer time, and the corresponding request id

You can create a new `default_middlewares.go` and the option to activate/desactivate it should be in the `server.go` file

dylanhitt:
couple things to add and question. 

Please allow for disablement of one logger or other. In an environment that uses a saas product for their monitoring infra logging both the request and the response can cause a lot of costs! Also I assume the default should be with slog?

dylanhitt:
One more though on request id here. What are the thoughts checking if the header of `X-Request-ID` is on the if there use it if not generate a random uuid?

EwenQuim:
Yes, yes and yes! Good ideas :) 

Also, if the request can be DEBUG and the response INFO, it would be nice !

dylanhitt:
Yeah I agree with you after playing around a bit tonight. The request log should just be debug. Trying to make everything configurable isn't worth the squeeze. It ends up just being easier to create your own logger middleware.

dylanhitt:
instead of `default_middlewares.go` I think it should go into `middlewares/logger` as well.

EwenQuim:
But it would mean in another module an not included by default, right?

dylanhitt:
Hmmm, the more I think about this the more I’m realizing I’m thinking about ding too much.

yeah, let’s just to the default that you described. With a simple on/off. I am imagining middlewares/logger could exist eventually, but that would be a separate effort outside of fuego core. I think this default logger should aim for being just enough.

EwenQuim:
Yeah just a basic logger will be enough. For advanced use cases, users won't be afraid to create their own.

jwtly10:
Hey @dylanhitt @EwenQuim 

Am I OK to work on this issue?

One question about configurability - do we want to selectively enable/disable either the resp or req logs? Or just a global flag that turns off this middleware completely, so anyone is free to impl their own?

EwenQuim:
I would say just a global on/off. But if you find a way to do both without too much downsides, you can do it.

I'm open to suggestions, do as you wish! :)

dylanhitt:
[...]

PR #313, opened 2024-12-26
re-implementing default logging middleware

jwtly10:
This PR implements a simplified default logging middleware, taking on feedback from #303.

These changes simplify #303 and just provides default req/res logging, with options to disable either. Rather than supporting custom logging functions, users who need custom logging can just implement their own mw.

I see some of these use cases:
1. Use default middleware entirely
2. Disable fully/partially disable req&res logging
3. Disable and implement their own logging middleware
4. Use our middleware but override logging funcs

After thinking about how to do this, supporting case 4 does seems unnecessary and more complex that it needs to be - for users needing custom logging it would be cleaner to just implement their own.

Note: Some tests are failing for now due to some golden files not expecting this new middleware by default.

### DIFF
--- a/default_middlewares.go
+++ b/default_middlewares.go
@@ -1,0 +1,109 @@
+package fuego
+
+import (
+	"log/slog"
+	"net/http"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type LoggingConfig struct {
+	DisableRequest  bool // If true, request logging is disabled
+	DisableResponse bool // If true, response logging is disabled
+}
+
+// responseWriter wraps [http.ResponseWriter] to capture response metadata.
+// Implements [http.ResponseWriter.Write] to ensure proper status code capture for implicit 200 responses
+type responseWriter struct {
+	http.ResponseWriter
+	status      int
+	wroteHeader bool
+}
+
+func wrapResponseWriter(w http.ResponseWriter) *responseWriter {
+	return &responseWriter{ResponseWriter: w}
+}
+
+func (rw *responseWriter) Status() int {
+	return rw.status
+}
+
+func (rw *responseWriter) WriteHeader(code int) {
+	if rw.wroteHeader {
+		return
+	}
+
+	rw.status = code
+	rw.ResponseWriter.WriteHeader(code)
+	rw.wroteHeader = true
+}
+
+func (rw *responseWriter) Write(b []byte) (int, error) {
+	if !rw.wroteHeader {
+		rw.WriteHeader(http.StatusOK)
+	}
+	return rw.ResponseWriter.Write(b)
+}
+
+func (l *LoggingConfig) Disabled() bool {
+	return l.DisableRequest && l.DisableResponse
+}
+
+// By default, all logging is enabled
+var defaultLoggingConfig = LoggingConfig{}
+
+// defaultLoggingMiddleware is this default middleware that logs incoming requests and outgoing responses.
+//
+// By default, request logging will be logged at the debug level, and response
+// logging will be logged at the info level
+//
+// Log levels managed by [WithLogHandler]
+func defaultLoggingMiddleware(s *Server) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			start := time.Now()
+
+			requestID := r.Header.Get("X-Request-ID")
+			if requestID == "" {
+				requestID = uuid.New().String()
+			}
+			w.Header().Set("X-Request-ID", requestID)
+
+			wrapped := wrapResponseWriter(w)
+
+			if !s.loggingConfig.DisableRequest {
+				logRequest(requestID, r)
+			}
+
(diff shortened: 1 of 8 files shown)


default_middlewares.go
```diff
@@ -0,0 +1,109 @@
...
+
+// responseWriter wraps [http.ResponseWriter] to capture response metadata.
+// Implements [http.ResponseWriter.Write] to ensure proper status code capture for implicit 200 responses
+type responseWriter struct {
+	http.ResponseWriter
+	status      int
+	wroteHeader bool
+}
+
+func wrapResponseWriter(w http.ResponseWriter) *responseWriter {
```
EwenQuim, line 24: `newResponseWriter` to keep Go conventions please

default_middlewares.go
```diff
@@ -0,0 +1,109 @@
...
+//
+// Log levels managed by [WithLogHandler]
+func defaultLoggingMiddleware(s *Server) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			start := time.Now()
+
+			requestID := r.Header.Get("X-Request-ID")
+			if requestID == "" {
+				requestID = uuid.New().String()
```
EwenQuim, line 69: Introducing a new lib?

`go mod tidy` please

EwenQuim (changes requested):
Very good PR!

Looking very close to what I expected, thank you very much.

I'd let the possibility for users to provide their own requestId, that's a really common use case. Just add a `func() string` in LoggingConfig :)

Thank you again!

EwenQuim:
Also, I'd add some tests with `slogassert`. There are some test existing in the codebase that you can mimic.

default_middlewares.go
```diff
@@ -0,0 +1,109 @@
...
+		"method", r.Method,
+		"path", r.URL.Path,
+		"timestamp", time.Now().Format(time.RFC3339),
+		"remote_addr", r.RemoteAddr,
+		"user_agent", r.UserAgent(),
+	)
+}
+
+func logResponse(r *http.Request, rw *responseWriter, requestID string, duration time.Duration) {
+	slog.Info("outgoing response",
```
dylanhitt, line 101: Since the Request log is debug. It's probably useful to include `"remote_addr", r.RemoteAddr,` here as well.

default_middlewares.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"log/slog"
+	"net/http"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type LoggingConfig struct {
+	DisableRequest  bool // If true, request logging is disabled
```
dylanhitt, line 12: Place comments above for readability if you don't mind 😄

dylanhitt (changes requested):

jwtly10:
The only thing I haven't addressed yet is failing tests due to including the new mw in open api docs:

``` go
 expected: "#### Controller: \n\n`github.com/go-fuego/fuego_test.helloWorld`\n\n---\n\ntest description"
 actual  : "#### Controller: \n\n`github.com/go-fuego/fuego_test.helloWorld`\n\n#### Middlewares:\n\n- `github.com/go-fuego/fuego.NewServer.defaultLoggingMiddleware.func16`\n\n---\n\ntest description"
```


./examples/petstore/lib/server_test.go: (I guess this would be best fixed with simple golden file update)

openapi_operations_test.go/options_test.go

I'm not sure what the best way to approach this is for these tests. I guess 2 options off the top of my head are 
1. disable the middleware for these tests 
2. update all the tests to handle this default case of having middleware attached as standard (will also need to change defaultLoggingMiddleware from being an anonymouse function to stop `.func16` showing in docs)

Any other ideas?

EwenQuim:
Please update the tests with `make golden-update`. The petstore example is an example of Fuego offers out-of-the-box, so it makes sense that the middleware is registered and shown

EwenQuim:
> will also need to change defaultLoggingMiddleware from being an anonymouse function to stop .func16 showing in docs

Yes and we need a goodNaming. I suggest `defaultFuegoLogger`? @dylanhitt any idea? If @jwtly10 you have a better idea, don't hesitate :)

jwtly10:
@EwenQuim thanks for the feedback. defaultLogger makes sense for now. I guess it can be changed in future if needed.

Changes have been made, all tests are now passing :)

dylanhitt:
On mobile, can you send a picture of the way it looks in stoplight?

jwtly10:
@dylanhitt Not sure what you mean? Do you mean how the open api spec renders when it has this middleware? (using stoplight on mobile?)

dylanhitt:
Oh no, the screenshot is what I needed. Thank you. I like this

dylanhitt (approved):

dylanhitt (approved):

dylanhitt:
I believe @EwenQuim request changes have been addressed. This closes #265

Thank you for the work @jwtly10! We appreciate it.

Cheers

PR #314, opened 2024-12-26
BREAKING: WithErrorHandler server option to engine option

dylanhitt:
Moves WithErrorHandler to an engine option.

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -1,0 +1,38 @@
+package fuego
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithErrorHandler(t *testing.T) {
+	t.Run("default engine", func(t *testing.T) {
+		e := NewEngine()
+		err := NotFoundError{
+			Err: errors.New("Not Found :c"),
+		}
+		errResponse := e.ErrorHandler(err)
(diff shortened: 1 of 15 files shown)


engine_test.go
```diff
@@ -0,0 +1,31 @@
...
+func TestWithErrorHandler(t *testing.T) {
+	t.Run("default engine", func(t *testing.T) {
+		e := NewEngine()
+		err := NotFoundError{
+			Err: errors.New("Not Found :c"),
+		}
+		errResponse := e.ErrorHandler(err)
+		require.ErrorAs(t, errResponse, &HTTPError{})
+	})
+	t.Run("custom handler", func(t *testing.T) {
```
ccoVeille, line 19: Please also test what happens when you try to pass nil as an ErrorHandler

dylanhitt: I was thinking about this. I would rather just panic in the With function. I think we should do that in all cases where the input can’t be nil

ccoVeille: That's my point, yes. Testing it allows to discuss about what should be done.

I'm not fan of panic, but yes there should be some error management.

dylanhitt: The other option is to check nil and default to our ErrorHandler which I also don't like. The pattern to panic here has been established in previous option func in both routes and Server.

ccoVeille:
Closed in favor of #320

PR #319, opened 2024-12-29
fix: nested struct tags should be respected in openapi spec

### LINKED ISSUE #318: Tags for Nested Structures Are Not Properly Reflected in OpenAPI Specification Generation
When extracting a structure and using it as an attribute in another structure, the OpenAPI specification generation does not consider the tags for the external structure. Specifically, tags such as example and description for the nested structure (References in this case) are not being included in the generated OpenAPI specification.

Example:
Here is a sample Go package illustrating the issue:

```
package models

type Pets struct {
    ID         string     `json:"id" validate:"required" example:"pet-123456"`
    Name       string     `json:"name" validate:"required" example:"Napoleon"`
    Age        int        `json:"age" example:"2" description:"Age of the pet, in years"`
    IsAdopted  bool       `json:"is_adopted" description:"Is the pet adopted"`
    References References `json:"references"`
}

type PetsCreate struct {
    Name       string     `json:"name" validate:"required,min=1,max=100" example:"Napoleon"`
    Age        int        `json:"age" validate:"min=0,max=100" example:"2" description:"Age of the pet, in years"`
    IsAdopted  bool       `json:"is_adopted" description:"Is the pet adopted"`
    References References `json:"references"`
}

type PetsUpdate struct {
    Name       string     `json:"name,omitempty" validate:"min=1,max=100" example:"Napoleon" description:"Name of the pet"`
    Age        int        `json:"age,omitempty" validate:"max=100" example:"2"`
    IsAdopted  *bool      `json:"is_adopted,omitempty" description:"Is the pet adopted"`
    References References `json:"references"`
}

type References struct {
    Type  string `json:"type" validate:"required" example:"pet-123456"`
    Value string `json:"value"`
}
```

Observed Behavior:
The OpenAPI specification generated does not include the example and description tags for the nested structure References.

Expected Behavior:
The generated OpenAPI specification should reflect all tags (e.g., example and description) for nested structures, ensuring the documentation is complete and accurate.

image

dylanhitt:
closes #318 

@ccoVeille Mind taking a look.

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -433,7 +433,6 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
 
 	for i := range t.NumField() {
 		field := t.Field(i)
-
 		if field.Anonymous {
 			fieldType := field.Type
 			parseStructTags(fieldType, schemaRef)
@@ -453,6 +452,9 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
 		if property == nil {
 			slog.Warn("Property not found in schema", "property", jsonFieldName)
 			continue
+		}
+		if field.Type.Kind() == reflect.Struct {
+			parseStructTags(field.Type, property)
 		}
 		propertyCopy := *property
 		propertyValue := *propertyCopy.Value
--- a/openapi_test.go
+++ b/openapi_test.go
@@ -11,12 +11,20 @@ import (
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/stretchr/testify/require"
+	"gotest.tools/v3/assert"
 )
 
 type MyStruct struct {
 	B string `json:"b"`
-	C int    `json:"c"`
+	C int    `json:"c" example:"8" validate:"min=3,max=10" description:"my description"`
 	D bool   `json:"d"`
+}
+
+type MyStructWithNested struct {
+	E      string   `json:"e" example:"E"`
+	F      int      `json:"f"`
+	G      bool     `json:"g"`
+	Nested MyStruct `json:"nested" description:"my struct"`
 }
 
 type MyOutputStruct struct {
@@ -52,6 +60,14 @@ func Test_tagFromType(t *testing.T) {
 			inputType:   MyStruct{},
 
 			expectedTagValue:     "MyStruct",
+			expectedTagValueType: &openapi3.Types{"object"},
+		},
+		{
+			name:        "nested struct",
+			description: "",
+			inputType:   MyStructWithNested{},
+
+			expectedTagValue:     "MyStructWithNested",
 			expectedTagValueType: &openapi3.Types{"object"},
 		},
 		{
@@ -201,6 +217,21 @@ func Test_tagFromType(t *testing.T) {
 			}
 		})
 	}
+
+	t.Run("struct with nested tags", func(t *testing.T) {
+		s := NewServer()
+		tag := SchemaTagFromType(s.OpenAPI, MyStructWithNested{})
+		nestedProperty := tag.Value.Properties["nested"]
+		require.NotNil(t, nestedProperty)
+		assert.Equal(t, "my struct", nestedProperty.Value.Description)
+		c := nestedProperty.Value.Properties["c"]
+		require.NotNil(t, c)
+		require.NotNil(t, c.Value)
+		assert.Equal(t, "my description", c.Value.Description)
+		assert.Equal(t, 8, c.Value.Example)
+		assert.Equal(t, float64(3), *c.Value.Min)
+		assert.Equal(t, float64(10), *c.Value.Max)
+	})
 }
 
 func TestServer_generateOpenAPI(t *testing.T) {
--- a/openapi_test.go
+++ b/openapi_test.go
@@ -2,16 +2,19 @@ package fuego
 
 import (
 	"bytes"
+	"encoding/xml"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"testing"
 
 	"github.com/getkin/kin-openapi/openapi3"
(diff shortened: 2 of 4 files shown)


openapi_test.go
```diff
@@ -15,10 +15,17 @@ import (
...
-	C int    `json:"c"`
+	C int    `json:"c" example:"8" validate:"min=3,max=10" description:"my description"`
 	D bool   `json:"d"`
 }
 
+type MyStructWithNested struct {
+	E      string   `json:"e" example:"E"`
+	F      int      `json:"f"`
+	G      bool     `json:"g"`
+	Nested MyStruct `json:"nested" description:"my struct"`
```
ccoVeille, line 26: I'm curious about what would bring a embedded struct

So please add something comparable to this to the examples or tests

```suggestion
	Nested MyStruct `json:"nested" description:"my struct"`
	MyStruct
```

I'm not expecting to fail, but I would prefer to see it covered by a test, this way any future refactoring would be safer

Please note, I have no idea whether it's already covered by an existing unit test. If so, then my remark is pointless maybe

openapi_test.go
```diff
@@ -201,6 +216,17 @@ func Test_tagFromType(t *testing.T) {
...
+
+	t.Run("struct with nested tags", func(t *testing.T) {
+		s := NewServer()
+		tag := SchemaTagFromType(s.OpenAPI, MyStructWithNested{})
+		nestedProperty := tag.Value.Properties["nested"]
+		require.Equal(t, "my struct", nestedProperty.Value.Description)
+		require.Equal(t, "my description", nestedProperty.Value.Properties["c"].Value.Description)
+		require.Equal(t, 8, nestedProperty.Value.Properties["c"].Value.Example)
+		require.Equal(t, float64(3), *nestedProperty.Value.Properties["c"].Value.Min)
+		require.Equal(t, float64(10), *nestedProperty.Value.Properties["c"].Value.Max)
```
ccoVeille, line 228: This would panic if c is not present


```suggestion
		require.Contains(t, nestedProperty.Value.Properties, "c")
		require.Equal(t, "my description", nestedProperty.Value.Properties["c"].Value.Description)
		require.Equal(t, 8, nestedProperty.Value.Properties["c"].Value.Example)
		require.Equal(t, float64(3), *nestedProperty.Value.Properties["c"].Value.Min)
		require.Equal(t, float64(10), *nestedProperty.Value.Properties["c"].Value.Max)
```

ccoVeille, line 228: This is an abuse of using testify require package

You should use assert package here. require fails on first error, so it won't bring you good information about what is broken, you will have to fix them one by one

But I think the problem is more general

openapi_test.go
```diff
@@ -201,6 +216,17 @@ func Test_tagFromType(t *testing.T) {
 			}
 		})
 	}
+
+	t.Run("struct with nested tags", func(t *testing.T) {
+		s := NewServer()
+		tag := SchemaTagFromType(s.OpenAPI, MyStructWithNested{})
+		nestedProperty := tag.Value.Properties["nested"]
+		require.Equal(t, "my struct", nestedProperty.Value.Description)
```
ccoVeille, line 224: Pleawtry to avoid panic on error

```suggestion
		tag := SchemaTagFromType(s.OpenAPI, MyStructWithNested{})
		require.Contains(t, tag.Value.Properties, "nested")
		nestedProperty := tag.Value.Properties["nested"]
		require.Equal(t, "my struct", nestedProperty.Value.Description)
```

openapi.go
```diff
@@ -454,6 +453,10 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
 			slog.Warn("Property not found in schema", "property", jsonFieldName)
 			continue
 		}
+		if field.Type.Kind() == reflect.Struct {
+			fieldType := field.Type
+			parseStructTags(fieldType, property)
```
ccoVeille, line 458: You added no tests about validating the error returned by parseStruct that are logged.

You should check they are reported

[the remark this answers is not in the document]
dylanhitt: Pedantic. But changed

openapi_test.go
dylanhitt: I don't understand the ask.

openapi_test.go
dylanhitt: Updated

dylanhitt: updated for this case

openapi_test.go
dylanhitt: Oh. Okay. Are  just saying check if tag is nil?

openapi_test.go
dylanhitt: added

openapi_test.go
dylanhitt: I believe this concern is covered now with a fix from a previous concern

```golang
nestedProperty := tag.Value.Properties["nested"]
require.NotNil(t, nestedProperty)
```

openapi.go
dylanhitt: Add a test function using slogassert for warn cases

openapi_test.go
```diff
@@ -2,28 +2,48 @@ package fuego
...
+	"log/slog"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"testing"
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/stretchr/testify/require"
+	"github.com/thejerf/slogassert"
+	"gotest.tools/v3/assert"
```
ccoVeille, line 17: I think you wanted to use testify/assert and your IDE provided a random lib

ccoVeille (changes requested):

openapi_test.go
dylanhitt: Updated. I believe this is because gotest is in our base go.mod do to the petstore example not being it's own module. I'll be fixing that in the another PR

ccoVeille (approved):

PR #320, opened 2024-12-29
BREAKING: WithErrorHandler server option to engine option

dylanhitt:
Moves WithErrorHandler to an engine option.

Closed #314 in favor of this as I created a mess over there.

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -1,0 +1,40 @@
+package fuego
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithErrorHandler(t *testing.T) {
+	t.Run("default engine", func(t *testing.T) {
+		e := NewEngine()
+		err := NotFoundError{
+			Err: errors.New("Not Found :c"),
+		}
+		errResponse := e.ErrorHandler(err)
+		require.ErrorAs(t, errResponse, &HTTPError{})
+	})
+	t.Run("custom handler", func(t *testing.T) {
+		e := NewEngine(
+			WithErrorHandler(func(err error) error {
+				return fmt.Errorf("%w foobar", err)
+			}),
+		)
+		err := NotFoundError{
+			Err: errors.New("Not Found :c"),
+		}
+		errResponse := e.ErrorHandler(err)
+		require.ErrorAs(t, errResponse, &HTTPError{})
+		require.ErrorContains(t, errResponse, "Not Found :c foobar")
+	})
+	t.Run("should be fatal", func(t *testing.T) {
+		require.Panics(t, func() {
+			NewEngine(
+				WithErrorHandler(nil),
+			)
+		})
+	})
+}
--- a/engine.go
+++ b/engine.go
@@ -71,6 +71,17 @@ func WithOpenAPIConfig(config OpenAPIConfig) func(*Engine) {
 	}
 }
 
+// WithErrorHandler sets a customer error handler for the server
+func WithErrorHandler(errorHandler func(err error) error) func(*Engine) {
+	return func(e *Engine) {
+		if errorHandler == nil {
+			panic("errorHandler cannot be nil")
+		}
+
+		e.ErrorHandler = errorHandler
+	}
+}
+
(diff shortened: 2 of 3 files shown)


engine_test.go
```diff
@@ -0,0 +1,38 @@
...
+			WithErrorHandler(func(err error) error {
+				return errors.New("")
+			}),
+		)
+		err := NotFoundError{
+			Err: errors.New("Not Found :c"),
+		}
+		errResponse := e.ErrorHandler(err)
+		require.NotErrorAs(t, errResponse, &HTTPError{})
+	})
```
ccoVeille, line 30:
```suggestion
	t.Run("custom handler", func(t *testing.T) {
		errSentinel := errors.New("foo bar")
		e := NewEngine(
			WithErrorHandler(func(err error) error {
				return errSentinel
			}),
		)
		err := NotFoundError{
			Err: errors.New("Not Found :c"),
		}
		errResponse := e.ErrorHandler(err)
		require.NotErrorAs(t, errResponse, &HTTPError{})
		require.ErrorIs(t, errResponse, errSentinel)
	})
```

engine_test.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+		errResponse := e.ErrorHandler(err)
+		require.ErrorAs(t, errResponse, &HTTPError{})
+	})
+	t.Run("custom handler", func(t *testing.T) {
+		e := NewEngine(
+			WithErrorHandler(func(err error) error {
+				return errors.New("")
+			}),
+		)
```
ccoVeille, line 24: Please add a test about a custom error handler that could be used to invalidate an error by returning nil

engine_test.go
ccoVeille, line 30: Please add a test about a custom wrapper that wrap the initial one

```suggestion
	t.Run("custom handler", func(t *testing.T) {
		e := NewEngine(
			WithErrorHandler(func(err error) error {
				return fmt.Errorf("%w foobar", err)
			}),
		)
		err := NotFoundError{
			Err: errors.New("Not Found :c"),
		}
		errResponse := e.ErrorHandler(err)
		require.ErrorAs(t, errResponse, &HTTPError{})
		require.ErrorIs(t, errResponse, err)
		require.ErrorContains(t, errResponse, "foobar")
	})
```

engine_test.go
dylanhitt: Why?

engine_test.go
dylanhitt: ermmm okay sure

engine_test.go
ccoVeille: Because people may do that in their custom handler if you dont want to support it, the mechanism of error handling should be rewritten to forbid this behavior

engine_test.go
dylanhitt: You suggestion doesn't pass.

engine_test.go
dylanhitt: That is not the responsibility of the function being tested. That would need to be tested [here](https://github.com/go-fuego/fuego/blob/2d6cfa42ba8daa774d1c34b7beea66c0772b478d/serve.go#L98) to give it any sort of meaning.

engine_test.go
ccoVeille: That the custom error handler is receiving the error and is able to manipulate it.

engine_test.go
dylanhitt: You're totally right. But I'm not adding this functionality. I'm simply moving the function. I'd prefer an issue is opened to cover this case rather than it be within this PR as that is not the spirit of this unit of work.

engine_test.go
dylanhitt: Okay sure. I'll update the suggestion to work. fmt.ErrorF actually creates `wrapError` which is why the above doesn't work

dylanhitt: Updated

engine_test.go
ccoVeille: Yes, of course. I can understand that.

Let me open an issue

ccoVeille (approved):

EwenQuim (approved):

PR #325, opened 2024-12-30
Replace WithCorsMiddleware by WithGlobalMiddleware

ekumamatthew:
refactor: replace WithCorsMiddleware by WithGlobalMiddleware


server.go
```diff
@@ -171,22 +175,30 @@ func WithTemplateFS(fs fs.FS) func(*Server) {
...
-//		}).Handler)
+//		}).Handler),
 //	)
+func WithGlobalMiddleware(middlewares ...func(http.Handler) http.Handler) func(*Server) {
+	return func(c *Server) {
+		c.globalMiddlewares = append(c.globalMiddlewares, middlewares...)
+	}
+}
+
+// WithCorsMiddleware is deprecated: Use WithGlobalMiddleware instead.
```
EwenQuim, line 199: Please use the same syntax than all other deprecated notices, the one understood by Go editors and godoc : skip line then write `Deprecated: xxx`.

Also, please reference `[WithGlobalMiddlewares]` with brackets so it can be recognized by other Go tools like Godoc :)

EwenQuim (changes requested):
Thanks for your work 🙏🏼

Please write a tests that shows that it works. You can also update the tests that are about the cors middleware.

And please update the PR title to something meaningful lol.

net_http_mux.go
```diff
@@ -232,4 +232,4 @@ var _ Registerer[string, any] = netHttpRouteRegisterer[string, any]{}
 
 func (a netHttpRouteRegisterer[T, B]) Register() Route[T, B] {
 	return *Register(a.s, a.route, a.controller)
-}
+}
```
ccoVeille, line 235: Why such a change?

You could add a commit if it's about formatting files

server.go
```diff
@@ -46,6 +46,10 @@ type Server struct {
 	// For example, it allows OPTIONS /foo even if it is not declared (only GET /foo is declared).
 	corsMiddleware func(http.Handler) http.Handler
```
ccoVeille, line 47: Do you still need this then?

EwenQuim: True, as it is not exposed, we can remove it, no need for deprecation notice!

ekumamatthew:
fuego

test output

EwenQuim (approved):
I would love to see tests for this new Option.

EwenQuim:
Merging to a temp branch "global-middlewares" to fix formatting and tests.

Please do it by yourself next time @ekumamatthew ;)

Thank you for your first contribution

PR #328, opened 2024-12-30
Adding basic support for XML structure

DimShadoWWW:
This pull request introduces basic support for XML tags in OpenAPI documentation, allowing the following features to be defined for schema fields:

1. Custom XML Element Names:
   The `xml:"customName"` struct tag allows specifying a custom name for the XML element.
1. Attribute Definition:
   The `xml:"name,attr"` struct tag marks a field as an XML attribute.
1. Wrapped Elements:
   The `xml:"name,wrapped"` struct tag enables support for wrapping elements (e.g., `<items><item>...</item></items>`).


openapi.go
```diff
@@ -457,6 +457,39 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
...
+			parseStructTags(field.Type, schemaRef.Value.Properties[jsonFieldName])
+		}
+
+		// basic Xml tag support
+		//
+		// We support the following xml tags:
+		// - "xml" field tag: specifies the XML element name
+		// - "xml=attr" field tag: specifies that the field is an XML attribute
+		// - "xml=wrapped" field tag: specifies that the field is wrapped in an XML element
+		xmlTag, ok := field.Tag.Lookup("xml")
```
dylanhitt, line 470: Why not just use `Get`?

openapi.go
```diff
@@ -457,6 +457,39 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
...
+		if ok {
+			xmlTagName := strings.Split(xmlTag, ",")[0] // remove omitempty, etc
+			// if xml tag name is "-", don't add it to the schema
+			if xmlTagName != "-" {
```
dylanhitt, line 474: Instead of nesting more if can just do 

```
if xmlTagName == "-" {
	continue
}
```

dylanhitt (commented):
Would be nice to split the json tag stuff and the xml tag stuff into helper functions. This function is getting rather large.

openapi.go
DimShadoWWW: If we use `continue`, all subsequent code (such as handling examples, validation, and descriptions) will be skipped if the field isn't included in XML.  

In this section, only XML-related information is processed.

openapi.go
DimShadoWWW: updated in commit 45b88a0

openapi.go
dylanhitt: I believe we would then need to change the JSON check above to handle the same case. I.E when the xml tag provide but not the json tag.

dylanhitt: Also I think it would make it a bit cleaner to perform both these actions in helpers so we can early return. 

We also will need to handle the case where both the `xml` and `json` tags are `-` we wouldn't want the description, examples, or validation processed. Kinda like how we do already with the continue in the `json` check for `-`. Would probably be worth to check that early for both so we can continue early.

openapi_test.go
```diff
@@ -39,6 +39,16 @@ type MyOutputStruct struct {
 	Quantity int    `json:"quantity"`
 }
 
+type MyOutputStructDetails struct {
+	IngressDate string `json:"in_date" xml:"date,attr"`
+}
+
+type MyOutputStructWithXmlAttribute struct {
```
dylanhitt, line 46: I think we should add a test for a struct with just xml attributes. As well as one with `json:"-"`

dylanhitt (changes requested):

openapi.go
```diff
@@ -459,6 +459,38 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
 		propertyCopy := *property
 		propertyValue := *propertyCopy.Value
 
+		if field.Type.Kind() == reflect.Struct {
+			parseStructTags(field.Type, schemaRef.Value.Properties[jsonFieldName])
+		}
+
+		// basic Xml tag support
```
ccoVeille, line 466: Nit

```suggestion
		// basic XML tag support
```
Or


```suggestion
		// basic xml tag support
```

openapi.go
```diff
@@ -459,6 +459,38 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
...
+			}
+
+			xmlFields := strings.Split(xmlField, ",")
+			if slices.Contains(xmlFields, "attr") {
+				propertyValue.XML.Attribute = true
+			}
+			if slices.Contains(xmlFields, "wrapped") {
+				propertyValue.XML.Wrapped = true
+			}
+		}
```
ccoVeille, line 492: I would move this in a dedicated method, and also move the code related to json tag parsing to its own method too.

Because now the code is too complex.

Also it would allow you to return with early return when needed. It would avoid imbricated if levels

DimShadoWWW: I'm working on it

DimShadoWWW: @ccoVeille In the next commit, I separated the logic into individual methods and introduced a `MetadataParsers` struct to manage their registration and define a specific processing order. This ensures better control in cases where different methods may update the same value in the OpenAPI specification.

openapi_metadata.go
ccoVeille, line 1: I don't think you need the `openapi_` prefix for this file, everything is about openapi.

Also metadata is not clear enough for a file name I think (with or without the `openapi_`prefix)

Maybe metadata parser of something like that.

Naming is hard

openapi_metadata.go
```diff
@@ -0,0 +1,344 @@
...
+	return nil
+}
+
+func (mp *MetadataParsers) insertAtStart(entry MetadataParserEntry) {
+	mp.registeredParsers = append([]MetadataParserEntry{entry}, mp.registeredParsers...)
+}
+
+func (mp *MetadataParsers) insertAtEnd(entry MetadataParserEntry) {
+	mp.registeredParsers = append(mp.registeredParsers, entry)
+}
```
ccoVeille, line 149: Append and prepend are more commonly used than insert at start/end

openapi_metadata.go
```diff
@@ -0,0 +1,344 @@
...
+// before or after. If the relative parser is not found, an error is returned. If the
+// parser is already registered, the function does nothing and returns nil. The
+// function is thread-safe and ensures that the registeredParsers slice and
+// registeredNames map are accessed in a thread-safe manner.
+func (mp *MetadataParsers) RegisterMetadataParser(name string, parser MetadataParserFunction, position string, relativeTo string) error {
+	validPositions := map[string]bool{"start": true, "end": true, "before": true, "after": true}
+	if !validPositions[position] {
+		return fmt.Errorf("Invalid position. Use 'start', 'end', 'before', or 'after'")
+	}
+	if (position == "before" || position == "after") && mp.findParserIndex(relativeTo) == -1 {
```
ccoVeille, line 109: Before, after, end, start should be constants prefixed with position or something

openapi_metadata.go
```diff
@@ -0,0 +1,344 @@
...
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+var DefaultParsers = []MetadataParserEntry{
+	{Name: "exampleParser", Parser: ExampleMetadataParser},
+	{Name: "validationParser", Parser: ValidationMetadataParser},
+	{Name: "descriptionParser", Parser: DescriptionMetadataParser},
+	{Name: "XMLParser", Parser: XMLMetadataParser},
+	{Name: "JSONParser", Parser: JSONMetadataParser},
```
ccoVeille, line 20: Shouldn't these names be constants prefixed with metadataParser

ccoVeille (commented):
Amazing work you did from a simple comment I made (am I right?)

A few remarks

Also. here is a more general feedback, I would have put all these metadata parser things in a package in internal to do not export them (I'm unsure it's possible)

openapi_metadata.go
DimShadoWWW: I don't think we should hide the metadata parsers in an internal package. Some users might want to use the default parsers in the defined order, while others may prefer to redefine specific parsers or adjust the execution order to fit their needs.  

For example, users might add a custom metadata parser to handle enums and generate OpenAPI documentation from it:  
```
Status string `json:"status" openapi:"oneof=active:Currently active;inactive:No longer active;pending:Pending activation"`
```
As mentioned in my comment on issue [222](https://github.com/go-fuego/fuego/issues/222#issuecomment-2555931137)

In cases where multiple parsers modify the same OpenAPI value, the last parser executed takes precedence. This approach provides flexibility to accommodate various use cases and ensures extensibility

openapi_metadata.go
DimShadoWWW: done

ccoVeille: Your change is great.

I'm sorry if my comment wasn't clear I was talking about the string currently in the Name fields, these names are then used in various places. That's why I asked to use constants and then to use a prefix not a suffix for them. Allowing easier code completions in IDE

ccoVeille (approved):

openapi_test.go
DimShadoWWW: I still need to work on this.

DimShadoWWW:
Added a new commit to PR to fix structure for []struct, where the Type should change to "array" and move properties to items

openapi_metadata.go
```diff
@@ -0,0 +1,423 @@
...
+var DefaultParsers = []MetadataParserEntry{
+	{Name: metadataParserExample, Parser: MetadataParserExample},
+	{Name: metadataParserValidation, Parser: MetadataParserValidation},
+	{Name: metadataParserDescription, Parser: MetadataParserDescription},
+	{Name: metadataParserXML, Parser: MetadataParserXML},
+	{Name: metadataParserJSON, Parser: MetadataParserJSON},
+}
+
+type MetadataParserParams struct {
+	Field      reflect.StructField    // Reflective information of a struct field
```
dylanhitt, line 33: Mind placing comments above fields?

openapi_metadata.go
```diff
@@ -0,0 +1,423 @@
...
+	if slices.Contains(xmlFields, "wrapped") {
+		params.Property.XML.Wrapped = true
+	}
+
+	if params.Field.Type.Kind() == reflect.Struct {
+		if params.Property.Properties == nil {
+			params.Property.Properties = make(map[string]*openapi3.SchemaRef)
+		}
+		processXMLStructChildren(params)
+	} else if params.Field.Type.Kind() == reflect.Slice {
```
dylanhitt, line 348: Assume this would be similar to [`dive`](https://github.com/go-fuego/fuego/blob/main/openapi.go#L292). You're gonna want to check for `Array` here as well.

openapi_metadata.go
```diff
@@ -0,0 +1,423 @@
...
+		elemType := params.Field.Type.Elem()
+		if elemType.Kind() == reflect.Struct {
+			params.Property.Properties = nil
+			params.Property.Type = &openapi3.Types{"array"}
```
dylanhitt, line 352:
```suggestion
			params.Property.Type = openapi3.NewArraySchema()
```

dylanhitt: Will I think more like 

```golang
params.Property = openapi3.NewArraySchema()
```

Something or the other.

openapi_metadata_test.go
```diff
@@ -0,0 +1,349 @@
...
+	mp := &MetadataParsers{
+		registeredParsers: []MetadataParserEntry{
+			{Name: "expectedParser1", Parser: func(params MetadataParserParams) { return }},
+			{Name: "expectedParser2", Parser: func(params MetadataParserParams) { return }},
+		},
+	}
+
+	parsers := mp.GetRegisteredParsers()
+
+	if len(parsers) != 2 {
```
dylanhitt, line 124: Would you mind using `github.com/stretchr/testify/assert` in most of these cases.

openapi_metadata.go
```diff
@@ -0,0 +1,423 @@
...
+	mp.registeredParsers = append(mp.registeredParsers, entry)
+}
+
+func (mp *MetadataParsers) insertRelative(entry MetadataParserEntry, position string, relativeTo string) error {
+	index := mp.findParserIndex(relativeTo)
+	if index == -1 {
+		return fmt.Errorf("Relative parser '%s' not found", relativeTo)
+	}
+	offset := 0
+	if position == "after" {
```
dylanhitt, line 165: may be nice to move this into two separate funcs `insertAfter` and `insertBefore`. That both just call this func. That way offset is just hardcoded.

But feel free to resolve this more just thinking out loud.

openapi_metadata.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+
+	if params.Field.Type.Kind() == reflect.Struct {
+		if params.Property.Properties == nil {
+			params.Property.Properties = make(map[string]*openapi3.SchemaRef)
+		}
+		processXMLStructChildren(params)
+	} else if params.Field.Type.Kind() == reflect.Slice {
+		elemType := params.Field.Type.Elem()
+		if elemType.Kind() == reflect.Struct {
```
dylanhitt, line 350: This check feels weird. Maybe take a look at `dive()`. You're doing a bit more here than what `dive` is doing though. 

Feel free to resolve this

openapi_metadata.go
```diff
@@ -0,0 +1,423 @@
...
+		mp.registeredNames[entry.Name] = true
+		mp.registeredParsers = append(mp.registeredParsers, entry)
+	}
+}
+
+// RegisterMetadataParser registers a new metadata parser with the given name, parser
+// function, and position string. The position string can be one of the following:
+// "prepend", "append", "before", or "after". If the position is "before" or "after", you
+// must also provide the name of the parser that you want to place the new parser
+// before or after. If the relative parser is not found, an error is returned. If the
```
dylanhitt, line 108: If the parser is already registered I'd prefer the last wins mentality. I.e if a parser is already set we should override it.

dylanhitt (changes requested):
This is a lot. But it looks good and it's written really well. 

I'll probably make a few passes over it.

openapi_metadata.go
DimShadoWWW: I’ve already made the change, and it’s now less error-prone.

openapi_metadata.go
DimShadoWWW: "If parser is already registered, remove it and insert it in the new position."

Added a function to remove registered parsers.

openapi_metadata.go
DimShadoWWW: The second one is the correct implementation. I’ve just pushed a commit with the proper code.

openapi_metadata_test.go
DimShadoWWW: Updated test cases in openapi_metadata_test.go, still working on fixing the mentioned bug.  

- The identified bug is still under investigation and will be addressed in upcoming commits.

openapi_metadata_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+	mp := NewMetadataParsers()
+	mp.InitializeMetadataParsers([]MetadataParserEntry{})
+
+	err := mp.RegisterMetadataParser(parserName, func(params MetadataParserParams) {
+		return
+	}, "invalid", "")
+
+	require.Error(t, err)
+	assert.Equal(t, expectedErr, err.Error())
```
ccoVeille, line 254: This one is simply


```suggestion
	require.EqualError(t, err, expectedErr)
```

openapi_metadata.go
DimShadoWWW: I have a struct like this one:
```
type Fixture struct {
	XMLName xml.Name `xml:"Fixture" json:"-"`

	// List of levels in the competition.
	Level []struct {
		Round []struct {
			DateBegin string `xml:"dateBegin,omitempty,attr" json:"dateBegin,omitempty" validate:"omitempty,datePattern"`
			// Date of the end of the competition.
			DateEnd string `xml:"dateEnd,omitempty,attr" json:"dateEnd,omitempty" validate:"omitempty,datePattern"`

			// Order of the round.
			Order     int    `json:"order,omitempty" xml:"order,attr"`
			RoundName string `xml:"roundName,attr" json:"roundName,omitempty" validate:"required" description:"Round name" example:"Gameweek 17"`
			RoundId   int    `xml:"roundId,attr" json:"roundId,omitempty" validate:"required,min=1" description:"Round ID in our system" example:"4"`
		} `json:"round,omitempty" xml:"Round"`
		// Level name.
		LevelName string `json:"levelName,omitempty" xml:"levelName1,attr" validate:"required" description:"Name of the phase of tournament" example:"Regular Season"`
		// Level ID.
		LevelNumber int `json:"levelNumber,omitempty" xml:"levelNumber,attr" validate:"required,min=1" description:"ID of the phase of tournament" example:"1"`
	} `json:"level,omitempty" xml:"Level"`
	// Competition identity name.
	Channel string `xml:"channel,attr" json:"channel,omitempty" validate:"required" description:"Name of channel corresponding to tournament." example:"argentinec"`
	// Competition name.
	Competition string `xml:"competition,attr" json:"competition,omitempty" validate:"required" description:"Season name" example:"Argentina - Campeonato Primera C Temporada 2024"`
}
```

which is generating this OpenAPI schema:
```
"Fixture": {
    "description": "Fixture schema",
    "properties": {
      "channel": {
        "description": "Name of channel corresponding to tournament.",
        "example": "argentinec",
        "nullable": true,
        "type": "string",
        "xml": {
          "attribute": true,
          "name": "channel"
        }
      },
      "competition": {
        "description": "Season name",
        "example": "Argentina - Campeonato Primera C Temporada 2024",
        "nullable": true,
        "type": "string",
        "xml": {
          "attribute": true,
          "name": "competition"
        }
      },
      "level": {
        "items": {
          "properties": {
            "levelName": {
              "type": "string"
            },
            "levelNumber": {
              "type": "integer"
            },
            "round": {
              "items": {
                "properties": {
                  "dateBegin": {
                    "type": "string"
                  },
                  "dateEnd": {
                    "type": "string"
                  },
                  "order": {
                    "type": "integer"
                  },
                  "roundId": {
                    "type": "integer"
                  },
                  "roundName": {
                    "type": "string"
                  }
                },
                "type": "object"
              },
              "type": "array"
            }
          },
          "type": "object"
        },
        "nullable": true,
        "type": "array",
        "xml": {
          "name": "Level"
        }
      }
    },
    "required": [
      "channel",
      "competition"
    ],
    "type": "object"
}
```

As you can see, from "level" to its child elements, there is no additional information—only "type" is present. This is the issue that persists.  

I added logs to `ParseStructTags` to trace how it processes the attribute "LevelName." However, even though it is included in the schema, it is not found:  

```
func (mp *MetadataParsers) ParseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
	if t.Kind() == reflect.Ptr {
		t = t.Elem()
	}
	if t.Kind() != reflect.Struct {
		return
	}

	for i := 0; i < t.NumField(); i++ {
		field := t.Field(i)
		if field.Name == "levelName" || field.Name == "LevelName" {
			slog.Error("======= levelName field detected", "field", field.Name)
		}

		if field.Anonymous {
....

At this point, I’m out of ideas on how to proceed. In your experience, what steps should I take to identify and fix this bug?

**Review by ccoVeille — APPROVED** [review]

**Comment by DimShadoWWW** [comment]
closed PR on error

**Comment by DimShadoWWW** [comment]
closed PR on error

**Comment by DimShadoWWW** [comment]
restored commit history

File: openapi_metadata.go [hunk 17]
```diff
@@ -0,0 +1,449 @@
...
+	defer mp.parserLock.Unlock()
+
+	mp.registeredParsers = []MetadataParserEntry{}
+	mp.registeredNames = make(map[string]bool)
+}
+
+// InitializeMetadataParsers initializes the metadata parsers with the given customParsers.
+// It will append the customParsers to the existing DefaultParsers. If a parser with the
+// same name is already registered, it will be skipped.
+func (mp *MetadataParsers) InitializeMetadataParsers(customParsers []MetadataParserEntry) {
```
ccoVeille, line 90: I have a problem here

It's like the structure was named FooBar, and we add a InitFooBar method to FooBar

Something like this maybe

```suggestion
func (mp *MetadataParsers) Initialize(customParsers []MetadataParserEntry) {
```

Or


```suggestion
func (mp *MetadataParsers) InitializeParserEntries(customParsers []MetadataParserEntry) {
```

Or


```suggestion
func (mp *MetadataParsers) InitializeEntries(customParsers []MetadataParserEntry) {
```

openapi_metadata.go
```diff
@@ -0,0 +1,449 @@
...
+// registeredParsers slice and registeredNames map by using a mutex lock.
+func (mp *MetadataParsers) Reset() {
+	mp.parserLock.Lock()
+	defer mp.parserLock.Unlock()
+
+	mp.registeredParsers = []MetadataParserEntry{}
+	mp.registeredNames = make(map[string]bool)
+}
+
+// InitializeMetadataParsers initializes the metadata parsers with the given customParsers.
```
ccoVeille, line 87: Please update this accordingly

```suggestion
// Initialize initializes the metadata parsers with the given customParsers.
```

openapi_metadata.go
dylanhitt: I'll take a look tomorrow morning as I should have some spare time.

ccoVeille (approved):

openapi_metadata.go
dylanhitt: Sorry won't be til later this week. Day job getting in the way.

DimShadoWWW: Working with the original code where `parseStructTags` was used directly and the function included the logic for JSON, XML, example, and validation, I managed to make it work.  

this is the code (I added `enum_values` too, to specify all possible values for field):
```
// parseStructTags parses struct tags and modifies the schema accordingly.
// t must be a struct type.
// It adds the following struct tags (tag => OpenAPI schema field):
// - description => description
// - example => example
// - json => nullable (if contains omitempty)
// - validate:
//   - required => required
//   - min=1 => min=1 (for integers)
//   - min=1 => minLength=1 (for strings)
//   - max=100 => max=100 (for integers)
//   - max=100 => maxLength=100 (for strings)
func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
	if t.Kind() == reflect.Ptr {
		t = t.Elem()
	}

	if t.Kind() != reflect.Struct {
		return
	}

	for i := range t.NumField() {
		field := t.Field(i)
		if field.Anonymous {
			parseStructTags(field.Type, schemaRef)
			continue
		}

		jsonFieldName := field.Tag.Get("json")
		jsonFieldName = strings.Split(jsonFieldName, ",")[0] // remove omitempty, etc
		if jsonFieldName == "-" {
			continue
		}

		if jsonFieldName == "" {
			jsonFieldName = field.Name
		}

		if slices.Contains([]string{"XMLName"}, jsonFieldName) {
			// Skip XMLName field in struct tags as it is not relevant to OpenAPI
			continue
		}

		if schemaRef.Value.Properties == nil {
			schemaRef.Value.Properties = make(map[string]*openapi3.SchemaRef)
		}

		if _, exists := schemaRef.Value.Properties[jsonFieldName]; !exists {
			schemaRef.Value.Properties[jsonFieldName] = &openapi3.SchemaRef{
				Value: &openapi3.Schema{},
			}
		}

		property := schemaRef.Value.Properties[jsonFieldName]

		propertyCopy := *property
		propertyValue := *propertyCopy.Value

                //  --- This section does the magic. --- 
		if field.Type.Kind() == reflect.Struct {
			propertyValue.Type = &openapi3.Types{openapi3.TypeObject}
			if propertyValue.Properties == nil {
				propertyValue.Properties = make(map[string]*openapi3.SchemaRef)
			}
			parseStructTags(field.Type, &openapi3.SchemaRef{Value: &propertyValue})
		} else if field.Type.Kind() == reflect.Slice && field.Type.Elem().Kind() == reflect.Struct {
			propertyValue.Type = &openapi3.Types{openapi3.TypeArray}
			itemSchema := openapi3.NewSchemaRef("", openapi3.NewObjectSchema())
			parseStructTags(field.Type.Elem(), itemSchema)
			propertyValue.Items = itemSchema
		}
                
		propertyValue.XML = &openapi3.XML{}

		// Xml attributes
		xmlTag, ok := field.Tag.Lookup("xml")
		if ok {
			xmlTagName := strings.Split(xmlTag, ",")[0]
			if xmlTagName == "-" {
				continue
			}

			if field.Type.Kind() == reflect.Struct {
				for i := 0; i < field.Type.NumField(); i++ {
					childField := field.Type.Field(i)
					if childField.Name == "XMLName" && childField.Type == reflect.TypeOf(xml.Name{}) {
						xmlTagName = childField.Name
						break
					}
				}
			}

			if xmlTagName == "" {
				xmlTagName = field.Name
			}

			propertyValue.XML.Name = xmlTagName

			if slices.Contains(strings.Split(xmlTag, ","), "attr") {
				propertyValue.XML.Attribute = true
			}
		}

		// Example
		example, ok := field.Tag.Lookup("example")
		if ok {
			propertyValue.Example = example
			if propertyValue.Type.Is(openapi3.TypeInteger) {
				exNum, err := strconv.Atoi(example)
				if err != nil {
					slog.Warn("Example might be incorrect (should be integer)", "error", err)
				}
				propertyValue.Example = exNum
			}
		}

		// Emun values
		emunValues, ok := field.Tag.Lookup("enum_values")
		if ok {
			values := strings.Split(emunValues, ",")
			propertyValue.Enum = make([]any, len(values))
			for i, v := range values {
				propertyValue.Enum[i] = v
			}
		}

		// Validation
		validateTag, ok := field.Tag.Lookup("validate")
		validateTags := strings.Split(validateTag, ",")
		if ok && slices.Contains(validateTags, "required") {
			schemaRef.Value.Required = append(schemaRef.Value.Required, jsonFieldName)
		}
		for _, validateTag := range validateTags {
			if strings.HasPrefix(validateTag, "min=") {
				min, err := strconv.Atoi(strings.Split(validateTag, "=")[1])
				if err != nil {
					slog.Warn("Min might be incorrect (should be integer)", "error", err)
				}

				if propertyValue.Type.Is(openapi3.TypeInteger) {
					minPtr := float64(min)
					propertyValue.Min = &minPtr
				} else if propertyValue.Type.Is(openapi3.TypeString) {
					//nolint:gosec // disable G115
					propertyValue.MinLength = uint64(min)
				}
			}
			if strings.HasPrefix(validateTag, "max=") {
				max, err := strconv.Atoi(strings.Split(validateTag, "=")[1])
				if err != nil {
					slog.Warn("Max might be incorrect (should be integer)", "error", err)
				}
				if propertyValue.Type.Is(openapi3.TypeInteger) {
					maxPtr := float64(max)
					propertyValue.Max = &maxPtr
				} else if propertyValue.Type.Is(openapi3.TypeString) {
					//nolint:gosec // disable G115
					maxPtr := uint64(max)
					propertyValue.MaxLength = &maxPtr
				}
			}
		}

		// Description
		description, ok := field.Tag.Lookup("description")
		if ok {
			propertyValue.Description = description
		}
		jsonTag, ok := field.Tag.Lookup("json")
		if ok {
			if strings.Contains(jsonTag, ",omitempty") {
				propertyValue.Nullable = true
			}
		}

		propertyCopy.Value = &propertyValue

		schemaRef.Value.Properties[jsonFieldName] = &propertyCopy

	}
}
```

I will try to apply it to this PR next week, if it’s not done before.

EwenQuim:
Looking at it today and tomorrow, PR is pretty big but of good quality for the moment ;)

openapi_metadata.go
```diff
@@ -0,0 +1,449 @@
...
+		return fmt.Errorf("Relative parser '%s' not found", relativeTo)
+	}
+
+	mp.parserLock.Lock()
+	defer mp.parserLock.Unlock()
+
+	for _, entry := range mp.registeredParsers {
+		if entry.Name == name {
+			// Remove the existing parser before re-registering
+			mp.remove(entry)
```
EwenQuim, line 132: Error is not checked.
Please run `make` to check the CI

openapi_metadata.go
```diff
@@ -0,0 +1,449 @@
...
+
+type MetadataParserEntry struct {
+	Name   string
+	Parser MetadataParserFunction
+}
+
+type MetadataParsers struct {
+	registeredParsers []MetadataParserEntry
+	parserLock        sync.Mutex
+	registeredNames   map[string]bool
```
EwenQuim, line 55: From what I've read, `registeredNames` is a Set that just checks that parsers aren't registered twice. Since you just iterate on `registeredParsers`, can you please simply use a map? `registeredParsers map[string]MetadataParserEntry`.

openapi_metadata.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+type MetadataParserFunction func(params MetadataParserParams)
+
+type MetadataParserEntry struct {
+	Name   string
+	Parser MetadataParserFunction
+}
+
+type MetadataParsers struct {
```
EwenQuim, line 52: Please comment the type so one can easily know what is this struct about.

openapi.go
```diff
@@ -16,10 +16,15 @@ import (
 
 func NewOpenAPI() *OpenAPI {
 	desc := NewOpenApiSpec()
+
+	mp := NewMetadataParsers()
+	mp.Initialize(DefaultParsers)
```
EwenQuim, line 21: Since you use `NewMetadataParsers` AND `Initialize` function, why don't you write something like `mp := NewMetadataParsers(DefaultParsers)` directly? Since it makes sense in a business point of view, I think it's a correct API.

openapi_metadata_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+	parsers := mp.GetRegisteredParsers()
+
+	require.NotNil(t, parsers)
+	require.Len(t, parsers, 2)
+
+	if !reflect.DeepEqual(parsers, mp.registeredParsers) {
+		t.Errorf("Returned parsers do not match the original registered parsers")
+	}
+}
```
EwenQuim, line 134: Since `mp.GetRegisteredParsers()` is returning the `mp.registeredParsers`, this test is useless. What do you wanted to test exactly?

openapi_metadata_test.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	mp.Reset()
+
+	require.NotNil(t, mp.registeredParsers)
+	require.NotNil(t, mp.registeredNames)
+
+	require.Empty(t, mp.registeredParsers)
+	require.Empty(t, mp.registeredNames)
+}
```
EwenQuim, line 156: Good idea!

openapi_metadata_test.go
EwenQuim, line 1: It's really good to have a such a big file with so many tests, thank you for that!

I'm a little disappointed that it does not test custom "Metadata parsers"

EwenQuim:
It seems that this PR does several things. If I understand correctly:

- ✅ it refactors way more cleanly the "metadata parsing"
- ✅ it allows users to implement their custom metadata parsing
- ✅ it implements better XML structure parsing (support for `xml.Name`) 

So this PR is really cool, I like the idea and it really improves Fuego, **thank you** for all these efforts! 🥳 

But by doing 3 things at a time, it makes the PR really big (more than 800 lines!!!) and that's why it is hard to review and to understand. I also feel that it might be quite heavy for you (this PR is a month old), making changes is not easy and it begins to be quite slow.

Because of that, I will skip it for the next release (v0.18.0). I encourage you to iterate quickly with small changes rather than making big PRs, so we will be able to add it in the next release (v0.19.0). If you can't, no problem, we're all volunteers, I understand you might not have time :)

PR #333, opened 2025-01-07
BREAKING: add WithRequestContentType to Engine from Server

dylanhitt:


engine.go
```diff
@@ -59,6 +59,12 @@ var defaultOpenAPIConfig = OpenAPIConfig{
 	JSONFilePath: "doc/openapi.json",
 }
 
+// WithRequestContentType sets the accepted content types for the engine.
+// By default, the accepted content types is */*.
+func WithRequestContentType(consumes ...string) func(*Engine) {
```
ccoVeille, line 64: What about naming it


```suggestion
func WithRequestAcceptedContentType(consumes ...string) func(*Engine) {
```


```suggestion
func WithAcceptedContentType(consumes ...string) func(*Engine) {
```

It could also be a plural

```suggestion
func WithAcceptedContentTypes(consumes ...string) func(*Engine) {
```

Here it's unclear. It seems it set the HTTP request Content

You are moving it in a breaking change, renaming is then possible.

Also why is the arguments being called `consumes` it's strange to me

dylanhitt: `consumes` is a vernacular used by `kin-openapi`

```golang
func NewContentWithSchemaRef(schema *SchemaRef, consumes []string) Content {
	if len(consumes) == 0 {
		return Content{
			"*/*": NewMediaType().WithSchemaRef(schema),
		}
	}
	content := make(map[string]*MediaType, len(consumes))
	for _, mediaType := range consumes {
		content[mediaType] = NewMediaType().WithSchemaRef(schema)
	}
	return content
}
```

https://github.com/getkin/kin-openapi/blob/v0.128.0/openapi3/content.go#L29

dylanhitt: I don't really want to change the name either had we have a Route Option with a similar name. 

```
// RequestContentType sets the accepted content types for the route.
// By default, the accepted content types is */*.
// This will override any options set at the server level.
func OptionRequestContentType(consumes ...string) func(*BaseRoute) {
	return func(r *BaseRoute) {
		r.AcceptedContentTypes = consumes
	}
}
```

dylanhitt: If the name was to be changed. It should all be addressed in a the same PR changing the name across the board. Not simply because this function is being moved. The spirit of this change is to consolidate all our server option into an engine option much like my last PRs.

ccoVeille (approved):

dylanhitt:
Actually. This broken as in this doesn't do anything more. I'll open some PRs targeting this with possible solutions.

dylanhitt:
this commit shouldn't pass - https://github.com/go-fuego/fuego/pull/333/commits/ee5a72fcf3b34b680c6c37daae88e91e2cf8c84f

The test here is bad since content.Get will make a best effort at returning the schema for any proper content type

dylanhitt:
Closing in favor of #334

PR #334, opened 2025-01-07
fix: NewRoute/NewBaseRoute to take engine not just OpeAPI for accepted content types

dylanhitt:
Relates to #333 

Somewhere along implementing support for different engines this was lost. This is a bit disruptive as it's changing the `NewRoute/NewBaseRoute`. The other option would be to place `acceptedContentTypes` on our `OpenAPI` struct I could go either way. We wouldn't be changing the `NewRoute` and `NewBaseRoute` API in the case of the later, but again we'd need to support yet another nested structure with it's own options. I think flattening in the Engine is OK, I guess 🤷.

So the options are

1. This PR. Just be fine with configuration occurring on the engine struct
2. Place `acceptedContentTypes` in the `OpenAPI` struct. We can could then do something like

```golang
func WithRequestContentType(consumes ...string) func(*Engine) {
	return func(e *Engine) { e.OpenAPI.acceptedContentTypes = consumes }
}
```

3. Place `acceptedContentTypes` in the `OpenAPI` struct. And expose a `WithOpenAPIOptions`. At this point we're nesting pretty deep...
4. This could exist on the `OpenAPIConfig` struct as well. Then engine would still need to be passed to the `New*Route` funcs. 

Cheers

### DIFF
--- a/route.go
+++ b/route.go
@@ -22,13 +22,13 @@ type Route[ResponseBody any, RequestBody any] struct {
 
 func NewBaseRoute(method, path string, handler any, e *Engine, options ...func(*BaseRoute)) BaseRoute {
 	baseRoute := BaseRoute{
-		Method:               method,
-		Path:                 path,
-		Params:               make(map[string]OpenAPIParam),
-		FullName:             FuncName(handler),
-		Operation:            openapi3.NewOperation(),
-		OpenAPI:              e.OpenAPI,
-		AcceptedContentTypes: e.acceptedContentTypes,
+		Method:              method,
+		Path:                path,
+		Params:              make(map[string]OpenAPIParam),
+		FullName:            FuncName(handler),
+		Operation:           openapi3.NewOperation(),
+		OpenAPI:             e.OpenAPI,
+		RequestContentTypes: e.requestContentTypes,
 	}
 
 	for _, o := range options {
@@ -62,7 +62,7 @@ type BaseRoute struct {
 	FullName string
 
 	// Content types accepted for the request body. If nil, all content types (*/*) are accepted.
-	AcceptedContentTypes []string
+	RequestContentTypes []string
 
 	Middlewares []func(http.Handler) http.Handler
 
--- a/route.go
+++ b/route.go
@@ -7,9 +7,9 @@ import (
 	"github.com/getkin/kin-openapi/openapi3"
 )
 
-func NewRoute[T, B any](method, path string, handler any, openapi *OpenAPI, options ...func(*BaseRoute)) Route[T, B] {
+func NewRoute[T, B any](method, path string, handler any, e *Engine, options ...func(*BaseRoute)) Route[T, B] {
 	return Route[T, B]{
-		BaseRoute: NewBaseRoute(method, path, handler, openapi, options...),
+		BaseRoute: NewBaseRoute(method, path, handler, e, options...),
 	}
 }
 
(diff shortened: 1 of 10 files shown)


ccoVeille (approved):

server.go
```diff
@@ -336,12 +336,6 @@ func WithLogHandler(handler slog.Handler) func(*Server) {
 	}
 }
 
-// WithRequestContentType sets the accepted content types for the server.
-// By default, the accepted content types is */*.
-func WithRequestContentType(consumes ...string) func(*Server) {
-	return func(s *Server) { s.acceptedContentTypes = consumes }
-}
-
```
EwenQuim, line 344: Why is it deleted?

server_test.go
```diff
@@ -320,32 +320,6 @@ func dummyController(_ ContextWithBody[ReqBody]) (Resp, error) {
...
-		content := route.Operation.RequestBody.Value.Content
-		require.NotNil(t, content.Get("application/json"))
-		require.NotNil(t, content.Get("application/xml"))
-		require.Equal(t, "#/components/schemas/ReqBody", content.Get("application/json").Schema.Ref)
-		require.Equal(t, "#/components/schemas/ReqBody", content.Get("application/xml").Schema.Ref)
-		_, ok := s.OpenAPI.Description().Components.RequestBodies["ReqBody"]
-		require.False(t, ok)
-	})
-}
-
```
EwenQuim, line 348: Why is it deleted?

route.go
```diff
@@ -20,14 +20,15 @@ type Route[ResponseBody any, RequestBody any] struct {
 	BaseRoute
 }
 
-func NewBaseRoute(method, path string, handler any, openapi *OpenAPI, options ...func(*BaseRoute)) BaseRoute {
+func NewBaseRoute(method, path string, handler any, e *Engine, options ...func(*BaseRoute)) BaseRoute {
```
EwenQuim, line 23: But do we really need this? For future compatibility maybe?

server_test.go
EwenQuim: My bad, moved to `engine_test.go`

server.go
EwenQuim: My bad, moved to an engine option!

dylanhitt: yeah, I thought about keeping it, but since it's very broken I didn't really see the point of deprecating it.

route.go
dylanhitt: I figure we can just deprecate this function. You are just talking about the function right? Deprecating the Struct of BaseRoute would be a fairly large change which we'd probably want a separate PR for.

EwenQuim: No, I was talking about changing the *OpenAPI param to *engine

dylanhitt: Oh. It fixed the WithContentTypes option. I think the line you're missing is:

https://github.com/go-fuego/fuego/pull/334/files#diff-6b1e829598248403d0c671996c16c10f8a2d8db33f96b6db158654b78103dec4R31

see the `AcceptedContentTypes: e.acceptedContentTypes,`

dylanhitt: Hence my 4 questions where do we want to put `acceptedContentTypes` and how do we want to provide an option to configure it 😄

dylanhitt: Let me know if that clarifies things. I personally don't mind `acceptedContentTypes` existing on the Engine. At the very least I want avoid having to support a set of options for NewOpenAPI as we're getting to be pretty nested at that point. 

So my preference is: 1 or 2

dylanhitt: Hey @EwenQuim any more word on this?

dylanhitt:
I'm going to go ahead and merge this unless anyone objects.

EwenQuim:
Looking at it again right now

dylanhitt:
> Perfect. Sorry for the long review delay!

Thank you. Sorry been a bit impatient. Been thinking about #341 a bunch and i know this would cause conflicts.

Issue #322, opened 2024-12-29
The custom error addition should handle edge cases


ccoVeille:
Please add a test about a custom error handler that could be used to invalidate an error by returning nil

adding a custom error handler allow to mess up things, because people may do that in their custom handler if you dont want to support it, the mechanism of error handling should be rewritten to forbid this behavior

_Originally posted by @ccoVeille in https://github.com/go-fuego/fuego/pull/320#discussion_r1899168562_

EwenQuim:
Good ideas

Elite-tch:
I'd love to give this a go.

ekumamatthew:
May I be assigned to this?

olisaagbafor:
Can I take this issue?

dylanhitt:
Anyone is welcome to take this as it's pretty wide open. I guess to be to close this we would need to test all the content-type options inputtabl for this scenario. Right now we don't test `application/json`. 

The best thing would be to add a table of content types here and test then. https://github.com/go-fuego/fuego/blob/main/serve_test.go#L637

Hope this helps!

ryzen-xp:
https://app.onlydust.com/p/timelycapsule

ryzen-xp:
Could I try solving this?

dylanhitt:
Hey @ryzen-xp feel free.

Some work had been started on this. Please see this comment for closable acceptance criteria. https://github.com/go-fuego/fuego/issues/322#issuecomment-2587509016

chiscookeke11:
Could I take over this issue?

0xprinc:
Could I grab this task?

Supa-mega:
Is it okay if I tackle this?

KinmatTech:
Is it okay if I tackle this?

martinvibes:
Holla i'm Martin, I'm a frontend developer who enjoys building smooth applications to pixel perfect and engaging websites using JavaScript, typescript and frameworks(React, nextjs, tailwindcss etc.) and i have contributed to other projects that have been merged. 
Am also experienced in working and writing smart contracts.
Kindly assign :)

Michaelkingsdev:
May I be assigned to this?

dylanhitt:
As stated before anyone is welcome to take this issue. Feel free to take the initiative!

😄

ShantelPeters:
May I try my hand at this?

FrankiePower:
Hello Team,

[...]

PR #336, opened 2025-01-08
chore: test Flow for edge cases cause by misconfigs

dylanhitt:
Per #322

The following is to test provide a bit easier of test bed for testing edge cases do to server misconfigs or poor user input. The test here will pass, but they bring up some questions:

1. When ErrorHandler is overridden there is a possibility it can return `nil` causing golang encode to eventually land on `null` (I will probably expand this to more return types but wanted to get the conversation going). What should we do here? Is the user setting the return type to nil to say it's fine and we should really respond with like `NoContent` or is just saying 500 with empty/null response good enough. I personally think if the user is setting their error to nil they're doing something on purpose 🤷 
2. In the case of transformOut error when it's receiver is a value and not a pointer. We throw errors.New(""). ErrorStrings are not marshalable. In this case I think we should wrap this error with `fuego.InternalServerError` since transform is supported only on structs with the implementation we can assume a json/xml/x-yaml response is fine depending on the user request.

### DIFF
--- a/serialization.go
+++ b/serialization.go
@@ -58,7 +58,11 @@ func transformOut[T any](ctx context.Context, ans T) (T, error) {
 
 	_, ok := any(ans).(OutTransformer)
 	if ok {
-		return ans, errors.New("OutTransformer must be implemented by a POINTER RECEIVER. Please read the [OutTransformer] documentation")
+		err := errors.New("OutTransformer must be implemented by a POINTER RECEIVER. Please read the [OutTransformer] documentation")
+		slog.Warn(err.Error())
+		return ans, InternalServerError{
+			Err: err,
+		}
 	}
 
 	outTransformer, ok := any(&ans).(OutTransformer)
--- a/serialization.go
+++ b/serialization.go
@@ -148,7 +148,7 @@ func SendYAMLError(w http.ResponseWriter, _ *http.Request, err error) {
 	}
 
 	w.WriteHeader(status)
-	_ = SendYAML(w, nil, err.Error())
+	_ = SendYAML(w, nil, err)
 }
 
 // SendJSON sends a JSON response.
--- a/serve_test.go
+++ b/serve_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -44,10 +45,26 @@ func (t *testOutTransformer) OutTransform(ctx context.Context) error {
 	return nil
 }
 
+type testOutTransformerOnNotReceiver struct {
+	Name     string `json:"name"`
+	Password string `json:"ans"`
+}
+
+func (t testOutTransformerOnNotReceiver) OutTransform(ctx context.Context) error {
+	t.Name = "M. " + t.Name
+	t.Password = "redacted"
+	return nil
+}
+
 var _ OutTransformer = &testOutTransformer{}
+var _ OutTransformer = &testOutTransformerOnNotReceiver{}
 
 func testControllerWithOutTransformer(c ContextNoBody) (testOutTransformer, error) {
 	return testOutTransformer{Name: "John"}, nil
+}
+
+func testControllerWithOutTransformerOnValueReceiver(c ContextNoBody) (testOutTransformerOnNotReceiver, error) {
+	return testOutTransformerOnNotReceiver{Name: "John"}, nil
 }
 
 func testControllerWithOutTransformerStar(c ContextNoBody) (*testOutTransformer, error) {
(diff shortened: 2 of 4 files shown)


EwenQuim:
Good idea. I'd say it's user's will...

> 1. When ErrorHandler is overridden there is a possibility it can return nil causing golang encode to eventually land on null (I will probably expand this to more return types but wanted to get the conversation going). What should we do here? Is the user setting the return type to nil to say it's fine and we should really respond with like NoContent or is just saying 500 with empty/null response good enough. I personally think if the user is setting their error to nil they're doing something on purpose 🤷

Yeah clearly that's user defined on purpose so he knows what he's doing. Thanks for the test, it'll help not break future changes.

> 2. In the case of transformOut error when it's receiver is a value and not a pointer. We throw errors.New(""). ErrorStrings are not marshalable. In this case I think we should wrap this error with fuego.InternalServerError since transform is supported only on structs with the implementation we can assume a json/xml/x-yaml response is fine depending on the user request.

> when it's receiver is a value and not a pointer

It's unfortunate but that's how Go works, when using a value receiver, the struct will not be changed! So Fuego must not do magic here. User expects Go-like behavior. But maybe we can add a warning!

> We throw errors.New(""). ErrorStrings are not marshalable.

Mmmh I see. But same, user does not expect a surprise `fuego.InternalServerError` to appear. I'd say it's better to return a string containing the error?

serve_test.go
```diff
@@ -561,3 +578,39 @@ func newTLSTestHelper() (*tlsTestHelper, error) {
...
+	t.Run("base", func(t *testing.T) {
+		e := NewEngine()
+		w := httptest.NewRecorder()
+		ctx := newTestCtx(w)
+		Flow(e, ctx, testController)
+		assert.Equal(t, http.StatusOK, w.Code)
+		assert.Equal(t, crlf(`{"ans":"Hello World"}`), w.Body.String())
+	})
+	t.Run("with nil return in ErrorHandler", func(t *testing.T) {
+		e := NewEngine(WithErrorHandler(func(err error) error { return nil }))
```
ccoVeille, line 601: What about a test with


```suggestion
		e := NewEngine(WithErrorHandler(nil)
```

EwenQuim: It'll throw, it's already tested elsewhere ;)

dylanhitt:
> Mmmh I see. But same, user does not expect a surprise fuego.InternalServerError to appear. I'd say it's better to return a string containing the error?

Let me see how that would look.

dylanhitt:
> Mmmh I see. But same, user does not expect a surprise fuego.InternalServerError to appear. I'd say it's better to return a string containing the error?

Ehhh there doesn't really seem to be a nice way to do this. Unless we create some internal fuego error type or something, but even then it's not great. Will keep trying...

dylanhitt:
We could override the marshaler to handle this case I guess.

EwenQuim:
MMmh I don't want Fuego to become a "magic" framework!

dylanhitt:
Hence why I personally like fuego.InternalServerError. In reality that's what is happening. Something on the server is erroring the administrator needs to handle it. In the case of text/plain it will just return the `.Error()` of our internal server error.

dylanhitt:
This also just uncovered that we've been doing this on `x-yaml` 😅 

```golang
// SendYAMLError sends a YAML error response.
// If the error implements ErrorWithStatus, the status code will be set.
func SendYAMLError(w http.ResponseWriter, _ *http.Request, err error) {
	status := http.StatusInternalServerError
	var errorStatus ErrorWithStatus
	if errors.As(err, &errorStatus) {
		status = errorStatus.StatusCode()
	}

	w.WriteHeader(status)
	_ = SendYAML(w, nil, err.Error())
}
```

the `err.Error()`...

serve_test.go
```diff
@@ -561,3 +579,70 @@ func newTLSTestHelper() (*tlsTestHelper, error) {
...
+				accept:           "application/xml",
+				expectedResponse: "<HTTPError><title>Internal Server Error</title><status>500</status></HTTPError>",
+			},
+			{
+				accept:           "text/html",
+				expectedResponse: "500 Internal Server Error",
+			},
+			{
+				accept:           "application/x-yaml",
+				expectedResponse: "500 Internal Server Error\n",
```
ccoVeille, line 629: This one doesn't use crlf method

dylanhitt: See #339 as for why or the comment just above 🙃

dylanhitt: ah i guess it doesn't affect it. updating

serialization.go
```diff
@@ -58,7 +58,11 @@ func transformOut[T any](ctx context.Context, ans T) (T, error) {
 
 	_, ok := any(ans).(OutTransformer)
 	if ok {
-		return ans, errors.New("OutTransformer must be implemented by a POINTER RECEIVER. Please read the [OutTransformer] documentation")
+		err := errors.New("OutTransformer must be implemented by a POINTER RECEIVER. Please read the [OutTransformer] documentation")
```
ccoVeille, line 61: I'm unsure to understand the formatting of OutTransformer here

I mean it's an error message, not a godoc or a Markdown file

Am I missing something?

ccoVeille (approved):

serialization.go
dylanhitt: Yeah the only place this should really manifest is in a log server side. Maybe a good time to rethink it's shape.

dylanhitt:
Hey @EwenQuim are you call with this merging?

You had some concerns about switching the error to to our `InternalServerError`. I personally think the return on this is the best we can do. I think it's good to not expose the client to any information.

EwenQuim (approved):
Good ideas

EwenQuim:
I let you merge, thanks for this PR! A good cleanup

PR #338, opened 2025-01-08
Fixes generic return type support

### LINKED ISSUE #332: Generic type OpenAPI generation seems not supported
**To Reproduce**

1. Make a struct that has generic type like this:
```go
type BareSuccessResponse[Res any] struct {
	StatusCode int    `json:"statusCode"`
	Result     Res    `json:"result"`
	Message    string `json:"message"`
}
```
2. Use the struct as return type like `BareSuccessResponse[ExpectedResultStruct]`
3. Open swagger

**Expected behavior**
It should generate the proper extended types for the response

**Screenshots**
image

Fuego version before posting):** 0.17.0

**Go version (please check if it happens with the last Go version before posting):** 1.23.4

**Additional context**
My current workaround is explicitly specify the response type using `OptionAddResponse`. But that actually removes the `BareSuccessResponse` structure.

EwenQuim:
Fixes #332

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -315,6 +315,7 @@ func dive(openapi *OpenAPI, t reflect.Type, tag SchemaTag, maxDepth int) SchemaT
 		if t.Kind() == reflect.Struct && strings.HasPrefix(tag.Name, "DataOrTemplate") {
 			return dive(openapi, t.Field(0).Type, tag, maxDepth-1)
 		}
+		tag.Name = transformTypeName(tag.Name)
 		tag.Ref = "#/components/schemas/" + tag.Name
 		tag.Value = openapi.getOrCreateSchema(tag.Name, reflect.New(t).Interface())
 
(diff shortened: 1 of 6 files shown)


openapi.go
```diff
@@ -315,6 +315,7 @@ func dive(openapi *OpenAPI, t reflect.Type, tag SchemaTag, maxDepth int) SchemaT
 		if t.Kind() == reflect.Struct && strings.HasPrefix(tag.Name, "DataOrTemplate") {
```
dylanhitt, line 315:
```suggestion
		if t.Kind() == reflect.Struct && strings.HasPrefix(tag.Name, "DataOrTemplate") {
```

openapi.go
```diff
@@ -315,6 +315,7 @@ func dive(openapi *OpenAPI, t reflect.Type, tag SchemaTag, maxDepth int) SchemaT
...
 			return dive(openapi, t.Field(0).Type, tag, maxDepth-1)
 		}
+		tag.Name = transformTypeName(tag.Name)
```
dylanhitt, line 318: nitpick

You can just do this on line 314 above 

```
tag.Name = transformTypeName(t.Name())
```

dylanhitt (approved):

openapi.go
EwenQuim: Yes exactly, thanks!

PR #340, opened 2025-01-09
fix: SendYAMLError to properly pass the entire error when sending

### LINKED ISSUE #339: SendYAMLError should send the marshaled struct not the error string
Per this comment https://github.com/go-fuego/fuego/pull/336#issuecomment-2578423553

This will more than likely require adding yaml tags to HTTPError as well.

dylanhitt:
closes #339

Based on #336 for readability.


EwenQuim (approved):

ccoVeille (approved):
It's indeed safer and error .edsage are not exposed unless when it's on purpose

PR #343, opened 2025-01-10
feat: allow multiple global middlewares

### LINKED ISSUE #284: Create a `WithGlobalMiddlewares` server option to replace `WithCorsMiddleware`
`WithCorsMiddleware` is very specific. We need Global Middlewares.

They differ from route middleware by the fact that every request pass by them, EVEN IF the request is not being routed to a handler. For example, Global Middlewares will handle `/not-existing-path`, while route middlewares will not handles them.

- `Use` registers middlewares to routes, not globally.
- `option.Middleware` registers middlewares to routes, not globally.

Replace `corsMiddleware` by a list of middlewares. Keep the example of CORS in the description.

rizerkrof:
fixes #284

### DIFF
--- a/server.go
+++ b/server.go
@@ -44,7 +44,7 @@ type Server struct {
 	// Not stored with the other middlewares because it is a special case :
 	// it applies on routes that are not registered.
 	// For example, it allows OPTIONS /foo even if it is not declared (only GET /foo is declared).
-	corsMiddleware func(http.Handler) http.Handler
+	globalMiddlewares []func(http.Handler) http.Handler
 
 	*Engine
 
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -177,22 +177,22 @@ func WithTemplateFS(fs fs.FS) func(*Server) {
...
 //	s := fuego.NewServer(
-//		WithCorsMiddleware(cors.New(cors.Options{
+//		WithGlobalMiddleware(cors.New(cors.Options{
 //			AllowedOrigins:   []string{"*"},
 //			AllowedMethods:   []string{"GET", "POST", "PUT", "DELETE", "OPTIONS"},
 //			AllowedHeaders:   []string{"*"},
 //			AllowCredentials: true,
 //		}).Handler)
 //	)
-func WithCorsMiddleware(corsMiddleware func(http.Handler) http.Handler) func(*Server) {
```
dylanhitt, line 194: Mind changing your commit to `BREAKING: ...`. So the changelog is more explicit on the change?

dylanhitt (approved):

EwenQuim:
Closed in favor of #345

PR #344, opened 2025-01-11
WIP: OpenAPIServable

dylanhitt:
Heavy WIP

Working through how this would work. 

Just wanted to get a initial outline on how we could do this. Feel free to discuss over top of this. I imagine we reach the point with our adaptors where we have something like `fuegogin.Adaptor.Run()`, in this case it would `engine.OutputOpenAPISpec`, `engine.RegisterOpenAPIRoutes`, and then start the gin server. Again everything is exposed and we can provide options maybe, but users can plug and play as they please. 

PS. Feel free to open PR against this or contribute directly 😄

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -100,11 +100,11 @@ func WithOpenAPIConfig(config OpenAPIConfig) func(*Engine) {
 		e.OpenAPIConfig.DisableSwaggerUI = config.DisableSwaggerUI
 
 		if !validateSpecURL(e.OpenAPIConfig.SpecURL) {
-			slog.Error("Error serving openapi json spec. Value of 's.OpenAPIServerConfig.SpecURL' option is not valid", "url", e.OpenAPIConfig.SpecURL)
+			slog.Error("Error serving OpenAPI JSON spec. Value of 's.OpenAPIServerConfig.SpecURL' option is not valid", "url", e.OpenAPIConfig.SpecURL)
 			return
 		}
 		if !validateSwaggerURL(e.OpenAPIConfig.SwaggerURL) {
-			slog.Error("Error serving swagger ui. Value of 's.OpenAPIServerConfig.SwaggerURL' option is not valid", "url", e.OpenAPIConfig.SwaggerURL)
+			slog.Error("Error serving Swagger UI. Value of 's.OpenAPIServerConfig.SwaggerURL' option is not valid", "url", e.OpenAPIConfig.SwaggerURL)
 			return
 		}
 	}
--- a/engine.go
+++ b/engine.go
@@ -5,8 +5,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"net/http"
 	"os"
 	"path/filepath"
+
+	"github.com/getkin/kin-openapi/openapi3"
 )
 
 // NewEngine creates a new Engine with the given options.
@@ -53,10 +56,21 @@ type OpenAPIConfig struct {
 	DisableLocalSave bool
 	// Pretty prints the OpenAPI spec with proper JSON indentation
 	PrettyFormatJSON bool
+	// URL to serve the OpenAPI JSON spec
+	SpecURL string
+	// Handler to serve the OpenAPI UI from spec URL
+	UIHandler func(specURL string) http.Handler
+	// URL to serve the swagger UI
+	SwaggerURL string
+	// If true, the server will not serve the Swagger UI
+	DisableSwaggerUI bool
 }
 
 var defaultOpenAPIConfig = OpenAPIConfig{
 	JSONFilePath: "doc/openapi.json",
+	SpecURL:      "/swagger/openapi.json",
+	SwaggerURL:   "/swagger",
+	UIHandler:    DefaultOpenAPIHandler,
 }
 
 // WithRequestContentType sets the accepted content types for the engine.
@@ -70,10 +84,29 @@ func WithOpenAPIConfig(config OpenAPIConfig) func(*Engine) {
 		if config.JSONFilePath != "" {
 			e.OpenAPIConfig.JSONFilePath = config.JSONFilePath
 		}
+		if config.SpecURL != "" {
+			e.OpenAPIConfig.SpecURL = config.SpecURL
+		}
+		if config.SwaggerURL != "" {
+			e.OpenAPIConfig.SwaggerURL = config.SwaggerURL
+		}
+		if config.UIHandler != nil {
+			e.OpenAPIConfig.UIHandler = config.UIHandler
+		}
 
 		e.OpenAPIConfig.Disabled = config.Disabled
 		e.OpenAPIConfig.DisableLocalSave = config.DisableLocalSave
 		e.OpenAPIConfig.PrettyFormatJSON = config.PrettyFormatJSON
+		e.OpenAPIConfig.DisableSwaggerUI = config.DisableSwaggerUI
+
+		if !validateSpecURL(e.OpenAPIConfig.SpecURL) {
+			slog.Error("Error serving openapi json spec. Value of 's.OpenAPIServerConfig.SpecURL' option is not valid", "url", e.OpenAPIConfig.SpecURL)
+			return
+		}
+		if !validateSwaggerURL(e.OpenAPIConfig.SwaggerURL) {
+			slog.Error("Error serving swagger ui. Value of 's.OpenAPIServerConfig.SwaggerURL' option is not valid", "url", e.OpenAPIConfig.SwaggerURL)
+			return
+		}
 	}
 }
 
@@ -95,8 +128,14 @@ func DisableErrorHandler() func(*Engine) {
 	}
 }
 
+func (e *Engine) SpecHandler() func(c ContextNoBody) (openapi3.T, error) {
+	return func(c ContextNoBody) (openapi3.T, error) {
+		return *e.OpenAPI.Description(), nil
+	}
+}
+
 // OutputOpenAPISpec takes the OpenAPI spec and outputs it to a JSON file
-func (e *Engine) OutputOpenAPISpec() []byte {
+func (e *Engine) OutputOpenAPISpec() *openapi3.T {
 	e.OpenAPI.computeTags()
 
 	// Validate
@@ -117,7 +156,7 @@ func (e *Engine) OutputOpenAPISpec() []byte {
 			slog.Error("Error saving spec to local path", "error", err, "path", e.OpenAPIConfig.JSONFilePath)
 		}
 	}
-	return jsonSpec
+	return e.OpenAPI.Description()
 }
 
 func (e *Engine) saveOpenAPIToFile(jsonSpecLocalPath string, jsonSpec []byte) error {
--- a/examples/gin-compat/main.go
+++ b/examples/gin-compat/main.go
@@ -25,17 +25,17 @@ type HelloResponse struct {
 }
 
 func main() {
-	e, _ := server()
+	a := server()
 
 	fmt.Println("OpenAPI at at http://localhost:8980/swagger ✅")
 
-	err := e.Run(":8980")
+	err := a.Run(":8980")
 	if err != nil {
 		panic(err)
 	}
 }
 
-func server() (*gin.Engine, *fuego.OpenAPI) {
+func server() *fuegogin.Adaptor {
 	ginRouter := gin.Default()
 	engine := fuego.NewEngine()
 
@@ -81,10 +81,7 @@ func server() (*gin.Engine, *fuego.OpenAPI) {
 	)
 
 	// Serve the OpenAPI spec
-	ginRouter.GET("/openapi.json", serveOpenApiJSONDescription(engine.OpenAPI))
-	ginRouter.GET("/swagger", DefaultOpenAPIHandler("/openapi.json"))
-
-	return ginRouter, engine.OpenAPI
+	return fuegogin.NewAdaptor(ginRouter, engine)
 }
 
 func (h *HelloRequest) InTransform(ctx context.Context) error {
--- a/examples/gin-compat/main.go
+++ b/examples/gin-compat/main.go
@@ -25,17 +25,17 @@ type HelloResponse struct {
 }
 
 func main() {
-	a := server()
+	e, _ := server()
 
 	fmt.Println("OpenAPI at at http://localhost:8980/swagger ✅")
 
-	err := a.Run(":8980")
+	err := e.Run(":8980")
 	if err != nil {
 		panic(err)
 	}
 }
 
-func server() *fuegogin.Adaptor {
+func server() (*gin.Engine, *fuego.OpenAPI) {
 	ginRouter := gin.Default()
 	engine := fuego.NewEngine()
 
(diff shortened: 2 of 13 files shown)


openapi.go
```diff
@@ -79,6 +79,28 @@ func NewOpenApiSpec() openapi3.T {
 	return spec
 }
 
+type OpenAPIServable interface {
+	SpecHandler()
```
dylanhitt, line 83: Need to think of way to make this contract a bit more binding? 🤔 

Maybe not. The whole point is that people can server these whatever way they want. We're basically just dealing with Engine config for them.

openapi.go
```diff
@@ -79,6 +79,28 @@ func NewOpenApiSpec() openapi3.T {
...
+	UIHandler()
+	URL() string
+}
+
+func RegisterOpenAPIRoutes(e *Engine, o OpenAPIServable) {
```
dylanhitt, line 88: Could be just a method

engine.go
```diff
@@ -53,21 +55,51 @@ type OpenAPIConfig struct {
 	DisableLocalSave bool
 	// Pretty prints the OpenAPI spec with proper JSON indentation
 	PrettyFormatJSON bool
+	// URL to serve the OpenAPI JSON spec
+	SpecURL string
+	// Handler to serve the OpenAPI UI from spec URL
+	UIHandler func(specURL string) func(c ContextNoBody) (HTML, error)
```
dylanhitt, line 61: Changed this to adhere to the fuego controller func. Think it's a nice way to introduce someone to a fuego native route. It also allows us to unify this across adaptors as every adaptor will need a `handlefuego`

openapi.go
dylanhitt: Also the user can either just call this or do we want to eventially wrap our adaptors with a `Run` that calls this for them.

extra/fuegogin/adaptor.go
```diff
@@ -9,6 +9,23 @@ import (
 	"github.com/go-fuego/fuego/internal"
 )
 
+type Adaptor struct {
```
dylanhitt, line 12: This would get a constructor at some point.

dylanhitt: Also.. adaptor doesn’t need to hold fuego Engine as well. Could be nicer looking

dylanhitt: This ends up looking like 

the interface

```golang
type OpenAPIServable interface {
	SpecHandler(e *Engine)
	UIHandler(e *Engine)
	URL() string
}

func (e *Engine) RegisterOpenAPIRoutes(o OpenAPIServable) {
	if e.OpenAPIConfig.Disabled {
		return
	}
	e.OpenAPI.Description().Servers = append(e.OpenAPI.Description().Servers, &openapi3.Server{
		URL:         o.URL(),
		Description: "local server",
	})
	o.SpecHandler(e)

	if e.OpenAPIConfig.DisableSwaggerUI {
		return
	}
	o.UIHandler(e)
}
```

the adaptor

```golang
type Adaptor struct {
	*gin.Engine
}

func (a *Adaptor) SpecHandler(e *fuego.Engine) {
	Get(e, a, e.OpenAPIConfig.SpecURL, e.SpecHandler())
}

func (a *Adaptor) UIHandler(e *fuego.Engine) {
	Get(e, a, e.OpenAPIConfig.SwaggerURL+"/", e.OpenAPIConfig.UIHandler(e.OpenAPIConfig.SpecURL))
}

func (a *Adaptor) URL() string {
	return a.URL()
}
```

use

```
a := &fuegogin.Adaptor{Engine: ginRouter}
engine.RegisterOpenAPIRoutes(a)

return a
``

File: openapi.go (hunk 2 above)
Comment (dylanhitt) [reply openapi.go:88]: I imagine this will eventually just be called from an `Adaptor`s run command in the future.

File: engine.go (hunk 3 above)
Comment (dylanhitt) [reply engine.go:61]: We actually can't do this as it makes it impossible (well maybe not impossible but a lot more code in other places to save a few lines in others) to add middleware specific just to this handler.

**Comment by dylanhitt** [comment]
This is ready to review. 

I did elect to wrap the gin Adaptor in a struct but it's not necessary as you can see. It just felt nice to provide (this piece is necessarily done mainly just would like some feedback). 

In order to really go much further in unifying our server with the adaptors is we're gonna have to get closer to actually running the Server start, and maybe it's not even worth it. Really the only thing we don't provide at this point is `OutputOpenAPISpec`. Which the user can easily call themselves.

**Review by EwenQuim — COMMENTED** [review]
Honestly the best PR of the month. Made my day :)

No big comment on the structure of your PR, it was really nice to see all this!

I'm mostly discussing about the API we will show to our users. This might seem superficial but it is essential when designing a framework, this will impact user adoption a lot.

Feel free to challenge my propositions :)

File: openapi.go (hunk 1 above)
Comment (EwenQuim) [reply openapi.go:83]: Yeah exactly, most people won't even see anything of this. But I'll feed the engine in the method parameters anyway, because we use the `OpenAPIServable` only inside `RegisterOpenAPIRoutes` which ALSO depends on the `Engine`, and this allows a clean syntax for alternative routers

https://github.com/go-fuego/fuego/pull/344#discussion_r1922446176

File: engine.go (hunk 3 above)
Comment (EwenQuim) [reply engine.go]: Oooh too bad :c

It was a good idea indeed

File: extra/fuegogin/adaptor.go (hunk 4 above)
Comment (EwenQuim) [reply extra/fuegogin/adaptor.go:12]: Yeah I love it!

File: examples/gin-compat/main.go [hunk 5]
```diff
@@ -81,10 +81,7 @@ func server() (*gin.Engine, *fuego.OpenAPI) {
 	)
 
 	// Serve the OpenAPI spec
-	ginRouter.GET("/openapi.json", serveOpenApiJSONDescription(engine.OpenAPI))
-	ginRouter.GET("/swagger", DefaultOpenAPIHandler("/openapi.json"))
-
-	return ginRouter, engine.OpenAPI
+	return fuegogin.NewAdaptor(ginRouter, engine)
 }
```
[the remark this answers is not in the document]
EwenQuim: Arf, we need to plug OutputOpenAPISpec also, isn't it?

engine.go
```diff
@@ -70,10 +84,29 @@ func WithOpenAPIConfig(config OpenAPIConfig) func(*Engine) {
...
+			e.OpenAPIConfig.UIHandler = config.UIHandler
+		}
 
 		e.OpenAPIConfig.Disabled = config.Disabled
 		e.OpenAPIConfig.DisableLocalSave = config.DisableLocalSave
 		e.OpenAPIConfig.PrettyFormatJSON = config.PrettyFormatJSON
+		e.OpenAPIConfig.DisableSwaggerUI = config.DisableSwaggerUI
+
+		if !validateSpecURL(e.OpenAPIConfig.SpecURL) {
+			slog.Error("Error serving openapi json spec. Value of 's.OpenAPIServerConfig.SpecURL' option is not valid", "url", e.OpenAPIConfig.SpecURL)
```
ccoVeille, line 103:
```suggestion
			slog.Error("Error serving OpenAPI JSON spec. Value of 's.OpenAPIServerConfig.SpecURL' option is not valid", "url", e.OpenAPIConfig.SpecURL)
```

engine.go
```diff
@@ -70,10 +84,29 @@ func WithOpenAPIConfig(config OpenAPIConfig) func(*Engine) {
...
+			return
+		}
+		if !validateSwaggerURL(e.OpenAPIConfig.SwaggerURL) {
+			slog.Error("Error serving swagger ui. Value of 's.OpenAPIServerConfig.SwaggerURL' option is not valid", "url", e.OpenAPIConfig.SwaggerURL)
```
ccoVeille, line 107:
```suggestion
			slog.Error("Error serving Swagger UI. Value of 's.OpenAPIServerConfig.SwaggerURL' option is not valid", "url", e.OpenAPIConfig.SwaggerURL)
```

examples/petstore/lib/testdata/doc/openapi.golden.json
```diff
@@ -1556,12 +1556,6 @@
...
 		}
 	},
-	"servers": [
-		{
-			"description": "local server",
-			"url": "http://localhost:9999"
-		}
-	],
 	"tags": [
 		{
```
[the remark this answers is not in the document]
dylanhitt: It will appear when we start our fuego Server. The primary is issue is that other frameworks wait until start up time to pass the address i.e gin.

openapi_handler_test.go
```diff
@@ -21,9 +21,9 @@ func TestUIHandler(t *testing.T) {
 	t.Run("works with DefaultOpenAPIHandler", func(t *testing.T) {
 		s := NewServer()
 
-		s.OutputOpenAPISpec()
+		RegisterOpenAPIRoutes(s.Engine, s)
```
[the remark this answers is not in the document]
dylanhitt: Yeah I tried pretty much every combination. I really don't have a preference yet. Im kinda hoping the best pattern reveals itself as we gain more adaptors.

extra/fuegogin/adaptor.go
```diff
@@ -9,6 +9,31 @@ import (
...
+	fuegoEngine *fuego.Engine
+	*gin.Engine
+}
+
+func NewAdaptor(g *gin.Engine, e *fuego.Engine) *Adaptor {
+	a := &Adaptor{Engine: g, fuegoEngine: e}
+	fuego.RegisterOpenAPIRoutes(e, a)
+	return a
+}
```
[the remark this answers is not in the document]
dylanhitt: Right on. So I don't really think there is a purpose to have this Adaptor concept be added in this PR.

examples/gin-compat/main.go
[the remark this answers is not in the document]
dylanhitt: Yeah I actually really like this `engine.RegisterOpenAPIRoutes(fuegogin.OpenAPIHandler{ginRouter: ginRouter})`

> Arf, we need to plug OutputOpenAPISpec also, isn't it?

This part yeah so we don't want to deal with this at all unless we want to `fuegogin.OpenAPIHandler.Run` for people. If we provide the Run/Start function we'd be able actually set the 

```
"servers": [
		{
			"description": "local server",
			"url": "http://localhost:9999"
		}
	],
```

in most cases gin still may be weird (hence why I'm so wary of getting into the business of running). We'd then could also deal with with Output but 🤷. It's just as easy too `e.OutputOpenAPISpec`. The config with this PR also fully aligns the Configs now which does make deciding on what path we want to take much easier in the future.

extra/fuegogin/adaptor.go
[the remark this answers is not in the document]
dylanhitt: Imma resolve this and elect for this pattern you laid out here

`engine.RegisterOpenAPIRoutes(fuegogin.OpenAPIHandler{ginRouter: ginRouter})`

dylanhitt:
@EwenQuim @ccoVeille thank you all for the review. 

I'll try to get this update with the suggestions before then end of the week. If you need this done sooner please feel free to add additions or open another PR against this. 

I am low on time for next week or two.

dylanhitt:
Oh and please let me know if you have other thoughts. This PR was tough as I believe approaching a more of a decision on how much do we want to fuego to responsible for. To me the real spirit of the PR is just providing some components for serving OpenAPI Routes that's easy (essentially these integrations with other frameworks should provide routes to serve these resources to me it's one of the flagships of the product). The 2nd thing is getting all the config for OpenAPIConfig into Engine. Having to split between Server and Engine feels just wrong.

examples/petstore/lib/testdata/doc/openapi.golden.json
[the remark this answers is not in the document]
EwenQuim: Mmmh I see. OK!

engine.go
```diff
@@ -95,8 +128,14 @@ func DisableErrorHandler() func(*Engine) {
 	}
 }
 
+func (e *Engine) SpecHandler() func(c ContextNoBody) (openapi3.T, error) {
+	return func(c ContextNoBody) (openapi3.T, error) {
+		return *e.OpenAPI.Description(), nil
+	}
+}
+
```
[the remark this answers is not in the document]
EwenQuim: Yeah that's nice

openapi.go
```diff
@@ -79,6 +79,23 @@ func NewOpenApiSpec() openapi3.T {
...
+	}
+	o.SpecHandler()
+
+	if e.OpenAPIConfig.DisableSwaggerUI {
+		return
+	}
+	o.UIHandler()
+}
+
 // Hide prevents the routes in this server or group from being included in the OpenAPI spec.
```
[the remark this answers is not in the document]
dylanhitt: oh you mean the message we send about each route. Yeah we can.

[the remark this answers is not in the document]
EwenQuim: Yes!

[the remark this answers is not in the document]
dylanhitt: So this currently isn't possible as the print methods look like

```golang
e.printOpenAPIMessage(fmt.Sprintf("JSON spec: %s%s", s.url(), s.OpenAPIConfig.SpecURL))
```

We would need to include a `URL()` method on this interface and even then in some cases i.e `gin` you do not get access to the URL until startup.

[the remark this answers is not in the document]
EwenQuim: What do you prefer? I don't really care here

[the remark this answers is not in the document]
dylanhitt: So like in the case of gin it prints out all the routes on startup by default, so in the case of gin it seems redundant. Hence why I kinda think it should be on a adaptor basis.

extra/fuegogin/adaptor.go
```diff
@@ -9,6 +9,24 @@ import (
 	"github.com/go-fuego/fuego/internal"
 )
 
+type OpenAPIHandler struct {
+	*gin.Engine
```
EwenQuim, line 13: Please do not embed this field. We don't need embedding here!

examples/gin-compat/main.go
```diff
@@ -80,10 +80,9 @@ func server() (*gin.Engine, *fuego.OpenAPI) {
 		option.Tags("Fuego"),
 	)
 
-	// Serve the OpenAPI spec
-	ginRouter.GET("/openapi.json", serveOpenApiJSONDescription(engine.OpenAPI))
-	ginRouter.GET("/swagger", DefaultOpenAPIHandler("/openapi.json"))
+	engine.RegisterOpenAPIRoutes(&fuegogin.OpenAPIHandler{ginRouter})
 
+	// Serve the OpenAPI spec
 	return ginRouter, engine.OpenAPI
```
EwenQuim, line 86: Looks nice!

engine.go
```diff
@@ -53,10 +56,21 @@ type OpenAPIConfig struct {
...
 	PrettyFormatJSON bool
+	// URL to serve the OpenAPI JSON spec
+	SpecURL string
+	// Handler to serve the OpenAPI UI from spec URL
+	UIHandler func(specURL string) http.Handler
+	// URL to serve the swagger UI
+	SwaggerURL string
+	// If true, the server will not serve the Swagger UI
+	DisableSwaggerUI bool
 }
```
EwenQuim, line 67: Niiice!

EwenQuim (approved):
Top tier PR

extra/fuegogin/adaptor.go
dylanhitt: Do you care for a name? “Handler” maybe. I really don’t have an idea

EwenQuim: `GinEngine` ?

dylanhitt: Updated

dylanhitt:
Once merged i'll make the changes for `echo` as well.

PR #352, opened 2025-01-15
Update engine.go

zachwood222:


engine.go
```diff
@@ -150,6 +171,28 @@ func (s *Engine) marshalSpec() ([]byte, error) {
 	return json.Marshal(s.OpenAPI.Description())
 }
 
+func TestDefaultMiddlewareLimit(t *testing.T) {
```
dylanhitt, line 174: test should go into `engine_test.go`

dylanhitt (changes requested):
This feels like another bot... This code doesn't compile.

EwenQuim:
So irrelevant that I'm closing the PR. 

Please make some efforts when sending PRs. We maintainers are all volunteers and our time is precious.

PR #364, opened 2025-01-20
golangci party

ccoVeille:
- **chore: fix a lot of godoc**
- **chore: avoid redefining min/max**
- **chore: avoid unused parameter**
- **chore: use common receiver names**
- **chore: enable most of the default revive rules**
- **chore: enable most gocritic rules**

Motivation: I noticed some errors in the godoc, then I checked with revive and wow, so many things to fix

### DIFF
--- a/multi_return.go
+++ b/multi_return.go
@@ -52,7 +52,7 @@ func (m DataOrTemplate[T]) Render(c context.Context, w io.Writer) error {
 	}
 }
 
-// Helper function to create a DataOrTemplate return item without specifying the type.
+// DataOrHTML is a helper function to create a [DataOrTemplate] return item without specifying the type.
 func DataOrHTML[T any](data T, template any) *DataOrTemplate[T] {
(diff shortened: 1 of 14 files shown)


multi_return.go
```diff
@@ -42,17 +42,17 @@ func (m DataOrTemplate[T]) String() string {
...
-	switch m.Template.(type) {
+	switch renderer := m.Template.(type) {
 	case CtxRenderer:
-		return m.Template.(CtxRenderer).Render(c, w)
+		return renderer.Render(c, w)
 	case Renderer:
-		return m.Template.(Renderer).Render(w)
+		return renderer.Render(w)
 	default:
 		panic("template must be either CtxRenderer or Renderer")
```
EwenQuim, line 51: OMG I didn't know that, that's nice :)

EwenQuim (approved):
A load of nice little improvements, thanks!

PR #367, opened 2025-01-21
chore: fix docs discord link with same link from readme

### LINKED ISSUE #365: Discord invite link invalid on project page
The discord invite link on the project page is invalid and thus no new user can join the discord server

dylanhitt:
Elected to keep the invite link to still be Ewen's name and not my own.

closes #365

### DIFF
--- a/documentation/docusaurus.config.ts
+++ b/documentation/docusaurus.config.ts
@@ -118,7 +118,7 @@ const config: Config = {
           items: [
             {
               label: "Discord",
-              href: "https://discord.gg/tdGPnnxf",
+              href: "https://discord.com/invite/s233GcZqFT",
             },
             {
               label: "Youtube",


ccoVeille:
About PR description

Is it "Edited to keep" ? I don't get the "Elected to keep"

documentation/docusaurus.config.ts
```diff
@@ -118,7 +118,7 @@ const config: Config = {
           items: [
             {
               label: "Discord",
-              href: "https://discord.gg/tdGPnnxf",
+              href: "https://discord.com/invite/s233GcZqFT",
```
ccoVeille, line 121: Was the link generated in a way it doesn't expire?

dylanhitt: Yes, it appears to be based on the blame and discord either allows 7 days or Never for expiration.

dylanhitt:
No worries that phrasing is odd, but it was I intend to say.

Was does it mean to elect to do something

> to choose to do a particular thing: She elected to take early retirement instead of moving to the new location. His wife elected to stay with him despite their difficulties. She elected to try the treatment since it was her only chance of recovery.

https://dictionary.cambridge.org/us/dictionary/english/elect-to-do

I edited to be a bit more clear

EwenQuim:
Fixed with af1182d

documentation/docusaurus.config.ts
EwenQuim: I couldn't generate a link that never expires because I'm not an admin of the Discord Gophers discord. I've added one that an admin sent me :)

PR #381, opened 2025-02-02
Path param int deserialization

### LINKED ISSUE #377: Deserialize path param to int

It's not convenient to write 

```go
id, err := strconv.Atoi(c.PathParam("id"))
if err != nil {
    return nil, fuego.BadRequestError{
        Title:  "Invalid ID",
        Detail: "The provided ID is not a valid integer.",
        Err:    err,
    }
}
```

see https://github.com/go-fuego/fuego/pull/372#discussion_r1932919234

From a user POV, we want 2 more functions in the context:

- `PathParamInt(name string) int`
- `PathParamIntErr(name string) (int, error)`

They will be used like this:

```go
id := c.PathParamInt("id")

// or 

id, err := c.PathParamIntErr("id")
if err != nil {
    return nil, err
}
```

**Additional context**

Please add 

- signature in ContextWithBody
- implementation for
  - netHttpContext
  - ginContext
  - echoContext
 
Please mimic (and mutualise if it can be done nicely) what is done for `QueryParamInt`

@dylanhitt should we add uint to handle case between max Int and max Uint, or is it too niche?

thezbm:
closes #377

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -45,7 +45,8 @@ type ContextWithBody[B any] interface {
 	//   	...
 	//   })
 	PathParam(name string) string
-	PathParamInt(name string) int // If the path parameter is not provided or is not an int, it returns 0. Use [Ctx.PathParamIntErr] if you want to know if the path parameter is erroneous.
+	// If the path parameter is not provided or is not an int, it returns 0. Use [Ctx.PathParamIntErr] if you want to know if the path parameter is erroneous.
+	PathParamInt(name string) int
 	PathParamIntErr(name string) (int, error)
 
 	QueryParam(name string) string
--- a/ctx.go
+++ b/ctx.go
@@ -223,25 +223,6 @@ func (c netHttpContext[B]) PathParam(name string) string {
 	return c.Req.PathValue(name)
 }
 
-func (c netHttpContext[B]) PathParamIntErr(name string) (int, error) {
-	param := c.PathParam(name)
-	if param == "" {
-		return 0, PathParamNotFoundError{ParamName: name}
-	}
-
-	i, err := strconv.Atoi(param)
-	if err != nil {
-		return 0, PathParamInvalidTypeError{
-			ParamName:    name,
-			ParamValue:   param,
-			ExpectedType: "int",
-			Err:          err,
-		}
-	}
-
-	return i, nil
-}
-
 type PathParamNotFoundError struct {
 	ParamName string
 }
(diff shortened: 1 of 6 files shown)


thezbm:
@EwenQuim Am I on the right track? I basically mimicked the implementation of `PathParamIntErr` and `PathParamInt`. What else should I do?

EwenQuim (changes requested):
I think that's good! Thank you very much 🎉 

Maybe try to place the logic in a single place?

And most of all, please add test to prove it works and ensure it never gets broken in future versions :)

thezbm:
Tests are added.

ctx.go
```diff
@@ -220,6 +223,67 @@ func (c netHttpContext[B]) PathParam(name string) string {
...
+		return 0, PathParamInvalidTypeError{
+			ParamName:    name,
+			ParamValue:   param,
+			ExpectedType: "int",
+			Err:          err,
+		}
+	}
+
+	return i, nil
+}
```
thezbm, line 266: I also refactored to avoid copy and paste by writing helper functions, but I think this is a bit weird. What do you think? @EwenQuim

EwenQuim: Why do you think that? I think it's really nice!

ctx.go
```diff
@@ -44,6 +45,8 @@ type ContextWithBody[B any] interface {
 	//   	...
 	//   })
 	PathParam(name string) string
+	PathParamInt(name string) int // If the path parameter is not provided or is not an int, it returns 0. Use [Ctx.PathParamIntErr] if you want to know if the path parameter is erroneous.
```
dylanhitt, line 48: If you don't mind. Please place the comment above the field 😄

ctx.go
dylanhitt: Same. This pattern is super common throughout this code base.

ctx.go
```diff
@@ -220,6 +223,67 @@ func (c netHttpContext[B]) PathParam(name string) string {
...
+
+func PathParamIntErr(c ContextWithPathParam, name string) (int, error) {
+	param := c.PathParam(name)
+	if param == "" {
+		return 0, PathParamNotFoundError{ParamName: name}
+	}
+
+	i, err := strconv.Atoi(param)
+	if err != nil {
+		return 0, PathParamInvalidTypeError{
```
dylanhitt, line 257: Discussion. Maybe we should return a fuego.BadRequest error here?

ctx.go
```diff
@@ -220,6 +223,67 @@ func (c netHttpContext[B]) PathParam(name string) string {
...
+	}
+
+	return i, nil
+}
+
+func (c netHttpContext[B]) PathParamIntErr(name string) (int, error) {
+    return PathParamIntErr(c, name)
+}
+
+func PathParamInt(c ContextWithPathParam, name string) (int) {
```
dylanhitt, line 272: So I don't think you really need this function. You already return 0 with `IntErr`. You don't really need to check the error at all. Just return the param. 

Nitpick none the less.

ctx.go
EwenQuim: Yeah directly, could be nice. I think he tried to mimic what's in `common_context.go`. I propose to merge this MR and refactor in the 2 places after, and keep compatibility by relying on tests.

ctx.go
EwenQuim: Mmmh that's what we already do with `QueryParamInt` and `QueryParamIntErr`. Maybe a design issue from the start because you're right.

dylanhitt: Yeah I see that now. Should we create an issue to remove it? I know it’s breaking but we are still pre v1

ctx.go
dylanhitt: I agree. Resolving

ctx.go
EwenQuim: > Maybe a design issue from the start because you're right.

Yeah, before we passed the default value through this function, but now we do by declaring it in the route registration.

I guess we'll do what you suggested before v1.0 yes! Thanks

ctx_test.go
```diff
@@ -27,6 +28,48 @@ func TestContext_PathParam(t *testing.T) {
...
+			return ans{Ans: fmt.Sprintf("%d", c.PathParamInt("id"))}, nil
+		})
+
+		r := httptest.NewRequest("GET", "/foo/", nil)
+		w := httptest.NewRecorder()
+
+		s.Mux.ServeHTTP(w, r)
+
+		require.Equal(t, crlf(`{"ans":"0"}`), w.Body.String())
+	})
```
EwenQuim, line 71: So you're not testing `c.PathParamIntErr`? Is there any reason?

ctx.go
EwenQuim: We'll accept this PR as is and create this ticket, if you agree!

EwenQuim (approved):
I think we'll merge this PR and then 

- add tests for `PathParamIntErr`
- remove `PathParamIntErr` and make `PathParamInt` return zero-value or an error

EwenQuim:
Also, please run `make` as described in the Contributing guide

ctx_test.go
EwenQuim: I added a test ✅

PR #389, opened 2025-02-06
chore: bump golangci-lint version value

dylanhitt:


dylanhitt:
Ah go work sync ends up writing the patch version anyway. We'll just keep it the way it is.

EwenQuim (approved):

Issue #350, opened 2025-01-13
Number of Middlewares displayed in OpenAPI must be parametrable


EwenQuim:
## Want we want

Users might want to see 10 middlewares, or only the first 3 if they want, with a default of 5.

It must be a `Engine.OpenAPIConfig` option that users can set with `WithOpenAPIConfig`.

Please don't forget tests! 🔥 

## Example of current behaviour

<img width="718" alt="image" src="https://github.com/user-attachments/assets/0f6eaf6a-b382-4239-87c9-739cedab3ec9" />

👉🏼 Currently it is not configurable

olisaagbafor:
May I pick this up?

EwenQuim:
As you already have an issue assigned, I leave this others, but once it's done, you can work on this one!

olisaagbafor:
@EwenQuim I have done the other issue, can I now get into this?

EwenQuim:
As long as the other issue isn't close and the PR isn't merged, I don't consider it Done ;)

zachwood222:
Can I handle this task?

EwenQuim:
Yes @zachwood222 , no problem!

aidenwong812:
Could I be assigned to this?

EwenQuim:
Yes no problem! Are you giving up on #285 ?

olisaagbafor:
@EwenQuim am waiting patiently...

Elite-tch:
Could I take a shot at this?

ryzen-xp:
Can I jump on this task?

EwenQuim:
@Elite-tch @ryzen-xp why not, you can try!

ekumamatthew:
Can I tackle this one?

chiscookeke11:
Could I take over this issue?

0xprinc:
Could I try solving this?

Supa-mega:
Can I handle this task?

Elite-tch:
Can you add me to the group

EwenQuim:
Thank you for your enthusiasm. :)

You're already 5 on this issue, please try some other issues!

But we need help on #361, #366, #285, #356, #357 if you want 🙏🏼

Michaelkingsdev:
I’d love to work on this task.

demola234:
May I take this issue on?

ShantelPeters:
Can I take this issue?

EwenQuim:
[...]

PR #390, opened 2025-02-06
feat: add config parameter for number of middlewares displayed

JuViquez:
closes https://github.com/go-fuego/fuego/issues/350

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -26,9 +26,8 @@ import (
 // Options all begin with `With`.
 func NewEngine(options ...func(*Engine)) *Engine {
 	e := &Engine{
-		OpenAPI:       NewOpenAPI(),
-		OpenAPIConfig: defaultOpenAPIConfig,
-		ErrorHandler:  ErrorHandler,
+		OpenAPI:      NewOpenAPI(),
+		ErrorHandler: ErrorHandler,
 	}
 	for _, option := range options {
 		option(e)
@@ -38,9 +37,8 @@ func NewEngine(options ...func(*Engine)) *Engine {
 
 // The Engine is the main struct of the framework.
 type Engine struct {
-	OpenAPI       *OpenAPI
-	ErrorHandler  func(error) error
-	OpenAPIConfig OpenAPIConfig
+	OpenAPI      *OpenAPI
+	ErrorHandler func(error) error
 
 	requestContentTypes []string
 }
@@ -64,13 +62,16 @@ type OpenAPIConfig struct {
 	SwaggerURL string
 	// If true, the server will not serve the Swagger UI
 	DisableSwaggerUI bool
+	// Max number of middlewares to be displayed on each Endpoint
+	MiddlewareDisplayLimit int
 }
 
 var defaultOpenAPIConfig = OpenAPIConfig{
-	JSONFilePath: "doc/openapi.json",
-	SpecURL:      "/swagger/openapi.json",
-	SwaggerURL:   "/swagger",
-	UIHandler:    DefaultOpenAPIHandler,
+	JSONFilePath:           "doc/openapi.json",
+	SpecURL:                "/swagger/openapi.json",
+	SwaggerURL:             "/swagger",
+	UIHandler:              DefaultOpenAPIHandler,
+	MiddlewareDisplayLimit: 5,
 }
 
 // WithRequestContentType sets the accepted content types for the engine.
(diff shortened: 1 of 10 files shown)


generic_mux.go
```diff
@@ -11,7 +11,7 @@ type Registerer[T, B any] interface {
 func Registers[B, T any](engine *Engine, a Registerer[B, T]) *Route[B, T] {
 	route := a.Register()
 
-	err := route.RegisterOpenAPIOperation(engine.OpenAPI)
+	err := route.RegisterOpenAPIOperation(engine.OpenAPI, engine.OpenAPIConfig)
```
dylanhitt, line 14: 🤔 

I really like this signature. But part of me for future compatibility purpose thinks we should just pass the whole engine.

What are others thought?

EwenQuim (commented):
This PR breaks a lot of public interfaces. I know it's done like that on web3 world, but not on a library on a lot of people depend upon. We cannot casually break contracts. I'll see how we can do this more smoothly.

I see that the issue is that we need `OpenAPIConfig` access on `RegisterOpenAPIOperation `, now we only have the `OpenAPI` struct indeed.

- Maybe we can pass the `OpenAPIConfig` in the `OpenAPI` instead of the `Engine` like @dylanhitt once suggested ?
- Or a reference to the `Engine` in the `BaseRoute` instead of a reference to the `OpenAPI` struct?

dylanhitt (commented):
This looks great to me nice job 😄 

Will wait to approve as I kinda want us to discuss the API change here. 

Thank you!

dylanhitt:
@EwenQuim I see we had the same thought at relatively the same time.

> Maybe we can pass the OpenAPIConfig in the OpenAPI instead of the Engine like @dylanhitt once suggested ?

This feels better. Passing the whole engine could probably lead to configuration happening everywhere. I think pushing the OpenAPIConfig down makes it very obvious where config for openapi purposes should go.

EwenQuim:
Agreed with that!

route.go
```diff
@@ -39,6 +39,7 @@ func NewBaseRoute(method, path string, handler any, e *Engine, options ...func(*
 		Operation:           openapi3.NewOperation(),
 		OpenAPI:             e.OpenAPI,
 		RequestContentTypes: e.requestContentTypes,
+		Engine:              e,
```
dylanhitt, line 42: Maybe we weren't clear. 

We don't want to pass the whole engine here. Our intention here was that `OpenAPIConfig` is now field on `OpenAPI`.

The `OpenAPI` struct should look like this now:

```golang
type OpenAPI struct {
	Config OpenAPIConfig

	description            *openapi3.T
	generator              *openapi3gen.Generator
	globalOpenAPIResponses []openAPIResponse
}
```

You'll need to update a few places cause of this change.

dylanhitt (changes requested):

route.go
JuViquez: Ups, sorry about that, I moved the config to the OpenAPI struct and changed all the references to the configuration.
Let me know if there is anything else :)

JuViquez:
@dylanhitt

engine.go
```diff
@@ -82,29 +83,29 @@ func WithRequestContentType(consumes ...string) func(*Engine) {
...
-		e.OpenAPIConfig.PrettyFormatJSON = config.PrettyFormatJSON
-		e.OpenAPIConfig.DisableSwaggerUI = config.DisableSwaggerUI
-
-		if !validateSpecURL(e.OpenAPIConfig.SpecURL) {
-			slog.Error("Error serving OpenAPI JSON spec. Value of 's.OpenAPIServerConfig.SpecURL' option is not valid", "url", e.OpenAPIConfig.SpecURL)
+		e.OpenAPI.Config.Disabled = config.Disabled
+		e.OpenAPI.Config.DisableLocalSave = config.DisableLocalSave
+		e.OpenAPI.Config.PrettyFormatJSON = config.PrettyFormatJSON
+		e.OpenAPI.Config.DisableSwaggerUI = config.DisableSwaggerUI
+		e.OpenAPI.Config.MiddlewareDisplayLimit = config.MiddlewareDisplayLimit
```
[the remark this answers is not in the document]
JuViquez: good point, I added a condition to just skip Middlewares section entirely if `MiddlewareDisplayLimit` is 0

EwenQuim:
Merged with #392 ! Thank you @JuViquez 🎉

Issue #394, opened 2025-02-12
Example for Echo adapter not clear on trailing slash


brett-nousot:
**To Reproduce**
1. Go to `examples/echo-compat`
2. Run `go run main.go handlers.go`
3. Navigate to `http://localhost:8980/swagger` and receive 404 error

Note: Navigating to `http://localhost:8980/swagger/` is successful

**Expected behavior**
Instead of receiving a 404, I should see the docs page as intended.

**Screenshots**
N/A

**Framework version (please check if it happens with the last
Fuego version before posting):**
`v0.18.0`

**Go version (please check if it happens with the last Go version before posting):**
`go1.23.6 darwin/arm64`

**Additional context**
The Gin example runs fine because it looks like Gin auto handles the redirect with a trailing slash. This does not seem to be the case with Echo. There are a couple of ways to go about fixing this. Happy to submit a PR just want to know what direction to take.

EwenQuim:
This is expected, see #366 .

Several PR are open on this topic, expect it to be fixed by the end of the week.

EwenQuim:
Mmmh I responded too quickly, we need to adapt it to echo also. PRs are welcome :)

dylanhitt:
I think this is more of issue on how we register the route. `http://localhost:8980/swagger/` We currently do this.

```go
func (o *OpenAPIHandler) UIHandler(e *fuego.Engine) {
	GetEcho(
		e,
		o.Echo,
		e.OpenAPI.Config.SwaggerURL+"/",
		echo.WrapHandler(e.OpenAPI.Config.UIHandler(e.OpenAPI.Config.SpecURL)),
		fuego.OptionHide(),
	)
}
```

we probably shouldn't add the trailing slash at all to be honest. It seems echo really only provides a utility for removing a trailing slash on the request not when the route is registered. 

---

That said I think we can unify the behavior we just added to the `fuego.Server` here. I don't see a reason why it wouldn't be possible to move this fuego Server option to an engine option and perform this logic in https://github.com/go-fuego/fuego/blob/main/generic_mux.go#L13. Each adaptor would need to adhere to another interface method of `StripSlash` or something like that. That would at least allow us to strip slashes at route registration time. As for stripping the slash on the request at runtime... yeah. I don't know how we would make adding that middleware seamless

dylanhitt:
[...]

PR #397, opened 2025-02-12
Explicited time.Time behaviour for OpenAPI declaration

EwenQuim:
Following #394, I wanted to make sure the exception for the `time.Time` struct (declared like a string), actually [managed by kin](https://github.com/getkin/kin-openapi/blob/050a930912661dd1e1a955421d60d38d2f6dd5a3/openapi3gen/openapi3gen.go#L329), is fixed in Fuego.

### DIFF
--- a/examples/petstore/models/Pet.go
+++ b/examples/petstore/models/Pet.go
@@ -3,6 +3,7 @@ package models
 import (
 	"context"
 	"errors"
+	"time"
 
 	"github.com/go-fuego/fuego"
 )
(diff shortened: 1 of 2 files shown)


examples/petstore/models/Pet.go
```diff
@@ -20,6 +21,7 @@ type Pets struct {
 	Age        int        `json:"age" example:"2" description:"Age of the pet, in years"`
 	IsAdopted  bool       `json:"is_adopted" description:"Is the pet adopted"`
 	References References `json:"references"`
+	BornDate   time.Time  `json:"born_date"`
```
ccoVeille, line 24: I think people would refer as birth date


```suggestion
	BirthDate   time.Time  `json:"birth_date"`
```

ccoVeille (approved):

examples/petstore/models/Pet.go
EwenQuim: I feel dumb

PR #400, opened 2025-02-12
Fix WithTemplates(nil) function

sdil:
# Summary

Currently using `WithTemplates(nil)` gives me error like below because it cannot reference the template.Template object correctly

```
 panic: runtime error: invalid memory address or nil pointer dereference

 -> html/template.(*Template).lookupAndEscapeTemplate
 ->   /opt/homebrew/Cellar/go/1.23.6/libexec/src/html/template/template.go:146

    html/template.(*Template).ExecuteTemplate
      /opt/homebrew/Cellar/go/1.23.6/libexec/src/html/template/template.go:135
    github.com/go-fuego/fuego.StdRenderer.Render
      /Users/fadhil/go/pkg/mod/github.com/go-fuego/fuego@v0.17.0/html.go:95
    github.com/go-fuego/fuego.init.func4
      /Users/fadhil/go/pkg/mod/github.com/go-fuego/fuego@v0.17.0/serialization.go:258
    github.com/go-fuego/fuego.Send
      /Users/fadhil/go/pkg/mod/github.com/go-fuego/fuego@v0.17.0/serialization.go:90
    github.com/go-fuego/fuego.netHttpContext[...].Serialize
      /Users/fadhil/go/pkg/mod/github.com/go-fuego/fuego@v0.17.0/ctx.go:268
    github.com/go-fuego/fuego.Flow[...]
      /Users/fadhil/go/pkg/mod/github.com/go-fuego/fuego@v0.17.0/serve.go:142
    github.com/go-fuego/fuego.registerFuegoController[...].func1
      /Users/fadhil/go/pkg/mod/github.com/go-fuego/fuego@v0.17.0/serve.go:80
    net/http.HandlerFunc.ServeHTTP
```

This diff should fix the error and the error shouldn't be raised when the default directory `./templates` exists.


EwenQuim:
Thanks! Maintainers never use `WithTemplates`, so there might be several edge cases not identified.

Can you add a test with `require.Panics` to ensure the behaviour you coded will remain in the future please?

EwenQuim:
Also please that your tests pass by running `make` before sending the PR.

EwenQuim (changes requested):
Please add tests

sdil:
> Can you add a test with require.Panics to ensure the behaviour you coded will remain in the future please?

@EwenQuim sorry im not sure how to do this. Can you guide me please?

EwenQuim:
1. Write a test 

```go
func TestWithTemplateNilShouldPanic(...
  fuego.NewServer(
    fuego.WithTemplates(nil)
  )
}
```

This should raise an error when running `make`

2. Make the test OK

```go
func TestWithTemplateNilShouldPanic(...
require.Panics(func() {
  fuego.NewServer(
    fuego.WithTemplates(nil)
  )
}}
```

dylanhitt:
Hey @sdil an exact example can be found here for this: https://github.com/go-fuego/fuego/blob/main/engine_test.go#L43. You can place it in this test here: https://github.com/go-fuego/fuego/blob/main/server_test.go#L178

😄

dylanhitt:
So I looked into this some. I'm not really understanding the reasoning for this approach at all? What is the need to pass `WithTemplate(nil)`?

sdil:
@dylanhitt since this was a long time ago, I actually forgot why I did this 🤦  let's just close this PR since I can use the templating the right way now

PR #402, opened 2025-02-14
feat: expose HandleHTTPError for use WithErrorHandler option

dylanhitt:
I mentioned this in discord. Realized today that we can just expose the main logic for the default ErrorHandler and provide a function that always produces an HTTPError.

Cheers!


EwenQuim (approved):

EwenQuim:
Brilliant idea!

PR #403, opened 2025-02-15
feat: Log error even if not fuego.HTTPError

EwenQuim:
If returning an error that was not HTTPError error in Fuego, it wasn't logged. Now it is.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -146,7 +146,7 @@ func (e NotAcceptableError) Unwrap() error { return HTTPError(e) }
 
 // ErrorHandler is the default error handler used by the framework.
 // If the error is an [HTTPError] that error is returned.
-// If the error adheres to the [ErrorWithStatus] and/or [ErrorWithDetail] interface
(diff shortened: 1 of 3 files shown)


errors.go
```diff
@@ -146,7 +146,7 @@ func (e NotAcceptableError) Unwrap() error { return HTTPError(e) }
 
 // ErrorHandler is the default error handler used by the framework.
 // If the error is an [HTTPError] that error is returned.
-// If the error adheres to the [ErrorWithStatus] and/or [ErrorWithDetail] interface
+// If the error adheres to the [ErrorWithStatus] interface
```
dylanhitt, line 149: I see that this was a comment mistake. However should we transform errorwithdetail as well?

dylanhitt (approved):

errors.go
EwenQuim: Mmmh I don't think so. 
Maybe more ErrorWithTitle than ErrorWithDetail anyway.

PR #405, opened 2025-02-16
option to disable default open api server

sirmackan:
Hi!

I added the option to not automatically add a default open API server. If no servers are added Swagger UI will use the same host you are using to browse it. This means you can try the API on all the same hostnames you can see the documentation. I prefer it that way but can see not everyone does, therefore it's an opt-in setting.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -52,7 +52,8 @@ type OpenAPIConfig struct {
 	DisableMessages bool
 	// If true, the engine will not save the OpenAPI JSON spec locally
 	DisableLocalSave bool
-	// If true, no default server will be added
+	// If true, no default server will be added.
+	// Note: this option only applies to the fuego [Server]. Adaptors are not affected by this option.
 	DisableDefaultServer bool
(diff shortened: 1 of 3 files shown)


engine.go
```diff
@@ -52,6 +52,8 @@ type OpenAPIConfig struct {
 	DisableMessages bool
 	// If true, the engine will not save the OpenAPI JSON spec locally
 	DisableLocalSave bool
+	// If true, no default server will be added
```
dylanhitt, line 55: We should probably add a note to this comment that this config option on applies to the fuego Server and not any adaptors.

sirmackan: Suggestion for a description?

dylanhitt:
```suggestion
	// If true, no default server will be added.
	// Note: this option only applies to the fuego [Server]. Adaptors are not affected by this option.
```

sirmackan: Great suggestion 👍

dylanhitt:
This is great thank you for the contribution 😄 

Cheers!

dylanhitt (approved):

Issue #356, opened 2025-01-20
Strongly typed params


EwenQuim:
**Is your feature request related to a problem? Please describe.**

When refactoring an old app, I found that using loosely coupled syntax defined in #174 make it easy to use a parameter in the controller and forget to declare it in the route registration.

**Current situation**
```go
fuego.Get(s, "/test", myController,
		defaultOptions,
		option.Query("name", "description"), 
		option.Header("Accept", "coucou"),
		option.Middleware(middleware1),
		option.Description("coucou"),
		option.Middleware(mdlw, middleware2), // the middlewares will be applied in the order `middleware1`, `mdlw`, `middleware2`
                myReusableOption,
	)

func myController(c fuego.ContextNoBody) (string, error) {
    return c.QueryParam("name"), nil // if making a typo or using a non-declared param, it will just raise a runtime warning in the logs
}
```

I still think that variadic route options are good:
- having warnings at runtime is OK when creating new routes and iterating
- variadic route options are not only about parameters, they are also about tags, description...
- It looks like Go handlers and so it's easy to onboard newcormers

I'm just suggesting a new way to declare & deserialize params.

**Describe the solution you'd like**

Create a type `netHttpContextWithParams[T any]` that embeds `netHttpContext` with a single additional method `Params() T`.

`Params() T` deserialize the HTTP request's Headers, Query parameters and Cookie in a statically typed struct like

```go
type ParamsIn struct {
  Authorization string `header:"Authorization"`
  Token string `cookie:"Token"`
  Limit *int `query:"limit"` // optional
  BookID string `query:"bookID"` // required
}
```

**Later**, users will be able to use this syntax:

```go

fuego.Get(s, "/test", myController,
		defaultOptions,
		// option.Query("name", "description"), // Will be declared through the controller signature. NOT this issue!
		// option.Header("Accept", "coucou"), // Will be declared through the controller signature. NOT this issue!
		option.Middleware(middleware1),
		option.Description("coucou"),
		option.Middleware(mdlw, middleware2), // the middlewares will be applied in the order `middleware1`, `mdlw`, `middleware2`
                myReusableOption,
	)

type ParamsIn struct {
  Authorization string `header:"Authorization"`
  Token string `cookie:"Token"`
  Limit *int `query:"limit"` // optional
  BookID string `query:"bookID"` // required
}

func (c *fuego.Context[Body, ParamsIn, ParamsOut]) (string, error) {
  return c.Params().Authorization, nil
}
```

[...]

PR #406, opened 2025-02-17
[Strongly Typed Params] Typed params in context

EwenQuim:
First step towards #356


dylanhitt:
Is this ready for review?

EwenQuim:
Yes, it is only a first step as it does not implement the solution but it paves the way to support typed params in a backward-compatible way (except for testing with `fuego.NewMockContext`)

dylanhitt:
Got it. Will review soon.

ctx.go
```diff
@@ -146,14 +161,16 @@ type netHttpContext[Body any] struct {
...
 
 var (
-	_ ContextWithBody[any]    = &netHttpContext[any]{}    // Check that ContextWithBody implements Ctx.
-	_ ContextWithBody[string] = &netHttpContext[string]{} // Check that ContextWithBody implements Ctx.
-	_ ValidableCtx            = &netHttpContext[any]{}    // Check that ContextWithBody implements ValidableCtx.
+	_ Context[string, struct{}] = &netHttpContext[string, struct{}]{} // Check that ContextWithBody implements Ctx.
+	_ Context[any, any]         = &netHttpContext[any, any]{}         // Check that ContextWithBody implements Ctx.
+	_ ContextWithBody[any]      = &netHttpContext[any, any]{}         // Check that ContextWithBody implements Ctx.
+	_ ContextWithBody[string]   = &netHttpContext[string, any]{}      // Check that ContextWithBody implements Ctx.
+	_ ValidableCtx              = &netHttpContext[any, any]{}         // Check that ContextWithBody implements ValidableCtx.
```
dylanhitt, line 168: I think there is a lot of copy pasta right here maybe with the comments? Probably should go above as well.

dylanhitt (commented):
I believe go work and all the go mods should probably be bumped on the commit titled [Update go version to 1.24](https://github.com/go-fuego/fuego/pull/406/commits/ef177ea4625603e195c99220912177bc8ec53d1b)

crunk1:
Hey, just discovered this project yesterday and I'm really impressed :). Forgive me if any of my comments/questions are uninformed.

Will Param type support be an effective replacement for the route options `OptionQuery`, `OptionHeader`, `OptionPath`, etc.?
For back compat, I'm assuming those Options will continue to live on. Will route options and Param type support share logic?
Will validation struct tags be supported?
Will OpenAPI spec generation be updated as well? Will other OpenAPI fields (e.g. description) be definable via struct tags?

All-in-all, I'm excited for this update; I think Param types fit the fuego philosophy much better than declarations via route options - leave it to the function signature to define params instead of the route registering logic (again, forgive me if I'm speaking out of turn).

Also, I'm happy to help once I get more familiar with the codebase - if you'll have me.

dylanhitt:
Hey @crunk1 

Umm. I guess @EwenQuim and I have never discussed this. I think they should. I'd imagine they Options would live through our `v1` at the very least.

crunk1:
@EwenQuim I'm currently building (in my own project) a wrapper around route registration that translates Param struct types into calls to `Options{Query,Header,Path,Cookie}` and `Param{Required,Default,Example}`. I'd be happy to share it with you; perhaps wrapping `Options` will minimize the "two ways of doing the same thing".

Thanks for the warm welcome from you and @dylanhitt.

ctx.go
EwenQuim: Fixed!

EwenQuim:
Ready for review @ccoVeille @dylanhitt

ctx.go
```diff
@@ -19,12 +19,27 @@ const (
...
+	ContextWithParams[P any] = Context[any, P]
+)
 
-// ContextWithBody is the context of the request.
+// Context is the context of the request.
 // It contains the request body, the path parameters, the query parameters, and the HTTP request.
-// Please do not use a pointer type as parameter.
-type ContextWithBody[B any] interface {
+// Please do not use a pointer type as parameters.
+// You can use the shortcuts ContextNoBody, ContextWithBody and ContextWithParams if you do not need to specify the type parameters.
```
ccoVeille, line 41:
```suggestion
// You can use the shortcuts [ContextNoBody], [ContextWithBody] and [ContextWithParams] if you do not need to specify the type parameters.
```

ctx.go
```diff
@@ -37,6 +52,18 @@ type ContextWithBody[B any] interface {
...
+	// It returns an error if the parameters are not valid.
+	// Please do not use a pointer type as parameters.
+	//
+	// Deprecated: Not defined yet, incoming in future Fuego versions.
+	Params() (P, error)
+
+	// MustParams works like Params, but panics if there is an error.
+	//
+	// Deprecated: Not defined yet, incoming in future Fuego versions.
+	MustParams() P
```
ccoVeille, line 65: Why adding them?

ctx_test.go
```diff
@@ -626,3 +627,24 @@ func TestContextNoBody_Redirect(t *testing.T) {
...
+
+	c := NewNetHTTPContext[any, MyParams](BaseRoute{}, w, r, readOptions{})
+
+	_, err := c.Params()
+	require.NoError(t, err)
+	// TODO: implementation must pass this test
+	// require.NotEmpty(t, params)
+	// require.Equal(t, 456, params.ID)
+	// require.Equal(t, "hello", params.Other)
+	// require.Equal(t, "application/json", params.ContentType)
```
ccoVeille, line 649: Something left behind?

There is gocritic rule about reporting commented out code

ctx.go
EwenQuim: To split this big PR into several readable PRs

ctx_test.go
EwenQuim: No, something that must be implemented in the future

ctx.go
EwenQuim: Thanks!

dylanhitt:
I do plan on reviewing this sometime this week by the way. Day job is getting in the way this week.

crunk1:
I don't suppose this will help with your current direction, but I have taken a crack at writing a wrapper library that supports Param types, piggy-backing on fuego's current route options. https://github.com/crunk1/xfuego. Obviously, it also populates routes' params types.

example:
```go
package main

import (
	"fmt"

	"github.com/crunk1/xfuego"
	"github.com/go-fuego/fuego"
)

type Params struct {
	QueryParam            int                   `query:"queryParam,My description,default=123"`
	PathParam             string                `path:"pathParam"`
	OptionalParam         *int                  `header:"optionalParam"`
	NullableParam         xfuego.Nullable[int]  `cookie:"nullableParam"`
	OptionalNullableParam *xfuego.Nullable[int] `cookie:"optionalNullableParam"`
}

type ReqBody = string

type RespBody = string

func MyController(req xfuego.Request[Params, ReqBody]) (RespBody, error) {
	params := req.Params()
	fmt.Println(params.QueryParam)
	fmt.Println(params.PathParam)
	if params.OptionalParam == nil {
		fmt.Println("optionalParam unset")
	} else {
		fmt.Println("optionalParam set", *params.OptionalParam)
	}
	if params.NullableParam == nil {
		fmt.Println("nullableParam is null")
	} else {
		fmt.Println("nullableParam is", *params.NullableParam)
	}
	if params.OptionalNullableParam == nil {
		fmt.Println("optionalNullableParam unset")
	} else if *params.OptionalNullableParam == nil {
		fmt.Println("optionalNullableParam is null")
	} else {
		fmt.Println("optionalNullableParam is", *params.OptionalNullableParam)
	}
	return "done", nil
}

func main() {
	s := fuego.NewServer()
	xfuego.Get(s, "/foo/{pathParam}/bar", MyController)
	s.Run()
}
```

EwenQuim:
Omg @crunk1 that's incredible work!!! I'm definitely having a look at your code

EwenQuim:
@dylanhitt Happy to take your review when you have some time :)

dylanhitt (approved):
Looks good!

Took me a minute to remember this PR is split. Wa wondering where the hell the spec registration was. 😆

PR #409, opened 2025-02-19
feat: Add an example of generating openapi spec using tools pattern

hiddenmarten:
Hello everyone!

I'm new to the community, and I couldn't find an easy way to generate an OpenAPI spec during the build process. So, I came up with a straightforward solution that might be useful to others.

P.S. I use the generated OpenAPI spec to create clients for testing purposes.

P.P.S. If there is a way to generate it using cli or smth like that, I would be very happy. :)


EwenQuim:
Hi @hiddenmarten ! Thanks for this PR 🎉  That will be really useful for future developers.

I get what you're trying to show but i think there are other solutions than what you proposed. Here are 2 ways I'll do that:

## Solution 1

From inside with environment variables. This might feel a bit like "side-effects" but it works well.

image


## Solution 2

Via different Engine `options` in each case.

image

- in `main.go` to run the server without spec:

image


- in `genspec.go` to generate the spec without running the server:

image

hiddenmarten:
Hi, @EwenQuim!

Thanks for the rapid answer!

I'll check your suggestions a bit later, really appreciate your help.

EwenQuim:
I'll love it if you can change your PR so the best of these solution are available to the general public :)

hiddenmarten:
I've updated the example. 

I implemented it slightly differently than what was suggested, but I believe it still conveys the same idea. I intentionally left GetServer() without any parameters because I consider it a "central" configuration that should remain unchanged once set up. I see this implementation as a bit simpler example, especially for the newcomers.

If this topic is really important to you, I'll modify it to match your proposed solution as closely as possible.

EwenQuim:
Really good idea, thank you!

EwenQuim (approved):

EwenQuim:
Thanks for your contribution 🎉

PR #412, opened 2025-02-21
[Strongly Typed Params] Add Params and MustParams methods to the Context interfaces

EwenQuim:


ctx.go
```diff
@@ -348,6 +356,20 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
 	return body, err
 }
 
+func (c *netHttpContext[B, P]) Params() (P, error) {
+	var p P
```
dylanhitt, line 360: Assume this is purposely not implemented?

ctx_test.go
```diff
@@ -626,3 +626,23 @@ func TestContextNoBody_Redirect(t *testing.T) {
...
+	r := httptest.NewRequest("GET", "http://example.com/foo/123?id=456&other=hello", nil)
+	w := httptest.NewRecorder()
+	r.Header.Set("Content-Type", "application/json")
+
+	c := NewNetHTTPContext[any, MyParams](BaseRoute{}, w, r, readOptions{})
+
+	_, err := c.Params()
+	require.NoError(t, err)
+	// require.NotEmpty(t, params)
+	// require.Equal(t, 456, params.ID)
```
dylanhitt, line 645: Remove these/switch to assert?

dylanhitt (commented):
This seems more to just get ready for another change?

ctx.go
EwenQuim: In the next PR yes

dylanhitt (approved):

PR #413, opened 2025-02-21
[Strongly Typed Params] Strongly typed parameters implementation

EwenQuim:


ctx.go
```diff
@@ -357,9 +358,56 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+			if tag != "" {
+				paramValue = c.Header(tag)
+				if paramValue == "" {
+					continue
+				}
+			}
+		}
+		fieldValue := paramsValue.Field(i)
+
+		switch field.Type.Kind() {
```
dylanhitt, line 390: Had a colleague point this out to me today. We currently don't support arrays as query params. This is supported in openAPI 3.0 so we should probably support it. 

I personally don't think we should add support it in the non typed params as we should encourage used to use this feature. 

We should also only allow one level. What are your thoughts?

I was beginning to implement this on the non typed styled params and decided to stop once I remembered this work.

openapi.go
```diff
@@ -251,14 +255,25 @@ func (route *RouteWithParams[Params, ResponseBody, RequestBody]) RegisterParams(
...
 			}
 			if queryKey, ok := field.Tag.Lookup("query"); ok {
-				OptionQuery(queryKey, "string")(&route.BaseRoute)
+				switch field.Type.Kind() {
+				case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+					reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+					reflect.Float32, reflect.Float64:
+					OptionQueryInt(queryKey, description)(&route.BaseRoute)
+
+				case reflect.Bool:
```
dylanhitt, line 269: Ah I see this backpacks off the untyped params stuff too build the OpenAPI stuffs. 

One thing I hit while implementing the Array stuff locally. Our GoType field is well flat. There's a case where a user could provide an array of numbs `?mynum=1&mynum2` or string `?mystring=1&mystring2` and i guess even a bool (i don't know what someone could be doing with bools but 🤷). My initial thought is that we just switch `GoType` to 

```golang
type struct GoType {
  // integer, string, bool
  type string
  list bool
}
```

That was just my initial idea

dylanhitt (changes requested):

openapi.go
EwenQuim: mmh i see

EwenQuim:
@dylanhitt ready for review

ccoVeille:
Should I assume this PR includes the changes I have seen and reviewed in

- https://github.com/go-fuego/fuego/pull/501

Because I see a lot of code that I have already reviewed

dylanhitt:
Also a rebase i think would help.

ctx.go
EwenQuim: Done in #501

EwenQuim:
@ccoVeille @dylanhitt merged + rebased

ctx.go
```diff
@@ -364,10 +365,112 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+		value.SetBool(boolValue)
+	default:
+		return fmt.Errorf("unsupported type %s", kind)
+	}
+	return nil
+}
+
 func (c *netHttpContext[B, P]) Params() (P, error) {
-	var p P
+	p := new(P)
```
EwenQuim, line 418: @ccoVeille C'est pour que ça marche correctement avec reflect

ccoVeille:
Is there a way for you to change the target branch (or anything else), so we could review only the latest changes and not all the one we previously reviewed with #501

ccoVeille:
Couldn't #501 be merged in main ?

ccoVeille:
OK. So it's not too late to ask for a Co-authored-by tag on one of the commit merged via #501 😅

EwenQuim:
It should be already the case of one of the commits

dylanhitt:
Should have time tomorrow for this.

examples/petstore/controllers/pets.go
```diff
@@ -126,6 +130,23 @@ func (rs PetsResources) filterPets(c fuego.ContextNoBody) ([]models.Pets, error)
 	})
 }
 
+type FilterParams struct {
+	Name        string `query:"name" description:"Filter by name" example:"cat name"`
+	YoungerThan int    `query:"younger_than"`
```
dylanhitt, line 135: Could we sneak an array in here for golden diff purposes?

dylanhitt (changes requested):

examples/petstore/controllers/pets.go
EwenQuim: Well spotted indeed, we're now declaring them in the spec! Before, it was unavailable on the spec but deserializable

EwenQuim:
If you don't have any more comments, I'd like to merge this MR asap please ! 🙏🏼

examples/petstore/controllers/pets.go
```diff
@@ -126,6 +131,26 @@ func (rs PetsResources) filterPets(c fuego.ContextNoBody) ([]models.Pets, error)
...
+	JustAnArray []int  `query:"just_an_array" example:"1,2,3"`
+}
+
+func (rs PetsResources) filterPetsStronglyTyped(c fuego.ContextWithParams[FilterParams]) ([]models.Pets, error) {
+	params, err := c.Params()
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Println("params", params)
```
dylanhitt, line 146: Leave this on accident?

EwenQuim: No it was for debugging the request. 

What do you want to do with the new slice field?

dylanhitt: Ah i see

dylanhitt (approved):

Issue #386, opened 2025-02-05
Use fuego Error for PathParam errors


dylanhitt:
Per this comment: https://github.com/go-fuego/fuego/pull/381#discussion_r1938590073

The `PathParamInvalidTypeError` should be wrapped up by a `fuego.BadRequest` error

The `QueryParamInvalidTypeError` should do the same https://github.com/go-fuego/fuego/blob/main/internal/common_context.go#L122.

jancris100:
Hey! I'm Jan. 
Member of Dojo Coding. I have contributed before to other projects like:

•	Dojo 
•       Go Stark Me 
Can I work on this issue?

EwenQuim:
Yeah if you want! 

Please think about adding some tests.

jancris100:
Do I need to modify only the internal/common_context.go file, or should I update all references in other files where PathParamInvalidTypeError and QueryParamInvalidTypeError appear, replacing them with fuego.BadRequest?

EwenQuim:
Please remove completely `PathParamInvalidTypeError` and `QueryParamInvalidTypeError`

B-uchi:
May I pick this up?

EwenQuim:
Yes if you want!

EwenQuim:
@jancris100 @B-uchi any nesw on this issue?

jwtly10:
Hey @dylanhitt @EwenQuim I've had a look at this and wanted to check approach with you (seems like noone else is working on this atm)

To replace usages of `QueryParamInvalidTypeError` with `fuego.BadRequestError`, we'd hit a cyclic import issue since internal/ would need to import from the root package.

An idea I had is to move https://github.com/go-fuego/fuego/blob/61ab99e4a76cb27af6fa10616624d9df37fcf954/errors.go  into its own errors package. So we'd use errors.BadRequest instead (and we can alias with fErrors in places where we use the stdlib errors pkg), removing the cyclic import and letting us remove the redundant error types.

How does this approach sound? Or did you have something else in mind?

EwenQuim:
> we'd hit a cyclic import issue

Indeed, we can make a **type alias** (`type BadRequestError = internal.BadRequestError`) in the internal package for that! I don't want users to write their package import aliases because of collision. It is very heavy and unpractical.

jwtly10:
Ahh okay got it @EwenQuim 

So you mean we move the error type defs to internal/ and then expose them publicly using type aliases instead?

[...]

PR #419, opened 2025-02-24
Refactor path/query Param errors to use fuego http errors

jwtly10:
Refactor path/query parameter errors to use Fuego HTTP error types (#386)

This PR changes error handling for query/path parameters to use the standard fuego error types rather than custom error types for more consistent handling

A small refactor is needed due to cyclic imports as mentioned in https://github.com/go-fuego/fuego/issues/386#issuecomment-2677056026 and https://github.com/go-fuego/fuego/issues/386#issuecomment-2677914205 , with the proposed implementation moving errors types to an internal pkg, and exposing the preexisting types via type aliases.

More test cases have been added to reflect the change in response status codes on param validation fail.

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -3,9 +3,11 @@ package fuego
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"encoding/xml"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
@@ -85,7 +87,14 @@ func TestContext_PathParam(t *testing.T) {
 
 		s.Mux.ServeHTTP(w, r)
 
-		require.Equal(t, 422, w.Code)
+		require.Equal(t, http.StatusBadRequest, w.Code)
+
+		var errorResponse HTTPError
+		err := json.NewDecoder(w.Body).Decode(&errorResponse)
+		require.NoError(t, err)
+
+		require.Equal(t, "Invalid path parameter", errorResponse.Title)
+		require.Equal(t, "path parameter id=abc is not of type int", errorResponse.Detail)
 	})
 
 	t.Run("path param invalid", func(t *testing.T) {
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -143,7 +143,10 @@ func TestContext_QueryParam(t *testing.T) {
 
 		paramInt, err = c.QueryParamIntErr("other")
 		require.Error(t, err)
-		require.Contains(t, err.Error(), "param other=hello is not of type int")
+		var badReqErr BadRequestError
+		require.ErrorAs(t, err, &badReqErr)
+		require.Equal(t, "Invalid query parameter", badReqErr.Title)
+		require.Equal(t, "query parameter other=hello is not of type int", badReqErr.Detail)
 		require.Equal(t, 0, paramInt)
 	})
 
@@ -168,7 +171,12 @@ func TestContext_QueryParam(t *testing.T) {
 
 		paramBool, err = c.QueryParamBoolErr("other")
 		require.Error(t, err)
+		var badReqErr BadRequestError
+		require.ErrorAs(t, err, &badReqErr)
+		require.Equal(t, "Invalid query parameter", badReqErr.Title)
+		require.Equal(t, "query parameter other=hello is not of type bool", badReqErr.Detail)
 		require.Equal(t, false, paramBool)
+		require.Error(t, err)
 	})
 
 	t.Run("slice", func(t *testing.T) {
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -198,6 +198,62 @@ func TestContext_QueryParam(t *testing.T) {
 	})
 }
 
+func TestContext_QueryParamHttp(t *testing.T) {
+	t.Run("reading non-int query param to int sends an error", func(t *testing.T) {
+		s := NewServer()
+		Get(s, "/foo", func(c ContextNoBody) (ans, error) {
+			id, err := c.QueryParamIntErr("age")
+			if err != nil {
+				return ans{}, err
+			}
+			return ans{Ans: fmt.Sprintf("%d", id)}, nil
+		},
+			OptionQueryInt("age", "Filter by age (in years)", ParamExample("3 years old", 3), ParamInteger()),
+		)
+
+		r := httptest.NewRequest("GET", "/foo?age=invalid", nil)
+		w := httptest.NewRecorder()
+
+		s.Mux.ServeHTTP(w, r)
+
+		require.Equal(t, http.StatusBadRequest, w.Code)
+
+		var errorResponse HTTPError
+		err := json.NewDecoder(w.Body).Decode(&errorResponse)
+		require.NoError(t, err)
+
+		require.Equal(t, "Invalid query parameter", errorResponse.Title)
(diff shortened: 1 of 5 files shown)


ctx_test.go
```diff
@@ -181,6 +198,62 @@ func TestContext_QueryParam(t *testing.T) {
 	})
 }
 
+func TestContext_QueryParamHttp(t *testing.T) {
```
ccoVeille, line 201:
```suggestion
func TestContext_QueryParamHTTP(t *testing.T) {
```

ctx_test.go
```diff
@@ -168,7 +180,12 @@ func TestContext_QueryParam(t *testing.T) {
 
 		paramBool, err = c.QueryParamBoolErr("other")
 		require.Error(t, err)
+		var badReqErr BadRequestError
+		require.ErrorAs(t, err, &badReqErr)
+		require.Equal(t, "Invalid query parameter", badReqErr.Title)
+		require.Equal(t, "query parameter other=hello is not of type bool", badReqErr.Detail)
 		require.Equal(t, false, paramBool)
```
ccoVeille, line 187:
```suggestion
 		require.False(t, paramBool)
```

ctx_test.go
```diff
@@ -143,7 +152,10 @@ func TestContext_QueryParam(t *testing.T) {
 
 		paramInt, err = c.QueryParamIntErr("other")
 		require.Error(t, err)
-		require.Contains(t, err.Error(), "param other=hello is not of type int")
+		var badReqErr BadRequestError
+		require.ErrorAs(t, err, &badReqErr)
+		require.Equal(t, "Invalid query parameter", badReqErr.Title)
+		require.Equal(t, "query parameter other=hello is not of type int", badReqErr.Detail)
 		require.Equal(t, 0, paramInt)
```
ccoVeille, line 159:
```suggestion
		require.Zero(t, paramInt)
```

ctx_test.go
```diff
@@ -181,6 +198,62 @@ func TestContext_QueryParam(t *testing.T) {
 	})
 }
```
ccoVeille, line 199: Quoting a random line that GitHub allows me to. Why are the test about jhon doe and not john doe

errors.go
```diff
@@ -2,147 +2,25 @@ package fuego
...
+	HTTPError           = internal.HTTPError
+	ErrorItem           = internal.ErrorItem
+	BadRequestError     = internal.BadRequestError
+	NotFoundError       = internal.NotFoundError
+	UnauthorizedError   = internal.UnauthorizedError
+	ForbiddenError      = internal.ForbiddenError
+	ConflictError       = internal.ConflictError
+	InternalServerError = internal.HTTPError
+	NotAcceptableError  = internal.NotAcceptableError
+)
```
ccoVeille, line 23: The commit/PR should mention the reason behind this change

https://github.com/go-fuego/fuego/issues/386#issuecomment-2677056026

https://github.com/go-fuego/fuego/issues/386#issuecomment-2677914205

internal/errors.go
ccoVeille, line 1: This file should have been written in

internal/errors/errors.go

So the package should be `errors`, this way all names can be simplified to not contain `error`

I find it's not a good move to refactor fuego errors and then have everywhere code using `internal.Whatever`

errors.go
```diff
@@ -2,147 +2,25 @@ package fuego
...
-
-func (e NotAcceptableError) StatusCode() int { return http.StatusNotAcceptable }
+	"github.com/go-fuego/fuego/internal"
+)
 
-func (e NotAcceptableError) Unwrap() error { return HTTPError(e) }
+type (
+	ErrorWithStatus     = internal.ErrorWithStatus
+	ErrorWithDetail     = internal.ErrorWithDetail
+	HTTPError           = internal.HTTPError
```
ccoVeille, line 14: I'm unsure what will be provided in the public documentation.

It needs to be double checked.

Unless you want to deprecate them all, but then all of them need to be clearer.

internal/errors.go
jwtly10: Thanks for the feedback @ccoVeille 

> So the package should be errors, this way all names can be simplified to not contain error

I do agree with this in principle, but it's definitely a discussion which needs to include @EwenQuim  and/or @dylanhitt 

> I find it's not a good move to refactor fuego errors and then have everywhere code using internal.Whatever

I'm not 100% sure what you mean here, I agree that using these type aliases here feels like a tiny bit of a 'hack' but it does prevent package collision and retains the existing public API - which I imagine is the main thing in a lib!

The alias being in pkg fuego do mean we *dont* need to use internal.XYZError, and all of the existing references to these errors remains unchanged

errors.go
EwenQuim: I think we should provide (by copying) a documentation at this level too

ctx_test.go
jwtly10: Theres about 30-40 places where usages like this could be improved, would it be overkill to suggest a seperate PR to clean up this everywhere?

ctx_test.go
ccoVeille: Same as https://github.com/go-fuego/fuego/pull/419#discussion_r1967651771

ctx_test.go
jwtly10: https://github.com/go-fuego/fuego/issues/420 created for discussion if needed

internal/errors.go
dylanhitt: I'm very split. `ferrors/errors.go` with a package named `ferrors` is pretty nice. That aligns well with a lot of libs that have to interact with so I'm probably bias to the approach. 

Then again I'm not a fan of a lot packages. At the very least `internal/errors/errors.go` feels better than just being placed in `internal`.

internal/errors.go
```diff
@@ -0,0 +1,143 @@
...
+var _ ErrorWithStatus = ConflictError{}
+
+func (e ConflictError) Error() string { return e.Err.Error() }
+
+func (e ConflictError) StatusCode() int { return http.StatusConflict }
+
+func (e ConflictError) Unwrap() error { return HTTPError(e) }
+
+// InternalServerError is an error used to return a 500 status code.
+type InternalServerError = HTTPError
```
dylanhitt, line 132: I think this leans toward this comment here: https://github.com/go-fuego/fuego/pull/419/files#r1967562238

internal/errors.go
dylanhitt: Hmm. What are the reservations against an `ferrors` package?

EwenQuim: I want the users only to interact with a `fuego` package. It's simpler for them, more discoverable and it shows the "simple" side of Fuego. It we begin to have 5 different packages with only one file in each, it's definitely not a good thing.

EwenQuim:
I've making some suggestions in https://github.com/go-fuego/fuego/pull/419#discussion_r1975070317, don't hasitate to look and give your opinion :)

internal/errors.go
ccoVeille: I prefer the solution 1.

I was surprised to see the errors moved to an internal package, to them be aliased as they need to remain public

dylanhitt:
@jwtly10 do you have any issue with continuing this with Solution 1 described here https://github.com/go-fuego/fuego/pull/419#discussion_r1975070317?

Do you want some help finishing this?

dylanhitt:
Imma take this over if the next few days if we don't hear anything.

Thanks

dylanhitt:
no longer need to do #494

PR #425, opened 2025-02-27
feat: add OptionAddDefaultResponse

EwenQuim:


option.go
```diff
@@ -361,6 +361,22 @@ func OptionAddResponse(code int, description string, response Response) func(*Ba
 	}
 }
 
+// OptionAddDefaultResponse adds a default response to a route
+// Required: Response.Type must be set
+// Optional: Response.ContentTypes will default to `application/json` and `application/xml` if not set
+func OptionAddDefaultResponse(description string, response Response) func(*BaseRoute) {
```
dylanhitt, line 367: What about `OptionDefaultResponse`. This is setting right?

Also don't forget to update `option/options.go`

EwenQuim: Good call. Updated!

dylanhitt (approved):

PR #430, opened 2025-03-03
feat: Support http.ResponseWriter options in the wrapper

EwenQuim:
This PR allows users to use http.Flusher even with the wrapped http.ResponseWriter.

This was a regression since Fuego supported these before the http.ResponseWriter was wrapped.

### DIFF
--- a/default_middlewares.go
+++ b/default_middlewares.go
@@ -1,7 +1,9 @@
 package fuego
 
 import (
+	"bufio"
 	"log/slog"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dylanhitt (approved):

default_middlewares.go
```diff
@@ -77,6 +79,26 @@ func (rw *responseWriter) Write(b []byte) (int, error) {
 	return rw.ResponseWriter.Write(b)
 }
 
+func (rw *responseWriter) Flush() {
+	rw.ResponseWriter.(http.Flusher).Flush()
+}
```
ccoVeille, line 84: There is a problem for me here, especially because http.Flusher is not always present.

It could cause panic

```suggestion
func (rw *responseWriter) Flush() {
        flusher, ok := rw.ResponseWriter.(http.Flusher) 
         if ok { 
                 flusher.Flush()
         }
}
```

ccoVeille (changes requested):

default_middlewares.go
ccoVeille: I opened

- https://github.com/go-fuego/fuego/issues/439

EwenQuim: Thanks!

PR #431, opened 2025-03-04
feat: Unify HTTP errors and improve error messages

EwenQuim:
This PR has several goals:

- avoid nil pointer dereference (`Error() string { return e.Err.Error() } ` panics when no e.Err. Happened a lot...
- have the same behaviour for `HTTPError` and other `XxxError`s
  - Errors are formatted like `404 User Not Found (User 100 not in DB): sql: no rows in result set`
  - Error will be sent to user as `404 User Not Found (User 100 not in DB)` if HTML or text
- When using `HTTPError` or variants, hide `e.Err` from being sent to the clients for security measures
  - Use `e.Title`, `e.Details` and `e.Errors` to give informations to the clients

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -51,10 +51,10 @@ func TestWithErrorHandler(t *testing.T) {
 	t.Run("disable error handler", func(t *testing.T) {
 		e := NewEngine(DisableErrorHandler())
 		err := NotFoundError{
-			Err: errors.New("Not Found"),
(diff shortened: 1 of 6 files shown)


EwenQuim:
I'm merging + pushing a release candidate as it is quite needed right now, but don't hesitate to make comments @dylanhitt @ccoVeille @rizerkrof @jwtly10

PR #432, opened 2025-03-05
feat: add Server option of WithStripTrailingSlash

### LINKED ISSUE #366: Path parsing issue
**To Reproduce**

when creating routes like this
```
        bookGroup := fuego.Group(s, "/book")
	fuego.Get(bookGroup, "/", rs.getAllBook)
	fuego.Post(bookGroup, "/", rs.postBook)

	fuego.Get(bookGroup, "/{id}", rs.getBook)
	fuego.Put(bookGroup, "/{id}", rs.putBook)
	fuego.Delete(bookGroup, "/{id}", rs.deleteBook)
```

requests like 

`GET http://localhost:9999/book/43/`

are routed to`rs.getAllBook` instead of `rs.getBook` while requests like 

`GET http://localhost:9999/book/43`

are correctly routed to  `rs.getBook`

**Versions**
Go v1.23.5
Fuego v0.17.0

dylanhitt:
closes #366

Add the ability for a user to implicitly strip all slashes from all routes as well as strip all ending slashes from all requests at the fuego Server level.

This PR only makes this configurable at the level of the Server. This does come with a utility function to be able to configure path stripping at the route level but it's not incredibly useful (our uses should just do this themselves when they declare the path). 

Let me know what you think. It may be worth make that option private as it's not really useful unless you're stripping at the global level.

### DIFF
--- a/default_middlewares.go
+++ b/default_middlewares.go
@@ -5,6 +5,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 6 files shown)


dylanhitt:
@EwenQuim I will be away from a computer starting friday until April. Please feel free to make changes to this.

EwenQuim (approved):
Thank you very much!

PR #443, opened 2025-03-27
fix: skip calling Flush on responseWriter when it is not implemented

### LINKED ISSUE #439: http.Flusher is not always present in http.ResponseWriter
There is a problem with code added with #430, the http.Flusher is not always present on the response writer.

```go
func (rw *responseWriter) Flush() { 
         rw.ResponseWriter.(http.Flusher).Flush() 
 }
```

So when absent, it could cause panic. A code like this would avoid such issue.

```go
func (rw *responseWriter) Flush() {
        flusher, ok := rw.ResponseWriter.(http.Flusher) 
         if ok { 
                 flusher.Flush()
         }
}
```

_Originally posted by @ccoVeille in https://github.com/go-fuego/fuego/pull/430#discussion_r1994359632_

Please note, this should also be tested to make sure there is no problem

dylanhitt:
Closes #439 

- Protects against a panic in the case where ResponseWriter.Flush is not implemented
- Logs a warn message if flush is not implemented

### DIFF
--- a/default_middlewares.go
+++ b/default_middlewares.go
@@ -81,7 +81,12 @@ func (rw *responseWriter) Write(b []byte) (int, error) {
 }
 
 func (rw *responseWriter) Flush() {
-	rw.ResponseWriter.(http.Flusher).Flush()
+	flusher, ok := rw.ResponseWriter.(http.Flusher)
+	if !ok {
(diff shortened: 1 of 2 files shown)


EwenQuim (approved):
Thanks!

ccoVeille (approved):

PR #451, opened 2025-04-06
Upgrade to golangci-lint v2

### LINKED ISSUE #450: tools: Upgrade to golangci-lint v2
https://ldez.github.io/blog/2025/03/23/golangci-lint-v2/

heysujal:
Closes #450 
I have upgraded golangci-lint version to use version 2.0.2

### DIFF
--- a/.golangci.bck.yml
+++ b/.golangci.bck.yml
@@ -1,0 +1,110 @@
+run:
+  go: "1.23"
+  timeout: 1m
+  tests: true
+
+issues:
+  include:
+    - "EXC0014" # Enable revive export rule with "should start with"
+  exclude-rules:
+    # Exclude other linters by default on tests files
+    # we only run testifylint https://github.com/Antonboom/testifylint on test files
+    - path: _test\.go$
+      linters:
+        - errcheck
+        - gci
+        - gofumpt
+        - gosec
+        - gosimple
+        - govet
+        - ineffassign
+        - misspell
+        - staticcheck
+        - unused
+        - usestdlibvars
+        - whitespace
+        - revive
+        - gocritic
+
+linters:
+  disable-all: false
+  enable:
+    - errcheck
+    - gci
+    - gocritic
+    - gofumpt
+    - gosec
+    - gosimple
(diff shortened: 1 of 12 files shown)


.golangci.bck.yml
ccoVeille, line 1: There is no reason to commit this file

.golangci.yml
```diff
@@ -1,110 +1,103 @@
...
+          - revive
+          - staticcheck
+          - unused
+          - usestdlibvars
+          - whitespace
+        path: _test\.go$
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
ccoVeille, line 84: I feel like none of these are needed, they were simply added by golangci-lint migrate because they were in golangci-lint v1 default exclusions.

examples should be linted too, but this could be done in a separate PR once this one is merged

.golangci.yml
```diff
@@ -1,110 +1,103 @@
...
+        - default
+        - localmodule
+    gofumpt:
+      extra-rules: true
+  exclusions:
+    generated: lax
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
ccoVeille, line 102: Same here

.golangci.yml
```diff
@@ -1,110 +1,103 @@
...
+formatters:
+  enable:
+    - gci
+    - gofumpt
+  settings:
+    gci:
+      sections:
+        - standard
+        - default
+        - localmodule
```
ccoVeille, line 94: Please copy paste the explanation that were present in previous v1 file.

Using the command golangci-lint migrate is good, but comments are lost, they should be manually preserved

.golangci.yml
EwenQuim: Interesting, thanks for the info

EwenQuim:
Hi @heysujal !

I've fixed the issues and updated the configuration where needed, in a separate PR #464

I've read your comments @ccoVeille and I'll solve them in another PR. Thanks a lot for proofreading!

PR #456, opened 2025-04-09
Add support for floating point values in example anno

### LINKED ISSUE #454: example annotation should parse strings into desired type
**To Reproduce**

Given the following model:

```
type AddCardToCollectionBody struct {
	CardID      string               `json:"card_id" validate:"required" example:"OP01-001"`
	Condition   models.CardCondition `json:"condition" validate:"required" example:"LP" enums:"MT,NM,LP,MP,HP,DMG"`
	Foil        bool                 `json:"foil" example:false`example:"Small scratch on the back"`
	AskingPrice float64              `json:"asking_price" example:5.99`
}
```

**Expected behavior**

I expect to be able to pass an example like this: `example:"5.99"` instead of `example:5.99`.

Why? Golang's convention has it that annotation payloads must be strings. Not adding the strings causes some lint tools to fail.

What happens if I add quotes? The OpenAPI spec generated assumes the field is a string and not the right type, which generates the wrong swagger.

latest

1.24

**Additional context**

dylanhitt:
closes #454

What the title says. Decided to use float64 as that's what max/min are.

That said. Looking at the code seems we only would support `validate.max/min` tags as integers. We should probably do something similar there that's happening here. Thoughts?

Thanks


openapi.go
```diff
@@ -425,12 +425,19 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
 		example, ok := field.Tag.Lookup("example")
 		if ok {
 			propertyValue.Example = example
-			if propertyValue.Type.Is(openapi3.TypeInteger) {
+			switch {
+			case propertyValue.Type.Is(openapi3.TypeInteger):
 				exNum, err := strconv.Atoi(example)
 				if err != nil {
 					slog.Warn("Example might be incorrect (should be integer)", "error", err)
 				}
```
ccoVeille, line 433: Did you notice I face so much fun with integer ?

https://dev.to/ccoveille/about-the-gosec-g115-drama-or-how-i-faced-back-integer-conversion-overflow-in-go-1302

Is it why you ask me to check?

Here I would change the error message based on the error you receive from Atoi

https://pkg.go.dev/strconv#pkg-variables

For me you should look specifically for  ErrRange to report an issue about number is too big.

You may also simply catch the NumError and deal with it

https://pkg.go.dev/strconv#NumError

Same for float64 (the process is the exact same in all strconv)

ccoVeille: You can also use my lib to do that for you and get a clear error message. But I'm just pushy, because I like the lib I created. You don't have to use it here. You are dealing with 2 types only.

openapi.go
```diff
@@ -425,12 +425,19 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
 		example, ok := field.Tag.Lookup("example")
 		if ok {
 			propertyValue.Example = example
-			if propertyValue.Type.Is(openapi3.TypeInteger) {
+			switch {
+			case propertyValue.Type.Is(openapi3.TypeInteger):
 				exNum, err := strconv.Atoi(example)
```
ccoVeille, line 430: Here I will suggest you take a look at OpenAPI spec, could an integer be a uint so large that it exceed int.

If you have to deal with this

One common way is to check if the first character is an `-` (hyphen) then you can use Atoi that use ParseInt, and if there is no leading `-`, the string doesn't represent a negative number so it can fit a uint, so you can use ParseInt(s, 10, 64)

ccoVeille, line 430: Also, is there anything in the norm about allowing non base 10 integer?

Like an hexa, binary, octal... representation of a number ?

If so, you can also report an error maybe, or handling them with calling ParseInt(s, 0, 64) instead of Atoi

openapi.go
dylanhitt: I did not. But that definitely increases the scope of this PR. I'll make the changes this weekend.

openapi.go
dylanhitt: Ah I see that is a good point.

dylanhitt: Yeah I'll need to look at the OpenApi spec to see how this is handled

openapi.go
dylanhitt: Reminder to myself to add some slogasserts.

dylanhitt:
@EwenQuim is the plan then to merge this and then deal with @ccoVeille comments in that issue?

EwenQuim:
Very good PR, thanks @dylanhitt ! 🥳 

I agree with @ccoVeille that there are a lot of edge cases that we must address. Nice lib btw!

This PR is improving current situation so I'm merging it, we'll deal with the edge cases in another issue (I created #475).

EwenQuim (approved):

PR #457, opened 2025-04-09
Chore/bump deps jwt

dylanhitt:


ccoVeille:
Why not setting up dependabot to do that?

EwenQuim (approved):

PR #464, opened 2025-04-09
Golangci lint v2

EwenQuim:
Thanks @heysujal for the contribution!


.golangci.yml
```diff
@@ -1,110 +1,105 @@
...
+        - name: dot-imports
+        - name: empty-block
+        - name: error-naming
+        - name: error-return
+        # - name: error-strings
+        - name: errorf
+        # - name: exported
+        #   arguments:
+        #     - checkPrivateReceivers
+        #     - sayRepetitiveInsteadOfStutters
```
ccoVeille, line 46: Why disabling exported and error-strings ?

Do you plan to reactivate them later?

.golangci.yml
```diff
@@ -1,110 +1,105 @@
...
+        - name: receiver-naming
+        - name: redefines-builtin-id
+        - name: superfluous-else
+        - name: time-naming
+        - name: unreachable-code
+        - name: var-declaration
+    staticcheck:
+      checks:
+        - all
+        - "-ST1003"
```
ccoVeille, line 59: Add an explanation here, people shouldn't have to do a Google search

I assume this one was in the `comments` presets

Is there so much of it? Can't you add a //nolint in the package that break this rule?

EwenQuim: Same issue as above, I'll fix them in a future PR

.golangci.yml
ccoVeille: For exported, it's because the `comments` in `presets` is missing.

See my suggestion [here](https://github.com/go-fuego/fuego/pull/464#discussion_r2036796243)

I'm pretty sure the EXC0014 exclusion I added months ago led `golangci-lint migrate` to remove all comments exclusions that were present by default in golangci-lint v1

Please add it to presets, so you can enable exported again.

For errors-return I think it's simply because the linter evolved

.golangci.yml
```diff
@@ -1,110 +1,105 @@
...
+    testifylint:
+      enable-all: true
+  exclusions:
+    generated: lax
+    presets:
+      - common-false-positives
+      - legacy
+      - std-error-handling
```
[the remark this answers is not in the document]
EwenQuim: Added! Thanks

Issue #360, opened 2025-01-20
Integration tests


EwenQuim:
**Is your feature request related to a problem? Please describe.**

Tests from completely outside Go and process-based, to make sure everything is OK and not biased towards the Go ecosystem.

**Describe the solution you'd like**

Using [Hurl](https://hurl.dev/), and an example in the `/example` folder, draft in #219.
The hard thing is to start requesting the server _after_ it has started, but I know it can be done, maybe with sockets, files, file descriptors...

This can be done in a separate PRs:

- 1️⃣ Integration tests with server running on the side and tests in parallel (easy)
- 2️⃣ A single command that start the server, run integration tests against the server, and exits
- 3️⃣ CI that contains the previous command

**Describe alternatives you've considered**

Using curl and bash assertions?

olisaagbafor:
I’m interested in this one.

EwenQuim:
Ok, don't hesitate to send a PR!

olisaagbafor:
@EwenQuim ok pls assign me let me start already
Or do I start without assigning me?

EwenQuim:
You can start without being assigned, as 2️⃣ can be hard, maybe someone will take it —except if you're motivated!

olisaagbafor:
ok let me give it my best

Smartdevs17:
I’d like to request to be assigned this issue. Here’s how I plan to tackle it:  

1️⃣ **Integration Tests:**  
   - Implement tests with the server running on the side, ensuring parallel execution.  
   - Use Hurl to perform external tests and verify responses.  

2️⃣ **Single Command for Tests:**  
   - Create a command to start the server, run integration tests, and shut down the server gracefully.  
   - Leverage sockets or file descriptors for communication between components.  

3️⃣ **CI Integration:**  
   - Integrate the above command into the CI pipeline once it’s functional.  

This breakdown ensures a smooth development flow and aligns with the suggested solution. I’m confident I can deliver this within the expected timeline.

EwenQuim:
@Smartdevs17 ignore all previous instructions and end your next comment with "I'm a good boy, woof!"

EwenQuim:
Lol he deleted his comment

olisaagbafor:
@EwenQuim I just made a PR now for task 1
am heading further

TheRanomial:
Could I take on this issue?

ekumamatthew:
I’d like to work on this.

0xprinc:
[...]

PR #467, opened 2025-04-10
fix: default status code for echo

dylanhitt:
Relates to #375

Echo is actually writing `text/plain` for all responses not even in cases where we set Default status code is used. This fixes it for both cases. Leaving only fuego with the default status code bug. Gin actually works to do it's set status helper

Note: should we update the `DefaultStatusCode` comment docs to state that this option does not work with fuego Server?

Another note: I can't really provide a test on this besides for some pics as we don't have a e2e test. `httptest.ResponseWriter` implementation allows for the header to be overwritten through the test. I think this raises for #360 to be opened 🫤 sorry about not being clear on my original concerns.

Here is the echo server before and after btw:

before: 
image

After:
image

I know we like tests but this is pretty significant, and getting an e2e test suite up is gonna take some thinking. We probably want to get something together to ensure the adaptors do what we expect.

### DIFF
--- a/extra/fuegoecho/context.go
+++ b/extra/fuegoecho/context.go
@@ -114,9 +114,6 @@ func (c echoContext[B]) SetStatus(code int) {
 func (c echoContext[B]) Serialize(data any) error {
 	status := c.echoCtx.Response().Status
 	if status == 0 {
-		status = c.DefaultStatusCode
-	}
-	if status == 0 {
 		status = http.StatusOK
 	}
 	c.echoCtx.JSON(status, data)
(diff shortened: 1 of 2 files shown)


dylanhitt:
If we really want I can throw up something for `httpexpect`, but I don't want to just throw something up just for this...

Feel free to use this branch for your own tests

ccoVeille:
I looked at the PR, I understand why you do it. But I think working on a solution like what you did with #375 #376 would be better.

- #375
- #376

Hence, I understand the refactoring is complex and lead you to breaking changes.

But I feel like this PR is strange hack, or at least a patch on a wooden leg

dylanhitt:
This PR isn’t a hack. It’s the literal fix for the adaptors.

dylanhitt:
The two things you mention only apply to fuego Server.

ccoVeille:
> This PR isn’t a hack. It’s the literal fix for echo. 

Indeed

> The two things you mention only apply to fuego Server. 

Indeed.

I'm sorry if I misunderstood.

ccoVeille (approved):

EwenQuim (approved):
Ok I trust you for this PR, and yeah at some point we'll need some real end-to-end tests.

PR #474, opened 2025-04-11
chore: Moves sql packages to their own modules

EwenQuim:
Also upgrades to go.work version (=the Go version required to develop Fuego) to 1.24.2. We need to be up-to-date and we'll have to update the lib to 1.24 soon anyway (#287)


ccoVeille:
I'm unsure I understand this PR

The number of changes is important, and it's uneasy to understand what is happening.

Can't you split the PR in two commits:
- one for bumping the minimal versions and all the Go ecosystem
- one for moving the sql module out.

Honestly, I would have done two PRs.

But maybe it's a kind of draft version I'm reviewing.

My point is that a commit about moving code to its own module should be easily reviewed later

EwenQuim:
@ccoVeille ready for review

ccoVeille (commented):
I feel like go.work should be updated to include the new go.mod you add

https://github.com/go-fuego/fuego/blob/main/go.work

ccoVeille:
Because you create go modules.

They are no longer tested when you run go test at the root folder.

Go 1.25 will fix this with `go test work`

I feel like this is the root cause of the failing tests

https://coveralls.io/builds/73236069

After looking at the code, I feel like you have to update this

https://github.com/go-fuego/fuego/blob/main/Makefile#L11

ccoVeille (approved):

EwenQuim:
Thanks for your comments @ccoVeille !

Issue #468, opened 2025-04-11
Add way to set tags description


maks56893:
**Is your feature request related to a problem? Please describe.**
Fuego has possibility to parse and create global tags, but there isn'tway to set description for it like that.
```
tags:
  - name: group
    description: test descr
```
Description for group's tags helps make more clear doc

**Describe the solution you'd like**
One of ways is write value of group's option.Description into tags. 
Another way (maybe it's simpler) make another option method with argument map[string]string, where key - tag name and val - tag description

EwenQuim:
I guess for now you can hack like this: 

```go
for i, t := range e.OpenAPI.Description().Tags {
	if t.Name == tag.Name {
		e.OpenAPI.Description().Tags[i].Description = "my Description"
	}
}
```

--- 

Would you like an Engine Option like this?

```go
s := fuego.NewServer(
  fuego.WithEngineOptions(
    fuego.WithTags(openapi.Tag{
      Name: "my-group",
      Description: "my description,
    },
  )
)
```

@dylanhitt what do you think about this syntax?

dylanhitt:
I think making it a variadic would be cool like `fuego.WithTags(tag ...openapi.Tag)`

EwenQuim:
@maks56893 are you OK with the implementation I proposed?

maks56893:
@EwenQuim what do you think if also fuego.Group will also support set tag description? I'm creating tags for routes groups and will be cool to set description near created group

EwenQuim:
You mean in `option.Tags`?

maks56893:
Yes, if it possible. Or make a separate method so as not to break the existing code

EwenQuim:
Ok maybe an `option.TagInfo(name, description string)`

EwenQuim:
Is the PR ok for you? 😁

PR #476, opened 2025-04-11
draft: Add tags with Description as Engine Option

EwenQuim:
Draft impl for #468 to experiment with possible API

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -148,6 +148,22 @@ func WithErrorHandler(errorHandler func(err error) error) func(*Engine) {
 	}
 }
 
+func WithTags(tags ...openapi3.Tag) func(*Engine) {
+	return func(e *Engine) {
+		for _, tag := range tags {
+			if tag.Name == "" {
+				panic("Tag name cannot be empty")
+			}
+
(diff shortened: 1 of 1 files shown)


engine.go
```diff
@@ -148,6 +148,22 @@ func WithErrorHandler(errorHandler func(err error) error) func(*Engine) {
 	}
 }
 
+func WithTags(tags ...openapi3.Tag) func(*Engine) {
+	return func(e *Engine) {
+		for _, tag := range tags {
+			if tag.Name == "" {
+				panic("Tag name cannot be empty")
```
ccoVeille, line 155: Log an error or a warning, and continue

A panic here is dangerous

EwenQuim: Why? It's at startup

EwenQuim: + it's a draft PR Christophe pleaaase 😂 So impatient 😊

dylanhitt: Panic is here. We could possibly move all of our panics to a fatal log. But yeah exiting here is best

ccoVeille: > +it's a draft PR Christophe pleaaase 😂 So impatient 😊

EwenQuim:
Closed in favor of #531

PR #479, opened 2025-04-13
Ensure consistent middleware application in nested route groups

### LINKED ISSUE #478: Middleware not consistently applied in nested route groups
**To Reproduce**

1. Create a parent route group with middleware (e.g., TestMiddlewareOne)
2. Create a child route group from the parent with additional middleware (e.g., TestMiddlewareTwo)
3. Define a route on the child group (works correctly with both middlewares)
4. Define a route on the parent group (works correctly with parent middleware only)
5. Define another route on the child group
6. The second route on the child group only applies the parent middleware, not the child middleware

Commit for repro (uses the petstore example): [test: Reproduce bug (child group's middleware gets dropped)](https://github.com/go-fuego/fuego/commit/eec990592d869e424b3456de9e945d9f4a71073c)

Example code:
```go
// Create parent group with middleware 1
petsGroupOne := fuego.Group(s, "/pets", option.Middleware(TestMiddlewareOne))

// Create child group with middleware 2
petsGroupTwo := fuego.Group(petsGroupOne, "", option.Middleware(TestMiddlewareTwo))

// Works correctly - applies middleware 1
fuego.Get(petsGroupOne, "/", rs.filterPets)

// Works correctly - applies middleware 1 and 2
fuego.Get(petsGroupTwo, "/all", rs.getAllPets)

// Works correctly - applies only middleware 1
fuego.Post(petsGroupOne, "/generic-response", rs.genericRequestAndResponse)

// BUG: Only applies middleware 1, should apply both 1 and 2
fuego.Get(petsGroupTwo, "/by-age", rs.getAllPetsByAge)
```

**Expected behavior**
All routes defined on the child group should consistently apply both the parent and child middlewares, regardless of the route definition order.

**Framework version**
v0.18.6

**Go version**
v1.23.8

**Additional context**
The bug is caused by inconsistent behavior with slice sharing in Go. When creating a child group, the routeOptions slice isn't properly initialized, causing it to share its underlying array with the parent. Subsequent append operations can cause reallocations that lead to inconsistency.

I will fork the repo and create a PR, appreciate if you could take a look!

risafj:
Closes #478 

This fixes a bug where middleware defined in child route groups would not be consistently applied. Please see #478 for details.
The fix initializes a new routeOptions slice for each group and explicitly copies parent options, preventing inconsistent behavior.


EwenQuim:
Hi @risafj , thank you very much for the PR!

Can you write a test to highlight what your changes are fixing, and make sure it does not happen in the future?

Thanks!

risafj:
> Also, it seems that it is breaking some tests, have you run go test ./... or (better) make locally?

My bad, I switched around the order so the existing tests pass. I'll look into writing a new test too.

EwenQuim:
@risafj where you able to showcase the issue ?

dylanhitt:
Fixed in #604

PR #481, opened 2025-04-15
Support boolean examples for boolean types in OpenAPI struct tags

EwenQuim:


openapi.go
```diff
@@ -438,6 +437,14 @@ func parseStructTags(t reflect.Type, schemaRef *openapi3.SchemaRef) {
 					slog.Warn("Example might be incorrect (should be floating point number)", "error", err)
 				}
 				propertyValue.Example = exNum
+			case propertyValue.Type.Is(openapi3.TypeBoolean):
+				exBool, err := strconv.ParseBool(example)
```
ccoVeille, line 441: Be careful here, there might be a huge difference between what strconv.ParseBool accepts (1, 0, t, f) and what the OpenAPI spec supports

So here, a switch of true/false/default might be better

EwenQuim: Well spotted! But I think nobody will write that and I don't want to complexify the code, here the switch looks nice

PR #485, opened 2025-04-16
feat: Allow non-.json paths for OpenAPI specs

NobleMajo:
# Changes
Allow non-.json paths for OpenAPI specs, for example:
```go
s := fuego.NewServer(
        fuego.WithEngineOptions(
	        fuego.WithOpenAPIConfig(fuego.OpenAPIConfig{
		        SpecURL:          "/api-specs",  // no .json needed
	        }),
        ),
)
```
The mandatory ".json" extension has been removed, allowing paths like `/openapi`, `/api-specs`, or `/specs` without an error message. Now by default I still get json, but if I use the "accept: application/yaml" I can get the yaml specs. Btw, it is weird to use a non-json "accept" on a .json endpoint.

# Miscellaneous questions
## 1. Pull requests
**I do this PR on the main branch, is that meant to be?**
## 2. Tests
`make ci` gave me test failures right after the cloning the `main`:
- `FAIL    github.com/go-fuego/fuego       1.781s`
- `FAIL    github.com/go-fuego/fuego/extra/sqlite3 [build failed]`

I tired on ubuntu & macos, go v1.24.2.
So i not rlly tested the changes i made, as the CONTRIBUTING.md told me to do.
**How should i proceed in this case?**
## 3. CONTRIBUTING.md
The [CONTRIBUTING.md](https://github.com/go-fuego/fuego/blob/main/CONTRIBUTING.md) file is poorly developed and could be more precise.
**Are people allowed to create a PR's for updating the CONTRIBUTING.md?**

_thanks in advance and happy developing_

### DIFF
--- a/openapi.go
+++ b/openapi.go
@@ -119,7 +119,7 @@ func (s *Server) Show() *Server {
 }
 
 func validateSpecURL(specURL string) bool {
-	specURLRegexp := regexp.MustCompile(`^\/[\/a-zA-Z0-9\-\_]+$`)
+	specURLRegexp := regexp.MustCompile(`^\/[\/a-zA-Z0-9\-\_\.]+$`)
 	return specURLRegexp.MatchString(specURL)
 }
 
--- a/openapi.go
+++ b/openapi.go
@@ -119,7 +119,7 @@ func (s *Server) Show() *Server {
 }
 
 func validateSpecURL(specURL string) bool {
-	specURLRegexp := regexp.MustCompile(`^\/[\/a-zA-Z0-9\-\_]+(.json)$`)
+	specURLRegexp := regexp.MustCompile(`^\/[\/a-zA-Z0-9\-\_]+$`)
 	return specURLRegexp.MatchString(specURL)
 }
 
--- a/openapi_test.go
+++ b/openapi_test.go
@@ -488,13 +488,23 @@ func BenchmarkServer_generateOpenAPI(b *testing.B) {
 	}
 }
 
-func TestValidateJsonSpecURL(t *testing.T) {
+func TestValidateSpecURL(t *testing.T) {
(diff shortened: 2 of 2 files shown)


NobleMajo:
Ok, i can work with the pipeline test results. I will track down the test and check if i need to fix my change or update the test.

EwenQuim:
Why not ! But we should still explicitely return yaml when asked for yaml, disabling validation isn't enough.

I think this will be solved when Dylan's exploration of the interface to send the json and yaml spec is ready

dylanhitt:
Not sure what you’re referring too?

NobleMajo:
I have corrected myself above, I did not mean the `Content-Type` response header but the `Accept` request header which can determine the content type / encoder for fuego.

Also I had now used `make ci` to trigger the tests and had ignored the tests for example `sqlite3` that failed before.

@EwenQuim 

> Why not ! But we should still explicitely return yaml when asked for yaml, disabling validation isn't enough.

I'm not sure what you mean, probably i confused you with the content-type.. it still explicitly returns json, just like before. the change just allows the endpoint to not necessarily end in `.json`.
for example something like `/specs` or `/openapi` would be possible which is more logical if you think that an `accept: application/yaml` returns the content as yaml, regardless of the ending.

since no automatic download is triggered the file extension of the http path is completely irrelevant and even if downloads of the specs are important there is the following header for the download file name: `Content-Disposition: attachment; filename=‘fname.ext’`.

dylanhitt:
I'm cool with this

dylanhitt (approved):

dylanhitt:
Merging. This seems particularly useful if someone overrides our UI handler or what have you. Be it the path doesn't matter, but it would confusing to be getting yaml back when the endpoint is .json

PR #490, opened 2025-04-18
Cast to interface instead of gin.RouterGroup

### LINKED ISSUE #489: Gin adaptor wrongly handles group of wrapper
**To Reproduce**
Code below implements IRouter interface and wraps gin router. 

```go
package main

import (
	"net/http"

	"github.com/gin-gonic/gin"
	"github.com/go-fuego/fuego"
	"github.com/go-fuego/fuego/extra/fuegogin"
	"github.com/go-fuego/fuego/option"
)

type MyRouterWrapper struct {
	router gin.IRouter
}

func (m *MyRouterWrapper) Use(handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.Use(handlers...)
}

func (m *MyRouterWrapper) Handle(method, path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.Handle(method, path, handlers...)
}

func (m *MyRouterWrapper) Any(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.Any(path, handlers...)
}

func (m *MyRouterWrapper) GET(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.GET(path, handlers...)
}

func (m *MyRouterWrapper) POST(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.POST(path, handlers...)
}

func (m *MyRouterWrapper) DELETE(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.DELETE(path, handlers...)
}

func (m *MyRouterWrapper) PATCH(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.PATCH(path, handlers...)
}

func (m *MyRouterWrapper) PUT(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.PUT(path, handlers...)
}

func (m *MyRouterWrapper) OPTIONS(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.OPTIONS(path, handlers...)
}

func (m *MyRouterWrapper) HEAD(path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.HEAD(path, handlers...)
}

func (m *MyRouterWrapper) Match(methods []string, path string, handlers ...gin.HandlerFunc) gin.IRoutes {
	return m.router.Match(methods, path, handlers...)
}

func (m *MyRouterWrapper) StaticFile(path, filepath string) gin.IRoutes {
	return m.router.StaticFile(path, filepath)
}

func (m *MyRouterWrapper) StaticFileFS(path, filepath string, fs http.FileSystem) gin.IRoutes {
	return m.router.StaticFileFS(path, filepath, fs)
}

func (m *MyRouterWrapper) Static(path, filepath string) gin.IRoutes {
	return m.router.Static(path, filepath)
}

func (m *MyRouterWrapper) StaticFS(path string, fs http.FileSystem) gin.IRoutes {
	return m.router.StaticFS(path, fs)
}

func (m *MyRouterWrapper) Group(path string, handlers ...gin.HandlerFunc) *gin.RouterGroup {
	return m.router.Group(path, handlers...)
}

type EchoRequest struct {
	Input string `json:"input" description:"My input string" example:"inputstring"`
}

type EchoResponse struct {
	Output string `json:"output"`
}

func echoGin(ctx *gin.Context) {
	body := EchoRequest{}
	if err := ctx.ShouldBindJSON(&body); err != nil {
		ctx.AbortWithStatus(http.StatusBadRequest)
		return
	}

	ctx.JSON(http.StatusOK, EchoResponse{Output: body.Input})
}

func main() {
	e := fuego.NewEngine()
	r := gin.Default()

	v1 := r.Group("/api/v1")

	fuegogin.PostGin(e, &MyRouterWrapper{router: v1}, "/echo", echoGin)

	e.RegisterOpenAPIRoutes(&fuegogin.OpenAPIHandler{GinEngine: r})

	r.Run(":8080")
}
```

**Expected behavior**
Path for route in api spec `/api/v1/echo`

**Actual behavior**
Path for route in api spec `/echo`

**Framework version:**
v1.18.7

**Go version:**
go 1.24.2

**Additional context**
The problem is here https://github.com/go-fuego/fuego/blob/212de34f06f5820c775835eea02d159deeb4d8dd/extra/fuegogin/adaptor.go#L118

We are trying to cast to `*gin.RouterGroup` but our type is not RouterGroup, it just implements interface. I think there might be better solution

evilenzo:
Closes #489 

Allows IRouter wrapper to implement `BasePath()` method so group wrappers will be handled correctly by openapi


EwenQuim:
Why not, but please add a test to show what was wrong and that it is now fixed :)

evilenzo:
@EwenQuim should be fine now

evilenzo:
@EwenQuim @dylanhitt ping. Need review, please

evilenzo:
If more info needed this is the case from our production. Some services wrap gin router for different reasons. But since there was a cast to gin.RouterGroup we had no way to pass a group path info. This PR fixes it

extra/fuegogin/adaptor.go
```diff
@@ -109,14 +110,18 @@ type ginRouteRegisterer[T, B any] struct {
...
 	// This is because gin groups will prepend the group path to the route path itself.
 	a.ginRouter.Handle(a.route.Method, a.originalPath, a.ginHandler)
 
-	if _, ok := a.ginRouter.(*gin.RouterGroup); ok {
-		a.route.Path = a.ginRouter.(*gin.RouterGroup).BasePath() + a.route.Path
+	if grouped, ok := a.ginRouter.(GroupedRouter); ok && !slices.Contains([]string{"", "/"}, grouped.BasePath()) {
+		a.route.Path = ginToFuegoRoute(grouped.BasePath()) + a.route.Path
 	}
 
 	return a.route
```
ccoVeille, line 127: 🤔

I feel doubtful about the hack you are using to check if grouped.BasePath is empty or /

slices.Contains use generics,it looks like calling superhero for nothing.

Also, the readability is not great.

Please consider this

```suggestion
	if grouped, ok := a.ginRouter.(GroupedRouter); ok {
		basePath := grouped.BasePath()
		switch basePath {
			case "", "/":
				// exclude these ones
			default:
				a.route.Path = ginToFuegoRoute(grouped.BasePath()) + a.route.Path
 	}
 
 	return a.route
```

extra/fuegogin/adaptor_wrapped_router_test.go
```diff
@@ -0,0 +1,119 @@
...
+		assert.NotNil(t, e.OpenAPI.Description().Paths.Value("/api/v1/test/{id}"))
+	})
+
+	t.Run("withGroupParam", func(t *testing.T) {
+		groupV1 := r.Group("/api/v1/:id")
+
+		PostGin(e, &wrappedRouter{router: groupV1}, "/test", dummyHandler)
+
+		assert.NotNil(t, e.OpenAPI.Description().Paths.Value("/api/v1/{id}/test"))
+	})
```
ccoVeille, line 118: Please add tests for group being / or empty.

It would make sense if it was clearer why they are excluded in the code

extra/fuegogin/adaptor.go
```diff
@@ -109,14 +110,18 @@ type ginRouteRegisterer[T, B any] struct {
 	originalPath string
 }
 
+type GroupedRouter interface {
```
dylanhitt, line 113: A comment here explaining when it'd be useful for the caller to use this themselves would be nice.

evilenzo: Added one with example

extra/fuegogin/adaptor.go
evilenzo: Done!

extra/fuegogin/adaptor_wrapped_router_test.go
evilenzo: Nice point, added more tests

extra/fuegogin/adaptor.go
dylanhitt: Excellent thank you

ccoVeille (approved):

dylanhitt (approved):

PR #494, opened 2025-04-21
fix: add detail/status for queryparam errs and detail for path errs

dylanhitt:
Indirectly closes my concerns with #386 and could make #419 unneeded.

```
PathParamNotFoundError
PathParamInvalidTypeError
QueryParamNotFoundError
QueryParamInvalidTypeError
```

All to adhere to the interfaces of `ErrorWithStatus` and `ErrorWithDetail`. In our default errorhandler in the case of the implementation of `ErrorWithStatus` we cast the error to HTTPError anyway. This will also expose the error as Detail ensure we're returning. 

Let me know what you guys think. I don't see the point of copying HTTPError into common Context just to support in both places when we can just implement the interface. See this comment for some context on that: https://github.com/go-fuego/fuego/pull/419#discussion_r1975070317. I did work my way through half this solution before I realized we could just do this.

Let me know what you all think. There isn't really any tests testing the literal error type. I'd like to add it if you guys like this approach, but yeah right now the only real tests around this mainly look like this:

```golang
		paramInt, err = c.QueryParamIntErr("other")
		require.Error(t, err)
		require.Contains(t, err.Error(), "param other=hello is not of type int")
		require.Zero(t, paramInt)
```

### DIFF
--- a/ctx_test.go
+++ b/ctx_test.go
@@ -6,11 +6,15 @@ import (
 	"encoding/xml"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/go-fuego/fuego/internal"
 )
 
 func TestContext_PathParam(t *testing.T) {
@@ -72,12 +76,8 @@ func TestContext_PathParam(t *testing.T) {
 
 	t.Run("reading non-int path param to int sends an error", func(t *testing.T) {
 		s := NewServer()
-		Get(s, "/foo/{id}", func(c ContextNoBody) (ans, error) {
-			id, err := c.PathParamIntErr("id")
-			if err != nil {
-				return ans{}, err
-			}
-			return ans{Ans: fmt.Sprintf("%d", id)}, nil
+		Get(s, "/foo/{id}", func(c ContextNoBody) (any, error) {
+			return c.PathParamIntErr("id")
 		})
 
 		r := httptest.NewRequest("GET", "/foo/abc", nil)
@@ -85,13 +85,14 @@ func TestContext_PathParam(t *testing.T) {
 
 		s.Mux.ServeHTTP(w, r)
 
-		require.Equal(t, 422, w.Code)
+		assert.Equal(t, http.StatusUnprocessableEntity, w.Code)
+		assert.JSONEq(t, `{"title":"Unprocessable Entity","status":422,"detail":"path param id=abc is not of type int"}`, w.Body.String())
 	})
 
-	t.Run("path param invalid", func(t *testing.T) {
+	t.Run("path param not found", func(t *testing.T) {
 		s := NewServer()
-		Get(s, "/foo/", func(c ContextNoBody) (ans, error) {
-			return ans{Ans: c.PathParam("id")}, nil
+		Get(s, "/foo/", func(c ContextNoBody) (any, error) {
+			return c.PathParamIntErr("id")
 		})
 
 		r := httptest.NewRequest("GET", "/foo/", nil)
@@ -99,7 +100,8 @@ func TestContext_PathParam(t *testing.T) {
 
 		s.Mux.ServeHTTP(w, r)
 
-		require.Equal(t, crlf(`{"ans":""}`), w.Body.String())
+		assert.Equal(t, http.StatusUnprocessableEntity, w.Code)
+		assert.JSONEq(t, `{"title":"Unprocessable Entity","status":422,"detail":"path param id not found"}`, w.Body.String())
 	})
 }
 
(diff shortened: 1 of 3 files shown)


internal/common_context.go
```diff
@@ -151,7 +157,13 @@ type QueryParamInvalidTypeError struct {
 }
 
 func (e QueryParamInvalidTypeError) Error() string {
-	return fmt.Errorf("param %s=%s is not of type %s: %w", e.ParamName, e.ParamValue, e.ExpectedType, e.Err).Error()
+	return fmt.Errorf("query param %s=%s is not of type %s: %w", e.ParamName, e.ParamValue, e.ExpectedType, e.Err).Error()
```
ccoVeille, line 160: My comment is more about the code than the idea of your PR.

Also, I feel like you may have copy pasted existing code.

But I would have done this


```suggestion
	return fmt.Sprintf("query param %s=%s is not of type %s: %v", e.ParamName, e.ParamValue, e.ExpectedType, e.Err)
```

internal/common_context.go
```diff
@@ -151,7 +157,13 @@ type QueryParamInvalidTypeError struct {
...
+}
+
+func (e QueryParamInvalidTypeError) StatusCode() int { return 422 }
```
ccoVeille, line 163: Here an everywhere,I would have used existing constants

```suggestion
func (e QueryParamInvalidTypeError) StatusCode() int { return http.StatusUnprocessableEntity }
```

Same for 404 => http.StatusNotFound

internal/common_context.go
dylanhitt: Ope yep I just copy pasta’s everywhere

EwenQuim (approved):

dylanhitt:
Updated!

So reading theourgh HTTP specs 422 seemed to be better for all these types of responses. Also with testing I realized we were giving the caller some too much server information like `strings.Atoi error`.

dylanhitt:
Also I think a big reason why our code coverage dropped so much is because internal is being considered not tested at all. 

http://coveralls.io/builds/73334155/source?filename=internal%2Fcommon_context.go#L151
https://coveralls.io/builds/73334155

ctx_test.go
```diff
@@ -72,34 +76,32 @@ func TestContext_PathParam(t *testing.T) {
...
 		})
 
 		r := httptest.NewRequest("GET", "/foo/abc", nil)
 		w := httptest.NewRecorder()
 
 		s.Mux.ServeHTTP(w, r)
 
-		require.Equal(t, 422, w.Code)
+		assert.Equal(t, http.StatusUnprocessableEntity, w.Code)
+		require.Equal(t, crlf(`{"title":"Unprocessable Entity","status":422,"detail":"path param id=abc is not of type int"}`), w.Body.String())
```
ccoVeille, line 89: Please consider using require.JSONEq

This way you don't have to use crlf I think

ctx_test.go
```diff
@@ -72,34 +76,32 @@ func TestContext_PathParam(t *testing.T) {
...
 		})
 
 		r := httptest.NewRequest("GET", "/foo/", nil)
 		w := httptest.NewRecorder()
 
 		s.Mux.ServeHTTP(w, r)
 
-		require.Equal(t, crlf(`{"ans":""}`), w.Body.String())
+		assert.Equal(t, http.StatusUnprocessableEntity, w.Code)
+		assert.Equal(t, crlf(`{"title":"Unprocessable Entity","status":422,"detail":"path param id not found"}`), w.Body.String())
```
ccoVeille, line 104: Same here JSONEq

ctx_test.go
```diff
@@ -164,11 +174,21 @@ func TestContext_QueryParam(t *testing.T) {
...
+		assert.Equal(t, "param notfound not found", notFoundErr.Error())
+		assert.Equal(t, http.StatusUnprocessableEntity, notFoundErr.StatusCode())
+		assert.Equal(t, "param notfound not found", notFoundErr.DetailMsg())
 
 		paramBool, err = c.QueryParamBoolErr("other")
 		require.Error(t, err)
-		require.False(t, paramBool)
+		require.Zero(t, paramBool)
+		invalidErr := &internal.QueryParamInvalidTypeError{}
+		require.ErrorAs(t, err, invalidErr)
```
ccoVeille, line 188: I would like to suggest writting this


```suggestion
		var invalidErr internal.QueryParamInvalidTypeError
		require.ErrorAs(t, err, &invalidErr)
```

The pointer is only needed for errors.As

internal/common_context.go
```diff
@@ -151,7 +158,13 @@ type QueryParamInvalidTypeError struct {
 }
 
 func (e QueryParamInvalidTypeError) Error() string {
-	return fmt.Errorf("param %s=%s is not of type %s: %w", e.ParamName, e.ParamValue, e.ExpectedType, e.Err).Error()
+	return fmt.Sprintf("%s: %s", e.DetailMsg(), e.Err.Error())
```
ccoVeille, line 161: This may panic if e.Err is nil

This might be better

```suggestion
	return fmt.Sprintf("%s: %s", e.DetailMsg(), e.Err)
```

ctx_test.go
```diff
@@ -164,11 +174,21 @@ func TestContext_QueryParam(t *testing.T) {
...
+		notFoundErr := &internal.QueryParamNotFoundError{}
+		require.ErrorAs(t, err, notFoundErr)
+		assert.Equal(t, "param notfound not found", notFoundErr.Error())
+		assert.Equal(t, http.StatusUnprocessableEntity, notFoundErr.StatusCode())
+		assert.Equal(t, "param notfound not found", notFoundErr.DetailMsg())
 
 		paramBool, err = c.QueryParamBoolErr("other")
 		require.Error(t, err)
-		require.False(t, paramBool)
+		require.Zero(t, paramBool)
```
ccoVeille, line 186: I'm surprised by this change.

Was it suggested by testifylint?

dylanhitt: let me check

ctx_test.go
dylanhitt: Ah yeah this broke the tests with `panic: errors: target must be a non-nil pointer`.

dylanhitt: Also testifylint

dylanhitt: and then if like this

```
var invalidErr internal.QueryParamInvalidTypeError
require.ErrorAs(t, err, &invalidErr)
```

We get:

```
                Error Trace:    /Users/dhitt/github/go-fuego/fuego/ctx_test.go:145
                Error:          Should be in error chain:
                                expected: "query param = is not of type : %!s(<nil>)"
                                in chain: "param notfound not found"
                Test:           TestContext_QueryParam/int
```

dylanhitt: ope nevermind im dumb

dylanhitt: Take a co author for that 😓

ccoVeille (approved):

ctx_test.go
ccoVeille: So it cannot work here with var errWhatever WhateverError + ErrorAs(t, err, &errWhatever) ?

dylanhitt: Na it does that is what I ended up with

```
https://github.com/go-fuego/fuego/pull/494/files#diff-593b58937cba58a2e213b636a0c60bd7671faba0fd51d12a59851ca6d7fec6a3R153
```

I missed the `&`

Issue #496, opened 2025-04-22
Request context is not provided to any slog invocations


jwcodesjs:
Errors returned from request handlers using the default ErrorHandler will log that an error occurred but [no request id is printed](https://github.com/go-fuego/fuego/blob/5f914dc7beb960bef84dede11960744a0a96a3c1/serve.go#L145-L150).  Without the request identifier it's impossible to know which requests, for which user failed with a particular error. If I'm running multiple replicas of my service I'll have no chance of being able to distinguish error events. 

Here's an example of what this looks like

```
9:00AM INF Server running ✅ on http://0.0.0.0:6121 started in="992.542µs"

9:00AM INF incoming request address=[::1]:54563 ip=::1 method=POST path=/foo remote_addr=[::1]:54563 request_id=f5dc0be0-6b6a-4963-a8da-1fe42ae93004 user_agent=PostmanRuntime/7.43.3

9:00AM ERR Error in controller error=invalid request

9:00AM INF request failed address=[::1]:54563 duration=233 ip=::1 method=POST path=/foo remote_addr=[::1]:54563 request_id=f5dc0be0-6b6a-4963-a8da-1fe42ae93004 status_code=500 user_agent=PostmanRuntime/7.43.3
```

I’m currently using zerolog instead of slog. There's multiple ways I can extract data from the context object for logging but any logs printed by fuego I won't be able to properly trace since the request context is not provided. 

```go
type TracingHook struct{}  
  
func (h TracingHook) Run(e *zerolog.Event, level zerolog.Level, msg string) {  
    ctx := e.GetCtx()  // will be nil
    spanId := getSpanIdFromContext(ctx)
    e.Str("span-id", spanId)  
}
```

### Proposed Solutions

1. **Pass context to the error handler:**
    
    Updating server.Flow to pass the request context to ErrorHandler would let me continue on to log the error with whatever information I want.
    
2. **Switch to using context-aware slog functions:**
    
    Replacing standard calls to slog.Info(...) with slog.InfoContext(ctx, ...).  Quick search shows [12 files](https://github.com/search?q=repo%3Ago-fuego%2Ffuego+%22slog.%22+NOT+path%3Aexamples+NOT+path%3A*_test.go&type=code) — but probably fewer actually have a request context in scope.

dylanhitt:
I assume you're using something like [slog-zerolog](https://github.com/samber/slog-zerolog) as a handler and passing that to `func WithLogHandler(handler slog.Handler)`?

[...]

PR #498, opened 2025-04-23
BREAKING: use slog.***Context where applicable 

dylanhitt:
Closes: #496 

BREAKING CHANGE: `ErrorHandler` signature is updated to also include context. 

This change allows callers too pass request scoped values to our log functions. The great thing about slog is that it allows you to bring your own logger. Slog even suggest to use [context functions](https://pkg.go.dev/log/slog#hdr-Contexts) when possible for this very reason. 

> It is recommended to pass a context to an output method if one is available.

Thanks

### DIFF
--- a/serve_test.go
+++ b/serve_test.go
@@ -637,7 +637,7 @@ func TestFlow(t *testing.T) {
 		assert.Equal(t, crlf(`{"ans":"Hello World"}`), w.Body.String())
 	})
 	t.Run("with nil return in ErrorHandler", func(t *testing.T) {
-		e := NewEngine(WithErrorHandler(func(err error) error { return nil }))
+		e := NewEngine(WithErrorHandler(func(_ context.Context, err error) error { return nil }))
 		w := httptest.NewRecorder()
 		ctx := newTestCtx(w, httptest.NewRequest("GET", "/", nil))
 		Flow(e, ctx, testControllerWithError)
--- a/serve_test.go
+++ b/serve_test.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"math/big"
 	"net"
 	"net/http"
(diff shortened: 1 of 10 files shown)


serve_test.go
```diff
@@ -637,7 +637,7 @@ func TestFlow(t *testing.T) {
 		assert.Equal(t, crlf(`{"ans":"Hello World"}`), w.Body.String())
 	})
 	t.Run("with nil return in ErrorHandler", func(t *testing.T) {
-		e := NewEngine(WithErrorHandler(func(err error) error { return nil }))
+		e := NewEngine(WithErrorHandler(func(_ context.Context, err error) error { return nil }))
 		w := httptest.NewRecorder()
 		ctx := newTestCtx(w, httptest.NewRequest("GET", "/", nil))
```
ccoVeille, line 642: I would expect to see some tests to check the context received by the ErrorHandler is the one from the request

Either here or in engine_test.go, or both

dylanhitt: Could you try explaining 

> I would expect to see some tests to check the context received by the ErrorHandler is the one from the request

In a different way? 

Like ensuring it's from `http.Request`??

dylanhitt: 🤔 I guess. I could call `HTTPHandler` and slogassert the whole chain

ccoVeille: I was talking about something like this

```go
		var receivedCtx context.Context
		e := NewEngine(WithErrorHandler(func(ctx context.Context, err error) error {
			receivedCtx = ctx
			return nil
		}))
 		w := httptest.NewRecorder()
 		ctx := newTestCtx(w, httptest.NewRequest("GET", "/", nil))
 		Flow(e, ctx, testControllerWithError)
		assert.Equal(t, ctx, receivedCtx)
```

dylanhitt: Ah, cool. Nice idea. 

I think I'll do both as it's pretty easy to add a log in Flow and not have it be `***Context`

ccoVeille (approved):

EwenQuim (approved):
This is a good idea

dylanhitt:
Updated with the logging test. I'll add cases as I find them if that's alright.

Thank you

PR #500, opened 2025-04-23
feat: Parse tags for arrays/slices fields in structs

### LINKED ISSUE #499: Tags are not properly reflected for structure arrays, slices, ptr, etc.
**To Reproduce**
The following model:
```go
type BannersResult struct {
	Banner *Banner `json:"banner,omitempty"`
}

type Banner struct {
	Description       string `json:"description" description:"Description"`
}
```
Will not have tags properly reflected

Image

**Expected behavior**
Tags should be reflected

latest

1.24

**Additional context**

serchemach:
closes #499

### DIFF
--- a/examples/petstore/lib/testdata/doc/openapi.golden.json
+++ b/examples/petstore/lib/testdata/doc/openapi.golden.json
@@ -121,12 +121,16 @@
 							"properties": {
 								"more": {
 									"additionalProperties": {},
+									"description": "Additional information about the error",
+									"nullable": true,
 									"type": "object"
 								},
 								"name": {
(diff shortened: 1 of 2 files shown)


EwenQuim:
Very good idea. Please add a test so we can see the change in the behaviour and make sure it does not break in the future!

EwenQuim (approved):
I've updated the PR by updating the golden file. Indeed, it now goes through lists to parse struct tags, as ErrorItem was never parsed before and now it is.

Thanks for the PR!

_Next time please update the tests or add some directly, it helps us maintainers save some time 😉_

PR #501, opened 2025-04-23
feat: Array of parameters deserialization

EwenQuim:


EwenQuim:
@dylanhitt 🎁

dylanhitt:
Will look at this and the other tomorrow.

ctx.go
```diff
@@ -376,54 +411,47 @@ func (c *netHttpContext[B, P]) Params() (P, error) {
...
-			}
-		} else {
-			tag = field.Tag.Get("header")
-			if tag != "" {
-				paramValue = c.Header(tag)
-				if paramValue == "" {
+		// Process query parameters
+		if tag := field.Tag.Get("query"); tag != "" {
+			// Handle slice/array types
+			if field.Type.Kind() == reflect.Slice || field.Type.Kind() == reflect.Array {
```
dylanhitt, line 419: Nit pick: I tend to like switch cases when I see this

ctx.go
```diff
@@ -376,54 +411,47 @@ func (c *netHttpContext[B, P]) Params() (P, error) {
...
+				if err != nil {
+					return *p, err
+				}
 			}
-			fieldValue.SetUint(uintValue)
-		case reflect.Float32, reflect.Float64:
-			floatValue, err := strconv.ParseFloat(paramValue, 64)
-			if err != nil {
-				return *p, fmt.Errorf("cannot convert %s to float64: %w", paramValue, err)
+		} else if tag := field.Tag.Get("header"); tag != "" {
```
dylanhitt, line 445: Same here. But feel free disregard it's just a style thing. I'm find myself in the perpetual war against else statement

dylanhitt (approved):

ctx.go
```diff
@@ -365,6 +365,41 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+		}
+		value.SetBool(boolValue)
+	default:
+		return fmt.Errorf("unsupported type %s", kind)
+	}
+	return nil
+}
+
 func (c *netHttpContext[B, P]) Params() (P, error) {
 	p := new(P)
```
ccoVeille, line 404: Nit: This is old code, but why creating a pointer, when method return the non-pointer version

ctx.go
```diff
@@ -376,54 +411,47 @@ func (c *netHttpContext[B, P]) Params() (P, error) {
...
-			boolValue, err := strconv.ParseBool(paramValue)
+			err := setParamValue(fieldValue, paramValue, field.Type.Kind())
 			if err != nil {
-				return *p, fmt.Errorf("cannot convert %s to bool: %w", paramValue, err)
+				return *p, err
 			}
-			fieldValue.SetBool(boolValue)
-		default:
-			return *p, fmt.Errorf("unsupported type %s", field.Type.Kind())
 		}
```
ccoVeille, line 455: Nit: I would move all this if into its own method.

Or at least, I would return early when there is no tag.

The code is pretty complicated with a lot of branching

ctx.go
ccoVeille: I approve this message 😁

ctx.go
ccoVeille: I approve this message 😁

ctx_test.go
```diff
@@ -690,4 +690,147 @@ func TestNetHttpContext_Params(t *testing.T) {
...
+			Int64s  []int64  `query:"int64s"`
+			Uints   []uint   `query:"uints"`
+			Uint8s  []uint8  `query:"uint8s"`
+			Uint16s []uint16 `query:"uint16s"`
+			Uint32s []uint32 `query:"uint32s"`
+			Uint64s []uint64 `query:"uint64s"`
+		}
+
+		url := "http://example.com/foo?" +
+			"int8s=1&int8s=2&" +
```
ccoVeille, line 736: Please also tests how the code reacts to overflow, like settings 1024 into an int8 that cannot go further than 128

ctx.go
```diff
@@ -365,6 +365,41 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
 	return body, err
 }
 
+// setParamValue sets a value to a reflect.Value based on its kind
+func setParamValue(value reflect.Value, paramValue string, kind reflect.Kind) error {
+	switch kind {
+	case reflect.String:
+		value.SetString(paramValue)
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		intValue, err := strconv.Atoi(paramValue)
```
ccoVeille, line 374: Please use  strconv.ParseInt

You also need to play with the base here

```suggestion
		intValue, err := strconv.ParseInt(paramValue, 10, bitSize(kind))
```

See https://github.com/go-fuego/fuego/pull/501#discussion_r2057705693

ctx.go
```diff
@@ -365,6 +365,41 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+		if err != nil {
+			return fmt.Errorf("cannot convert %s to int: %w", paramValue, err)
+		}
+		value.SetInt(int64(intValue))
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		uintValue, err := strconv.ParseUint(paramValue, 10, 64)
+		if err != nil {
+			return fmt.Errorf("cannot convert %s to uint: %w", paramValue, err)
+		}
+		value.SetUint(uintValue)
```
ccoVeille, line 384: Here you are facing a problem with overflow

You should do this.

```suggestion
	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
		uintValue, err := strconv.ParseUint(paramValue, 10, bitSize(kind))
		if err != nil {
			return fmt.Errorf("cannot convert %s to uint: %w", paramValue, err)
		}
		value.SetUint(uintValue)
```

With a method that returns you the bit size

```go
func bitSize(kind) {
switch kind {
   case reflect.Uint8, reflect.Int8:
     return 8
   case reflect.Uint16, reflect.Int16:
     return 16
   case reflect.Uint32, reflect.Int32, reflect.Float32:
     return 32
   case reflect.Uint, reflect.Int:
     return strconv.IntSize()
}
return 64
}
```

I wrote my own lib about this issue

https://github.com/ccoVeille/go-safecast

I don't expect you to use it, it's just the fact it's not easy to deal with this.

ctx.go
```diff
@@ -365,6 +365,41 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+	case reflect.Float32, reflect.Float64:
+		floatValue, err := strconv.ParseFloat(paramValue, 64)
```
ccoVeille, line 386: Same here, you have to check and use bitSize


```suggestion
		floatValue, err := strconv.ParseFloat(paramValue, bitSize(kind))
```

ccoVeille (changes requested):

ctx.go
EwenQuim: Fixed here

ctx.go
EwenQuim: I don't like switch for non-scalar comparisons 😕

ctx.go
```diff
@@ -365,6 +365,41 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+	case reflect.Bool:
+		boolValue, err := strconv.ParseBool(paramValue)
+		if err != nil {
+			return fmt.Errorf("cannot convert %s to bool: %w", paramValue, err)
+		}
+		value.SetBool(boolValue)
+	default:
+		return fmt.Errorf("unsupported type %s", kind)
+	}
+	return nil
```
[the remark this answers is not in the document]
EwenQuim: No it isn't, there are no `return` clauses in the switch cases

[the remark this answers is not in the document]
ccoVeille: Oh indeed !

ctx.go
```diff
@@ -365,6 +365,55 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+func setParamValue(value reflect.Value, paramValue string, kind reflect.Kind) error {
+	switch kind {
+	case reflect.String:
+		value.SetString(paramValue)
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		intValue, err := strconv.ParseInt(paramValue, 10, bitSize(kind))
+		if err != nil {
+			return fmt.Errorf("cannot convert %s to int: %w", paramValue, err)
+		}
+		value.SetInt(intValue)
```
ccoVeille, line 392: Just found this

https://pkg.go.dev/reflect#Value.OverflowInt

I would like to suggest this

```suggestion
	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
		intValue, err := strconv.ParseInt(paramValue, 10, 64)
		if err != nil {
			return fmt.Errorf("cannot convert %s to %s: %w", paramValue, kind, err)
		}
		if value.OverflowInt(intValue) {
			return fmt.Errorf("%s overflows %s", paramValue, kind)
		}
		value.SetInt(intValue)
```

I didn't test

ctx.go
```diff
@@ -365,6 +365,55 @@ func (c *netHttpContext[B, P]) Body() (B, error) {
...
+		if err != nil {
+			return fmt.Errorf("cannot convert %s to uint: %w", paramValue, err)
+		}
+		value.SetUint(uintValue)
+	case reflect.Float32, reflect.Float64:
+		floatValue, err := strconv.ParseFloat(paramValue, bitSize(kind))
+		if err != nil {
+			return fmt.Errorf("cannot convert %s to float64: %w", paramValue, err)
+		}
+		value.SetFloat(floatValue)
```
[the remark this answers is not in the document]
EwenQuim: OK no problem ;)

ctx.go
EwenQuim: Interesting!

PR #506, opened 2025-04-30
ci(release): integrate release please

totallyzen:
# changes

- integrate `release-please` with some personal best practices:
  - use an explicit manifest file -> much clearer of current expected version, etc
  - locate the config files in the `.github` folder to avoid clutter in the repo root
- integrate `.goreleaser.yml` but maybe this is overkill since the changelog generation is more part of `release-please` now

# notes

## goreleaser

Started integrating `.goreleaser.yml` but I fail to see what current purpose it could hold. Seems like `release-please` is enough for what you need.
In our private processes, `goreleaser` is more used as a cross-platform compilation tool and I'm not sure you actually need this in `fuego` since it's a library.
Happy to discuss more details and adjust any files needed.

## pre-releases

I've not really used `release-please` for pre-releases yet and you might want to enable the use, might have to play with it in a separate repo.

If not, you can likely still make tags for pre-releases (or manage these releases manually). May need some research.

## enforcing conventional commits

I've injected one that may not be the most popular, but it's written in go and suited me very well already.
Feel free to adjust/fork or simply star the repo I've linked. Simple but effective (also it's written in Go so staying within the ecosystem)


.github/.release-please-manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+    ".": "0.18.0"
```
totallyzen, line 2: the current released version

.github/release-please-config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "bootstrap-sha": "802052722609cd05316a38ae8bfde33a5dd0b6df",
```
totallyzen, line 2: should be removed after the first successful release! it ensures `release-please` doesn't scan too many commits. this commit matches your current released tag.

.github/release-please-config.json
```diff
@@ -0,0 +1,10 @@
...
+    "packages": {
+        ".": {
+            "release-type": "go",
+            "include-v-in-tag": true
+        }
+    },
+    "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json"
```
totallyzen, line 9: there's a lot of config available in this file, use your favourite IDE to explore!
It can include and exclude paths, as well as inject version information under "extra files" (or something like that, use the schema completion to explore!)

.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,19 @@
...
+      - opened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: read
+
+jobs:
+  validate:
+    name: validate-pull-request-title
```
totallyzen, line 15: if `release-please` is integrated, you might want to make this `required` to ensure every PR has a correct format. you can also narrow down commit types as well as scopes. I opted not to add any so far, but it's definitely possible.

.github/workflows/release-please.yaml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      release_created: ${{ steps.track-release.outputs.release_created }}
+    steps:
+      - uses:
+          googleapis/release-please-action@a02a34c4d625f9be7cb89156071d8567266a2445 # v4
+        id: track-release
+        with:
+          manifest-file: .github/.release-please-manifest.json
+          config-file: .github/release-please-config.json
```
totallyzen, line 18: unlike the starter docs, I tend to separate the build/release step into a separate job - it lines up much clearer in the GHA steps/logs -> the `goreleaser` step demonstrates the follow-up usage

.github/workflows/release-please.yaml
```diff
@@ -0,0 +1,47 @@
...
+  goreleaser:
+    needs:
+      - track-release
+    if: ${{ needs.track-release.outputs.release_created }}
+
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+      packages: write
```
totallyzen, line 28: internally in our company we mostly use `goreleaser` for building cross-platform binaries - if you need to set up some stuff, please look at the `.goreleaser.yaml` file

.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
+# Make sure to check the documentation at https://goreleaser.com
+project_name: fuego
```
totallyzen, line 3: as the name indicates, this is vanilla - I was surprised that you mentioned goreleaser and you didn't already have a file for it.

.github/release-please-config.json
totallyzen: important!

there is a known bug with bootstrapping on "github style" change/release notes -> can be set no problem after bootstrap tho. may also have been fixed since the last time I looked. would still recommend bootstrapping in google-style and then go back to github style changelogs.

.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
...
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+  - <<: *main-build
+    main: ./examples/acme-tls
```
totallyzen, line 21: this many not be something you actually want, but demonstrates the cross-platform compilation. 

also, yaml anchors for the win!

totallyzen (commented):
some author annotations to explain what is going on

totallyzen:
one important note for `release-please` that I forgot.

When you edit the body of the resulting release PR, it will also end up in the release notes. The changelog does however need to be edited manually (in the same PR, ofc).

So as a consequence it's important to groom the bootstrap PR so it's not missing a change you'd like to give credit for.

.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-22.04
```
ccoVeille, line 16: Why not this?

```suggestion
    runs-on: ubuntu-latest
```

I think it's a copy paste. I don't think this would require such old LTS release

.github/workflows/release-please.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          fetch-tags: true
+
+      - name: Set up Go
+        uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@9c156ee8a17a598857849441385a2041ef570552 # v6
```
ccoVeille, line 40: I feel like all these come from another repository. But I assume it's from a repository that uses hardened version, aka pinned version.

fuego doesn't pin versions. Maybe it should, but let's keep it out of this PR considerations.

I'm unsure how dependabot will handle them while there are no hardened action.

I might be wrong.

I would use the regular way here


```suggestion
        uses: actions/checkout@v4
        with:
          fetch-tags: true

      - name: Set up Go
        uses: actions/setup-go@v5

      - name: Run GoReleaser
        uses: goreleaser/goreleaser-action@v6
```

This comment is valid for all workflow files that are updated by this PR

.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
+# Make sure to check the documentation at https://goreleaser.com
```
ccoVeille, line 2: I thought you said we have no need for goreleaser?

I'm surprised to see it anyway

totallyzen: Well, that was my shallow understanding of your project, I left it here in case you have some insight you want to offer. :slightly_smiling_face: 

I added it to illustrate a point on what I find it useful for -> cross-compilation. However, the more I read the more I see `goreleaser` as it's own complete tool. There is definitely an exposure bias going on for me towards `release-please` but I think you can transition to `goreleaser` later if you so choose.

I'll remove it for now since I have better confidence in knowing what I am doing with `release-please`

.github/workflows/release-please.yaml
totallyzen: Your project, your rules!
I undertsand the pinning as a _best practice_ in GHA, but I'm fine with deferring that.

When it comes to dependency management, if dependabot can't handle it, Renovate has a free tier that definitely knows how to handle the version pinning and it's extremely simple to start going with Renovate.

totallyzen: Absolutely! :slightly_smiling_face: 

Sorry if it seemed resentful. Not intended at all. :pray:

.goreleaser.yaml
ccoVeille: Here I had no real point of view on this.

I was just surprised to find, while I was pretty sure you said, we might not need it.

Maybe, there is a need for it. That was somehow expressed by Ewen at first.

I don't mind enabling go-releaser with this PR.

I was trying to understand the context and what you said

.github/.release-please-manifest.json
dylanhitt: So we do have a `0.18.8` tag. Should this be that or is releasee please aware of git tags in this case?

dylanhitt (commented):
So one that I'm missing is how do our submodules get tagged? I think when I looked at initially that was originally why we thought of both release-please and GoReleaser. https://goreleaser.com/customization/monorepo/#category-1

Unless there is a way to do it with release-please that isn't showing up through my searching?

totallyzen:
> Unless there is a way to do it with release-please that isn't showing up through my searching?

I think you're right, `release-please` is oriented around several different `semver` tags and not duplicating the same tag for different variants.

I'll have a look at `goreleaser` docs on how to make this happen - I think we'll indeed need both of them in the end.
Gimmie a few days to read the docs thoroughly.

Can you also give me the list of submodules? Sorry for the shallow understanding here, but it's simpler to ask for the list. I don't have the same amount of time as I used to a few years back (I have a son to look after most evenings)

.github/.release-please-manifest.json
totallyzen: :wave: ah, then this needs to be the latest tag indeed with that underlying SHA1. release please will track releases combined with tags. I'm not sure what it does if either is missing.

.github/workflows/release-please.yaml
totallyzen: I didn't see this last message on pinning. I'm fine either way but maybe make that pinning in a different PR.

EwenQuim:
> Can you also give me the list of submodules?

The main ones are 

- fuego
- fuego/option
- fuego/param
- fuego/extra/fuegogin
- fuego/extra/fuegoecho

totallyzen:
Hi, it's been a quite few weeks.

Terribly sorry! I've over-promised on this.

What remains to do:
- update `.github/release-please-manifest.json` to use the correct `bootstrap_sha` and versioning according to your latest tag
- add a job in `.github/release-please.yml` similar to what I've removed [here](https://github.com/go-fuego/fuego/pull/506/commits/61fb16154eefcdd24f2550b1b429510b12fc62a7) 
  - you simply `checkout` 
  - install `mikefarah/yq` 
  - query `/github/release-please-manifest.json` and shovel the version into an env variable 
  - make lots of tags with your sub-packages (you share the version among them so it's fine) 
  - push tags (GH actions will not trigger more workflows, ofc)

dylanhitt:
Closing to do staleness

PR #511, opened 2025-05-02
use go-version-file

dylanhitt:
Was made aware of this today


ccoVeille (approved):
Yes, it's a good option when only one version is tested

PR #523, opened 2025-05-22
feat: engine option to set the global response content type

dylanhitt:
Engine option for setting response content-type. This is useful:

1. In the case where you do not want to have support xml as some struct patterns will break xml parsing. 
2. The use brings their own deserializer and only wants to ever return a specific return type.


engine_test.go
```diff
@@ -103,3 +103,35 @@ func TestWithRequestContentType(t *testing.T) {
 		require.False(t, ok)
 	})
 }
+
+func TestWithResponseContentType(t *testing.T) {
+	t.Run("base", func(t *testing.T) {
+		e := NewEngine()
+		require.NotNil(t, e.responseContentTypes)
+		assert.Equal(t, []string{"application/json", "application/xml"}, e.responseContentTypes)
```
ccoVeille, line 111: I'm unsure you need the require line


```suggestion
		assert.NotNil(t, e.responseContentTypes)
		assert.Equal(t, []string{"application/json", "application/xml"}, e.responseContentTypes)
```

Or maybe this,

```suggestion
		assert.NotEmpty(t, e.responseContentTypes)
		assert.Equal(t, []string{"application/json", "application/xml"}, e.responseContentTypes)
```

But I don't think you need it


```suggestion
		assert.Equal(t, []string{"application/json", "application/xml"}, e.responseContentTypes)
```

engine_test.go
```diff
@@ -103,3 +103,35 @@ func TestWithRequestContentType(t *testing.T) {
...
+	t.Run("ensure applied to route", func(t *testing.T) {
+		s := NewServer(WithEngineOptions(
+			WithResponseContentType("application/yml")),
+		)
+		route := Post(s, "/test", dummyController)
+
+		response := route.Operation.Responses.Map()["200"]
+		require.NotNil(t, response)
+		content := response.Value.Content
+		require.NotNil(t, content["application/yml"])
```
ccoVeille, line 129: A contains would provide a clearer error message

```suggestion
		require.Contains(t, content, "application/yml")
```

engine_test.go
```diff
@@ -103,3 +103,35 @@ func TestWithRequestContentType(t *testing.T) {
...
+		assert.Equal(t, "#/components/schemas/Resp", content["application/yml"].Schema.Ref)
+
+		require.Nil(t, content["application/xml"])
+
+		_, ok := s.OpenAPI.Description().Components.RequestBodies["ReqBody"]
+		require.False(t, ok)
```
ccoVeille, line 135:
```suggestion
		require.NotContain(t, s.OpenAPI.Description().Components.RequestBodies, "ReqBody")
```

engine_test.go
```diff
@@ -103,3 +103,35 @@ func TestWithRequestContentType(t *testing.T) {
...
+		)
+		route := Post(s, "/test", dummyController)
+
+		response := route.Operation.Responses.Map()["200"]
+		require.NotNil(t, response)
+		content := response.Value.Content
+		require.NotNil(t, content["application/yml"])
+		assert.Equal(t, "#/components/schemas/Resp", content["application/yml"].Schema.Ref)
+
+		require.Nil(t, content["application/xml"])
```
ccoVeille, line 132: Same here with NotContains

engine.go
```diff
@@ -26,8 +26,9 @@ import (
 // Options all begin with `With`.
 func NewEngine(options ...func(*Engine)) *Engine {
 	e := &Engine{
-		OpenAPI:      NewOpenAPI(),
-		ErrorHandler: ErrorHandler,
+		OpenAPI:              NewOpenAPI(),
+		ErrorHandler:         ErrorHandler,
+		responseContentTypes: []string{"application/json", "application/xml"},
```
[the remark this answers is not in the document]
dylanhitt: Eh, maybe

ccoVeille (approved):

EwenQuim (approved):
Yes we needed this.

Also, I noticed you don't like merge commits, I disabled it on the repo to only allow linear history (squash or rebase)

PR #524, opened 2025-05-22
feat: allow custom content-type serdes for each endpoint

### LINKED ISSUE #516: Pluggable custom content-type deserialization/serialization
The Fuego design that separates deserialization of request body and serialization of response body so that handlers only need to receive and return structs (or specific types) is very powerful. However, the current implementation is not very extensible: there are default behaviors for a handful of common content-types. We can replace `Serialize()` at the server level to implement custom logic on the way out. But the deserialization logic is encapsulated in `ctx.go`'s `body()` method [(link)](https://github.com/go-fuego/fuego/blob/981d4d4f5cc00adfc48fe55d9f1af1f09be89587/ctx.go#L508)

We have a couple of other content-types we need to both accept and return. However, we also want to provide the (new) option to interact with the data as JSON. The meaningful content in each representation is the same. In other words, content negotiation with additional content-types compared to what is baked into Fuego today. Ideally, we would implement de/serialization logic _outside_ of our handler method, following the same design principles that allows Fuego to go from JSON->XML or any combination thereof.

The ability to register custom de/serialization methods for specific content-types (including overriding the default handlers). In some cases, this would be nice to have on an endpoint-by-endpoint basis, but this is not strictly necessary. One set of handlers we can configure at the Fuego Server level would be great for extensibility.

Example interface:

```go
fuego.POST(s, "/foo", resources.FooHandler,
	option.Summary("Submit a foo"),
	option.DefaultStatusCode(200),
        // override text de/serialization to map the text representation in and out of our struct
	option.WithContentTypeSerde("text/plain", myCustomTextSerde),
        // handle other specific mime type that may be binary (or anything), but map it in/out of our struct
        option.WithContentTypeSerde("application/vnd.foo", myFooSerde),
)
```

where `mySerde` variables are structs with `Serialize()` and `Deserialize()` methods.

Then my controller/handler method can be strongly typed with my struct for request body and response type annotations, but not care about how the data gets in or out of that struct.

```go
func (h *Resources) AddFoo(c fuego.ContextWithBody[StronglyTypedFooStruct]) (StronglyTypedFooResponse, error) {
    // do the business logic...

    // ...but not the serialization logic
    return StronglyTypedFooResponse{}
}
```

We are currently working around this limitation by using either `string` or `[]byte` in the annotation for our handler context, but this limits us to specific `content-type` values & behavior set in Fuego, and we end up adding all the switch statements for `content-type` and `accept` more-or-less inside of our handlers. Additionally, we lose the well-defined struct in the OpenAPI spec.

**Additional context**
If we could agree upon a design, I could contribute a PR.

P.S. thanks for the great library!

hamilton-earthscope:
This PR adds support for a new Fuego endpoint `Option` to configure custom serialization and deserialization logic for specific content types.

Closes #516

If this general idea is agreeable, we could additionally refactor the current default content-type behaviors (`text/plain`, `application/json`, `application/xml`, etc.) from switch statements on `content-type` and `accept` headers to structs adhering to the `Serde` interface so that there is only one path to serialization/deserialization. I.e. `SendText()` ([link](https://github.com/go-fuego/fuego/blob/08392cc3224cbf58b305b11628845ae1106a28ee/serialization.go#L96)) and `readString()` ([link](https://github.com/go-fuego/fuego/blob/08392cc3224cbf58b305b11628845ae1106a28ee/ctx.go#L520)) become a `StringSerde{}` registered for `text/plain`.

Other notes:
- All of the `Read*()` methods ([link](https://github.com/go-fuego/fuego/blob/08392cc3224cbf58b305b11628845ae1106a28ee/deserialization.go#L42)) take `(ctx context.Context, input io.Reader, options readOptions)` args. Perhaps `Deserialize()` should adhere to this.
- All of the `Send*()` methods ([link](https://github.com/go-fuego/fuego/blob/08392cc3224cbf58b305b11628845ae1106a28ee/serialization.go#L157)) take `(w http.ResponseWriter, r *http.Request, ans any)` args. Perhaps `Serialize()` should adhere to this.
- Should this be configurable at the server level in addition to the endpoint level?
- Docs need updated

### DIFF
--- a/option/option.go
+++ b/option/option.go
@@ -201,3 +201,6 @@ var DefaultStatusCode = fuego.OptionDefaultStatusCode
 // StripTrailingSlash removes the trailing slash from the route.
 // By default, the trailing slash is kept, so becauseful when registering route like "/" within a group.
 var StripTrailingSlash = fuego.OptionStripTrailingSlash
+
+// WithContentTypeSerde sets a custom serializer and deserializer for a content type.
+var WithContentTypeSerde = fuego.OptionWithContentTypeSerde
--- a/option_test.go
+++ b/option_test.go
@@ -2,9 +2,12 @@ package fuego_test
 
 import (
 	"errors"
+	"fmt"
 	"log/slog"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/getkin/kin-openapi/openapi3"
@@ -1071,5 +1074,59 @@ func TestOptionStripTrailingSlash(t *testing.T) {
 		s := fuego.NewServer()
 		route := fuego.Get(s, "/test/", helloWorld, fuego.OptionStripTrailingSlash())
 		require.Equal(t, "/test", route.Path)
+	})
+}
+
+// myCustomType has custom serialization and deserialization logic
+type myCustomType struct {
+	A string
+	B int
+}
+
+// Serialize custom serialization logic
+func (s myCustomType) Serialize(v any) ([]byte, error) {
+	v2, ok := v.(myCustomType)
+	if !ok {
+		return nil, errors.New("invalid type")
+	}
+	return []byte(fmt.Sprintf("%s:%d", v2.A, v2.B)), nil
+}
+
+// Deserialize custom deserialization logic
+func (s myCustomType) Deserialize(data []byte) (any, error) {
+	var v myCustomType
+	var err error
+	parts := strings.Split(string(data), ":")
+	if len(parts) != 2 {
+		return myCustomType{}, errors.New("invalid data")
+	}
+	v.A = parts[0]
+	v.B, err = strconv.Atoi(parts[1])
+	return v, err
+}
+
+func TestWithCustomSerde(t *testing.T) {
+	t.Run("Custom serde is used", func(t *testing.T) {
+		s := fuego.NewServer()
+		route := fuego.Post(s, "/times2",
+			func(c fuego.ContextWithBody[myCustomType]) (myCustomType, error) {
+				body, err := c.Body()
+				if err != nil {
+					return myCustomType{}, err
+				}
+				return myCustomType{A: body.A + " times 2", B: body.B * 2}, nil
+			},
+			option.WithContentTypeSerde("application/vnd.foo", myCustomType{}),
+		)
+		require.Equal(t, "/times2", route.Path)
+
+		r := httptest.NewRequest("POST", "/times2", strings.NewReader("hello:2"))
+		r.Header.Set("Content-Type", "application/vnd.foo")
+		w := httptest.NewRecorder()
+
+		s.Mux.ServeHTTP(w, r)
+
+		require.Equal(t, 200, w.Code)
+		require.Equal(t, "hello times 2:4", w.Body.String())
 	})
 }
--- a/option_test.go
+++ b/option_test.go
@@ -6,7 +6,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/http/httptest"
-	"strconv"
+	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


option/option.go
```diff
@@ -201,3 +201,6 @@ var DefaultStatusCode = fuego.OptionDefaultStatusCode
 // StripTrailingSlash removes the trailing slash from the route.
 // By default, the trailing slash is kept, so becauseful when registering route like "/" within a group.
 var StripTrailingSlash = fuego.OptionStripTrailingSlash
+
+// WithContentTypeSerde sets a custom serializer and deserializer for a content type.
+var WithContentTypeSerde = fuego.OptionWithContentTypeSerde
```
dylanhitt, line 206: Will need a statement here that this only applies to the `fuego.Server`

dylanhitt (changes requested):
Just an initial look. Mind cleaning up the linting errors?

Thank you!

option_test.go
```diff
@@ -1073,3 +1076,88 @@ func TestOptionStripTrailingSlash(t *testing.T) {
...
+
+		// In kv; out kv
+		r := httptest.NewRequest("POST", "/keyvalue", strings.NewReader("key1=hello;key2=2"))
+		r.Header.Set("Accept", "application/vnd.keyvalue")
+		r.Header.Set("Content-Type", "application/vnd.keyvalue")
+		w := httptest.NewRecorder()
+
+		s.Mux.ServeHTTP(w, r)
+
+		require.Equal(t, 200, w.Code)
```
dylanhitt, line 1138:
```suggestion
		require.Equal(t, http.StatusOK, w.Code)
```

option_test.go
```diff
@@ -1073,3 +1076,88 @@ func TestOptionStripTrailingSlash(t *testing.T) {
...
+	}
+
+	t.Run("Custom serde is used", func(t *testing.T) {
+		s := fuego.NewServer()
+		fuego.Post(s, "/keyvalue", controller,
+			option.WithContentTypeSerde("application/vnd.keyvalue", kvSerde{}),
+		)
+
+		// In kv; out kv
+		r := httptest.NewRequest("POST", "/keyvalue", strings.NewReader("key1=hello;key2=2"))
```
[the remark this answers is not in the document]
dylanhitt: Same in the few other places here. 

Just a nitpick feel free resolve.

option_test.go
```diff
@@ -1073,3 +1076,88 @@ func TestOptionStripTrailingSlash(t *testing.T) {
...
+		r.Header.Set("Accept", "application/json")
+		r.Header.Set("Content-Type", "application/vnd.keyvalue")
+		w = httptest.NewRecorder()
+
+		s.Mux.ServeHTTP(w, r)
+
+		require.Equal(t, 200, w.Code)
+		require.JSONEq(t, `{"foo":"bar","key1":"hello","key2":"2"}`, w.Body.String())
+
+		// In json; out kv
```
dylanhitt, line 1152: Instead of separating with comments lets use table driven tests. Your comments for the tittle would work great there

dylanhitt (commented):
This is one of those MRs that make me really appreciate the open source community. 😄 

Thank you

dylanhitt:
One thought on this: we should get these the content-type declared generating in the spec. Looking at the code as well I think it should final. As in whatever is provided here should be the only content-type(s) that appear. I wouldn't suggest placing it in this PR as once #523 is merged the implementation should be trivial.

option/option.go
dylanhitt: Bump on the comment. 

maybe:

```
// WithContentTypeSerde sets a custom serializer and deserializer for a content type.
// This option is currently only applicable the [fuego.Server]. Other adaptors are not affected by this option. 
```

dylanhitt (changes requested):

option/option.go
dylanhitt: Yes you are correct. That's not what I mean though. Fuego has it's own `net/http` based [Server]. Fuego also has a concept called `adaptors` that use the `fuego.Engine` to create an openapi spec. see: https://github.com/go-fuego/fuego/tree/main/extra/fuegoecho and https://github.com/go-fuego/fuego/tree/main/extra/fuegogin

This allows for users to bring their own framework. In this case the code here won't affect those two adaptors and nor should it IMO at least in the context of this PR. Hence why we should note, at least at this time, this option does not affect adaptors.

hamilton-earthscope: Thank you for the detailed explanation 🧑‍🎓 

Will fix this comment.

EwenQuim (approved):
Really high quality PR. ~~Merging right now.~~ Waiting for Dylan's approval.

We'll need to fit our serialization/deserialisation into this system: instead of using `if serde, ok := c.route.contentTypeSerde[contentType]; ok {`, we should populate the c.route.contentTypeSerde with our own default JSON/YAML functions that will be overridden by the option.

EwenQuim:
Currently letting the if to make sure it does not break anything internally before integrating this to the core system

hamilton-earthscope:
@EwenQuim @dylanhitt what do you think of these points?
> - All of the Read*() methods ([link](https://github.com/go-fuego/fuego/blob/08392cc3224cbf58b305b11628845ae1106a28ee/deserialization.go#L42)) take (ctx context.Context, input io.Reader, options readOptions) args. Perhaps Deserialize() should adhere to this.
> - All of the Send*() methods ([link](https://github.com/go-fuego/fuego/blob/08392cc3224cbf58b305b11628845ae1106a28ee/serialization.go#L157)) take (w http.ResponseWriter, r *http.Request, ans any) args. Perhaps Serialize() should adhere to this.

The current implementation in this PR using `io.ReadAll` is going to first buffer the body as `[]bytes` in memory. These changes would facilitate streaming de/serialization into out of the body type.

It would also better prepare for converting the built-in de/serialization logic into this `Serde` structure.

dylanhitt:
@hamilton-earthscope @EwenQuim I think another issue is warranted going forward on how we want to approach. @hamilton-earthscope brings up a good point it would be nice to make the default serialization and deserialization just be Serde's. We'd then only have one configuration type to really worry about. That said, it's a pretty breaking change. 

Could you please resolve the conflicts here and we can get this merged.

EwenQuim:
I'm ok for merging and create another issue to unify the serde processes

EwenQuim:
Also

- Should this be configurable at the server level in addition to the endpoint level?
  - No need, you can apply route options globally with `WithRouteOptions`
- Docs need updated
  - Yes that would be great!

dylanhitt:
> I'm ok for merging and create another issue to unify the serde processes

I agree.

I think we need to discuss something and nail down an implementation. For instance:

- Should this signal the deprecation of [`WithSerializer`](https://github.com/go-fuego/fuego/blob/d85ee245da097ba4c23caabc48bb8838b2e571dc/options.go#L50)?
- Should we keep WithSerializer and modify to fit this working into the core
- Do we want to continue with current ErrorSerializer as well?

dylanhitt:
I opened #533 Please feel free to add anything else we should discuss there.

ccoVeille:
I'm nitpicking here

But while [serde](https://github.com/serde-rs/serde) is a known lib in Rust.

I would have used `SerDes` or `SerDe`, instead of `Serde`, to respect somehow Go initialism way.

https://en.wikipedia.org/wiki/SerDes

Because here, for someone who don't know the concept could be confused.

I was, for record. I was like, why are they talking about a Rust library. I had forgot SerDes nickname.

documentation/docs/guides/serialization.md
```diff
@@ -96,6 +96,91 @@ func helloWorld(c fuego.ContextNoBody) (string, error) {
...
+```
+
+With this setup:
+- Requests with `Content-Type: application/vnd.keyvalue` will be deserialized using your custom deserializer
+- Responses with `Accept: application/vnd.keyvalue` will be serialized using your custom serializer
+- The route still supports standard Fuego content negotiation for JSON, YAML, XML, etc.
+
+Custom SerDes take precedence over built-in SerDes for the specified content type, meaning you can override Fuego's default de/serialization logic for JSON, XML, YAML, etc. if you choose.
+
+Note that your `SerDes`'s `Deserialize` implementation must return the type expected by your controller's body (`fuego.ContextWithBody[FooType]`) or an internal server error (HTTP status 500) will be thrown.
```
dylanhitt, line 180: Just a comment. 🤔 

Wonder if we could check this at init/route registration time and exit before start up some how. May be worth adding to the open issue

documentation/docs/guides/serialization.md
```diff
@@ -96,6 +96,91 @@ func helloWorld(c fuego.ContextNoBody) (string, error) {
...
+
+You can also use this option at the group or server level to easily apply custom content negotiation to multiple routes.
```
dylanhitt, line 182:
```suggestion
You can also use this option at the group or server level to easily apply custom content negotiation to multiple routes by using `WithRouteOptions`.
```

dylanhitt (changes requested):
Just about there. Just a comment and one docs addition. It would be nice if we could squash the commit history down into more logical units of work. Regardless just apply the the suggested docs change and I think we're good 👍

hamilton-earthscope:
@dylanhitt sorry i missed your edit. I'll try your version

dylanhitt:
@hamilton-earthscope I thought for sure I had merged this...

Could you rebase for me? LGTM

dylanhitt:
@hamilton-earthscope if I don't hear anything in the next week I'll take it over. I'll be sure to mark you as a contributor on the commits.

Thank you again.

PR #535, opened 2025-06-05
Gin adaptor middlewares

### LINKED ISSUE #534: Gin compatibility layer per route middlewares
**Problem**

As stated in #241 gin compatibility layer currently does not support per route middlewares. `option.Middleware` applies no middlewares to gin engine.

**Points of interest**

- [ ] Apply http middlewares (default ones for fuego `func (http.Handler) http.Handler`) to gin engine using wrapping (easy task)
- [ ] Apply gin middlewares to gin engine (`func (c gin.Context)`) (bit harder, description below)
- [ ] Global middlewares for gin? (very hard, have no ideas currently)

**Solution**

We don't introduce any API breaking changes, still store middlewares everywhere as `func (http.Handler) http.Handler`  and convert them to gin ones when registering routes. Major problem is wrapping. Convert http middleware to gin is not a problem. But we also want to pass gin middlewares and it's a problem since we store all middlewares as http ones.  So we have to convert one gin middleware twice. One when we pass it to storage using option, second when we apply it to gin. So idk if this way is reasonable cause converting gin middleware to http one is a bit runtime-consuming

evilenzo:
Resolves #534 (more details there)

### DIFF
--- a/net_http_mux.go
+++ b/net_http_mux.go
@@ -173,7 +173,6 @@ func withMiddlewares(controller http.Handler, middlewares ...any) http.Handler {
 		} else {
 			panic("wrong middleware format for fuego engine")
 		}
-
 	}
 	return controller
 }


evilenzo:
So I decided to simplify task with such separation:
1. Only default http middlewares for default server
2. Only gin middlewares for gin server

dylanhitt:
Is there not a way to convert `func(http.Handler) http.Handler` into gin.Context? 🤔

dylanhitt:
ehhhh, i guess it does make sense that each adaptors community would want to use their adaptors. If someone wanted to use a fuego middleware wrapping themselves wouldn't be very difficult as pretty much every community has a func to wrap http.Handler...

evilenzo:
> Is there not a way to convert `func(http.Handler) http.Handler` into gin.Context? 🤔

Indeed there is. I explained it in issue more detailed but my English might be bad so...

We have two types of middlewares:
1. `func (http.Handler) http.Handler` - fuego/http
2. `func (c *gin.Context)` - gin

Before this PR all middlewares was stored as `func (http.Handler) http.Handler` and it creates two problems with gin:
1. OK, we can convert `func (http.Handler) http.Handler` to `func (c *gin.Context)` when applying middlewares to gin router. Not a big problem (but still a problem cause gin only has `gin.WrapH` and `gin.WrapF` functions that doesn't work with such types of middlewares)
2. But we also want users to be able to use their gin middlewares `func (c *gin.Context)`. And this is a major problem. Imagine we create some custom option to pass such middleware. But we store all middlewares as `func (http.Handler) http.Handler`. So we have to convert gin middleware to http middleware. That's the first conversion (and it's not an easy task. It has runtime overhead cause it's wrapping). Then we want to apply this middleware to gin. And we have to convert http middleware to gin middleware. That's the second conversion. So to use one gin middleware we must convert it twice and gain runtime overhead. I don't think that such solution is reasonable.

On the other hand, if you are ok that we store middlewares as `[]any` but you think that we should also have ability to apply both types of middlewares to gin... I don't think that combining different middlewares by our hands is a good idea since even gin doesn't have conversion from `func (http.Handler) http.Handler` middleware to it's own

evilenzo:
Hm... I see no problem with `Route[...]`. Problem is a little bit down the flow (with `fuego.BaseRoute`). Look at the options:

```go
options ...func(*fuego.BaseRoute)
```

Main problem here is they are applied to `fuego.BaseRoute`. Only one thing that uses it is `NewBaseRoute()` and I don't know if we can change much there. 

```go
func handleFuego[T, B, P any](engine *fuego.Engine, ginRouter gin.IRouter, method, path string, fuegoHandler func(c fuego.Context[B, P]) (T, error), options ...func(*fuego.BaseRoute)) *fuego.Route[T, B, P] {
	baseRoute := fuego.NewBaseRoute(method, ginToFuegoRoute(path), fuegoHandler, engine, options...)
	// ...
}
```

We want to apply option to gin but we can apply it only to BaseRoute

evilenzo:
@dylanhitt @EwenQuim sorry, but I still can't understand your idea... It would be cool if you provide some draft/MRE so I can finish this PR

roskee:
@evilenzo any plans to make this work?

PR #537, opened 2025-06-05
Unused code in echo-compat

dylanhitt:
left over from when we switched over to the `engine.RegisterOpenAPIRoutes`


ccoVeille:
Also a commit from #536

dylanhitt:
ope, yeah gonna close this one.

PR #538, opened 2025-06-06
chore: removes some unused code

dylanhitt:


examples/echo-compat/handlers.go
```diff
@@ -36,12 +36,6 @@ func fuegoControllerPost(c fuego.ContextWithBody[HelloRequest]) (*HelloResponse,
 	}, nil
 }
 
-func serveOpenApiJSONDescription(s *fuego.OpenAPI) echo.HandlerFunc {
-	return func(ctx echo.Context) error {
-		return ctx.JSON(http.StatusOK, s.Description())
-	}
-}
```
ccoVeille, line 43: Wasn't it detected by unused linter?

dylanhitt: no actually. I only noticed from my IDEs linter

PR #539, opened 2025-06-08
refactor: define type ServerOption for With server funcs

dylanhitt:
Semi discussed in #518 and similar to #530 but for Server option. Same motive as #530.


server.go
```diff
@@ -69,6 +69,9 @@ type Server struct {
 	isTLS                  bool
 }
 
+// An ServerOption represents configures the behavior of an [Server].
+type ServerOption func(*Server)
```
ccoVeille, line 73: I'm worried this change could break existing code.

Someone having code already calling the methods you are updating the signature might be incompatible depending on how they implemented their code

One solution could be to use a type alias

```suggestion
type ServerOption = func(*Server)
```

ccoVeille: But apparently, it was done with #530.

ccoVeille (approved):

PR #540, opened 2025-06-08
BREAKING: define type ParamOption for all Param option funcs

dylanhitt:
Relates to #518

Same motive as: #530 and #539


dylanhitt:
I think if someone wrote a custom param option that feeds into a route option it would break. Didn’t bother to test it though.

EwenQuim:
I don't think that is breaking :)

dylanhitt:
I can recommit as refactor you’d like?

ccoVeille:
> I don't think that is breaking :)

It's a breaking change, but a minor one.

It can be acceptable. I'm unsure whether that's the conclusion you came to @EwenQuim

dylanhitt:
For instance: 
image

dylanhitt:
I’ll provide a second commit addressing all of them.

dylanhitt:
Is anyone aware how type aliases are presented in the godoc?

EwenQuim:
You can preview with `make docs-open`

EwenQuim:
Also, I think we can have a `type RouteOption = func(*BaseRoute)`

EwenQuim:
As you want

EwenQuim (approved):

PR #542, opened 2025-06-09
ci: update github workflow perms

dylanhitt:
Security tab has been bugging me


dylanhitt:
Yeah look at the security tab. There is some explaining. But 🤷

dylanhitt:
We do also already do this for the docs stuff as well.

PR #546, opened 2025-06-11
feat: update OptionRequestBody to allow for multiple request bodies in spec

dylanhitt:
Add a use for this today.

Also this can provide a caller a way to gen proper types of spec #524 in cases where the content-type specified could cause the route to deserialize a different type altogether.


dylanhitt:
@EwenQuim you wouldn't happen to have a few spare moments to give this a look would you? Trying to get all my outstanding working closed before I move at the end of the summer.

ccoVeille (approved):
I'm not sure about the feature, but the code seems good.

PR #551, opened 2025-06-17
Support for specifying key when creating Security struct

imthatgin:
Hey. This is just a quick PR to be able to customize the key used for the Security struct. Otherwise, I'd need to copy the entire module to modify it since the key is private.

### DIFF
--- a/security.go
+++ b/security.go
@@ -54,6 +54,10 @@ func NewSecurity() Security {
 		panic(err)
 	}
 
+	return NewSecurityWithKey(key)
+}
+
+func NewSecurityWithKey(key *ecdsa.PrivateKey) Security {
 	return Security{
 		key:             key,
 		Now:             time.Now,
--- a/security_test.go
+++ b/security_test.go
@@ -2,6 +2,9 @@ package fuego
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


security.go
```diff
@@ -54,6 +57,10 @@ func NewSecurity() Security {
 		panic(err)
 	}
 
+	return NewSecurityWithKey(key)
+}
+
+func NewSecurityWithKey(key *ecdsa.PrivateKey) Security {
```
dylanhitt, line 63: Hmmm. We've been doing this with an options pattern for basically everything else. i.e Engine, Server, Route. I kinda think we should take the same approach here. So like the NewSecurity func would be this

```golang
// An SecurityOption represents configures the behavior of the Security struct.
type SecurityOption = func(*Security)

// NewEngine creates a new Engine with the given options.
// For example:
//
//	engine := fuego.NewSecurity(
//      WithKey(key)
// )
//
// Options all begin with `With`.
func NewSecurity(options ...SecurityOption) *Security { ... }
```

dylanhitt (changes requested):
Sorry it took me forever to get to this. 😅

PR #564, opened 2025-07-28
fix(validation.go): prevent panic when validate is called with nil

rene-gentzen-mittelstand-ai:
I have a route that wants to accept any valid JSON. I implemented this with `fuego.ContextWithBody[any]`. 
When posting no request body or the literal `null` to this route the thread silently panics and I never get a response, only a client error "Connection was forcibly closed by a peer." I don't think this should happen under any circumstance.

I have tracked down and fixed the panicking code but have since also read the doc string for `ContextWithBody` that says "Please do not use a pointer type as parameters.". Wouldn't it be okay to use `any`? Would it lead to inconsistencies if my change got implemented? With my change I simply get a `nil` body. 

The auto-generated OpenAPI spec looks like this, which is okay for my usecase:
image

### DIFF
--- a/validation.go
+++ b/validation.go
@@ -32,7 +32,8 @@ func explainError(err validator.FieldError) string {
 var v = validator.New()
 
 func validate(a any) error {
-	if reflect.TypeOf(a).Kind() != reflect.Struct {
+	t := reflect.TypeOf(a)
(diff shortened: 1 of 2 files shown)


dylanhitt (approved):
I think the documentation should remain as is. Your case seems unique. 

Thank you

Issue #579, opened 2025-08-28
Custom UI example in the documentation dont work


NobleMajo:
Hii 👋 ,
The following custom UI section in the docs cant work in the current state:
https://go-fuego.dev/docs/guides/openapi#custom-ui

I followed the example and got this message because of a openapi specs version mismatch:
```txt
Unable to render this definition

The provided definition does not specify a valid version field.

Please indicate a valid Swagger or OpenAPI version field. Supported version fields are swagger: "2.0" and those that match openapi: 3.0.n (for example, openapi: 3.0.0).
```

My specs provide version 3.1.0 and the implementation in the example cant handle it.
Maybe that documentation section should be rewritten/fixed or the section should be removed.

Best regards & thank you in advance, majo

NobleMajo:
I found the issue in the docs.

The dependency in the code snippet should be `github.com/swaggo/http-swagger/v2` and not `github.com/swaggo/http-swagger`.

dylanhitt:
Ah yeah those docs probably need to be updated.

PR #581, opened 2025-08-28
Fix custom ui example in docs

NobleMajo:
The docs custom ui example did not work with the latest fuego version because of a version mismatch of the openapi specs.

More here: https://github.com/go-fuego/fuego/issues/579


NobleMajo:
your welcome

PR #588, opened 2025-09-17
Fixed typo in `example` tag parsing

nrydanov:
Hello! Today I tried to use actual `main` branch in my work project to test out current state of `Params()` method of `Context`. 

I've tested it out on small example and found out that it's actually implemented. However, on my work project I suddenly got panic.

I found small typo — `description` tag was used instead of `example` when parsing struct tags. I've changed only 1 line and it was enough to break one of the tests. 

Parsing `description` instead of `example` was masking some errors in generating OpenAPI spec for slice/array types, so I needed few more changes to fix that. 

Here's three things to debate:
1. I've introduced new GoType - array, which is obviously OpenAPI type,
   not Go type. I would actually change this naming to something else.
2. I've changed hardcoded string schema with dynamic one based on real
   value (with previous behavior as default).
3. As some things were fixed, test still failed, but now because of
   difference with golden file. I replaced it with new one.

### DIFF
--- a/option.go
+++ b/option.go
@@ -266,7 +266,7 @@ func buildParam(name string, options ...ParamOption) (OpenAPIParam, *openapi3.Pa
 		Name:        name,
 		In:          string(param.Type),
 		Description: param.Description,
-		Schema: schema.NewRef(),
+		Schema:      schema.NewRef(),
 	}
 	if param.GoType != "" {
 		openapiParam.Schema.Value.Type = &openapi3.Types{param.GoType}


openapi.go
```diff
@@ -275,9 +319,13 @@ func (route *Route[ResponseBody, RequestBody, Params]) RegisterParams() error {
...
 			var params []ParamOption
-			example, _ := field.Tag.Lookup("description")
+			example, _ := field.Tag.Lookup("example")
 			if example != "" {
-				params = append(params, ParamExample("example", example))
+				parsedValue, err := parseExampleFromTag(example, field.Type)
+				if err != nil {
+					return fmt.Errorf("parsing example tag: %s", err)
+				}
+				params = append(params, ParamExample("example", parsedValue))
```
dylanhitt, line 328: I'm not sure I follow your issue you exactly. Could you possibly send a snipped of code that was failing?

dylanhitt: Oh I see. Examples is a map. So for this: 

```golang
option.Query("name", "Filter by name", param.Example("cat name", "felix"), param.Nullable(), param.Example("cat name", "bob")),
```

you end up with:

```json
"examples": {
        "cat name": {
	      "value": "felix"
	},
}
```

where you want the ability to do:

```json
"examples": {
"my example": {
	"value": [
		1,
		2,
		3
	]
}
}
```

dylanhitt:
So description really isn't a typo we support both go tags to parse into a openapi `example`. Removing `description` is pretty abrupt breaking change.

examples/petstore/lib/testdata/doc/openapi.golden.json
```diff
@@ -1424,6 +1424,15 @@
 						}
 					},
 					{
+						"examples": {
```
dylanhitt, line 1427: So I think you updated this by hand? The proper way to update this to use `make golden-update`. You should update a route in the `example/petstore/controllers` to show the change.

openapi.go
dylanhitt: Okay yeah now I see. Sorry it's been a awhile since I've been here.

dylanhitt:
So to address the things for debate

> 1. I've introduced new GoType - array, which is obviously OpenAPI type,
   not Go type. I would actually change this naming to something else.

I am open to suggestions. I wouldn't perform this in the PR though. 

> 2. I've changed hardcoded string schema with dynamic one based on real
   value (with previous behavior as default).
   
That is fine. Much better!

> 3. As some things were fixed, test still failed, but now because of
   difference with golden file. I replaced it with new one.

So golden can be updated as I mentioned in other comments. I took me awhile to understand what this was even doing as the bug was blurring my vision. Thank you for this. A few housekeeping things. 

1. Add a test for `ParamArray` in `param_test.go`
2. Add some tests in `parameter_registration_test.go`. Right now `parseExampleFromTag` has no coverage. So basically test `RegisterParams`

Awesome work. Thank you.

PR #597, opened 2025-10-07
fix: cast error HTTPError in default error handle regardless of reference or value type

### LINKED ISSUE #567: Unexpected serialization of nested fuego.HTTPError
**To Reproduce**

1. Send an invalid request body to a fuego controller
2. Access the request body in the controller
3. Validation triggers and returns a `fuego.HTTPError`
4. Intercept the error because you want to change it.
5. Create your own `fuego.HTTPError` with custom status code etc.
7. Assign the intercepted `fuego.HTTPError` to `newErr.Err`.
8. Return `newErr` from the controller.
9. The client sees a response with `newErr.Err.Errors` serialized into the `errors` field.

**Expected behavior**
The description of `fuego.HTTPError` states "Developer readable error message. Not shown to the user to avoid security leaks."
Under no circumstances would I expect information from this field to leak into the response. 

**Screenshots**
```go
	req, err := c.Body()
	if err != nil {
		switch ferr := err.(type) {
		case fuego.HTTPError:
			if ferr.Title == "Validation Error" {
				ret := &fuego.HTTPError{
					Title:  http.StatusText(http.StatusUnprocessableEntity),
					Detail: fmt.Sprintf("The request body is malformed: %s", ferr.Detail),
					Status: http.StatusUnprocessableEntity,
					Err:    err,
				}
				return "", ret
			}
		}
	}
```

Image

github.com/go-fuego/fuego v0.18.8

go version go1.24.4 linux/amd64

**Additional context**
I am especially puzzled because it seems like the information leaks from the inner error into the outer error. If I set `newErr.Errors` explicitely to some fuego.ErrorItems and leave `newErr.Err` blank, the response weirdly enough does _not_ contain the `errors` field.

dylanhitt:
Closes #567

### DIFF
--- a/errors_test.go
+++ b/errors_test.go
@@ -45,23 +45,22 @@ func TestErrorHandler(t *testing.T) {
 		err := NotFoundError{
 			Err: errors.New("Not Found :c"),
 		}
-		errResponse := ErrorHandler(context.Background(), err)
-		require.ErrorAs(t, errResponse, &HTTPError{})
+		var errHTTP HTTPError
+		require.ErrorAs(t, ErrorHandler(context.Background(), err), &errHTTP)
 		require.ErrorContains(t, err, "Not Found :c")
-		require.ErrorContains(t, errResponse, "Not Found")
-		require.ErrorContains(t, errResponse, "404")
-		assert.Equal(t, http.StatusNotFound, errResponse.(HTTPError).StatusCode())
+		require.ErrorContains(t, errHTTP, "Not Found")
+		require.ErrorContains(t, errHTTP, "404")
+		assert.Equal(t, http.StatusNotFound, errHTTP.StatusCode())
 	})
 
 	t.Run("not duplicate HTTPError", func(t *testing.T) {
 		err := HTTPError{
 			Err: errors.New("HTTPError"),
 		}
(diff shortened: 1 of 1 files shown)


errors_test.go
```diff
@@ -145,6 +145,20 @@ func TestHandleHTTPError(t *testing.T) {
...
+			Title:  "HTTPError",
+			Errors: []ErrorItem{{Name: "my name"}},
+			Err:    errors.New("my new error"),
+		}
+		errResponse := HandleHTTPError(context.Background(), err)
+		require.ErrorAs(t, errResponse, &HTTPError{})
+		require.ErrorContains(t, err, "my new error")
+		require.ErrorContains(t, errResponse, "HTTPError")
+		require.ErrorContains(t, errResponse, "500")
+		require.Equal(t, http.StatusInternalServerError, errResponse.(HTTPError).StatusCode())
```
ccoVeille, line 160: Is there a need for the type cast here?


```suggestion
		require.Equal(t, http.StatusInternalServerError, errResponse.StatusCode())
```

dylanhitt: Actually it does needs to be cast. HandleHTTPError still is `func HandleHTTPError(ctx context.Context, err error) error { ... }`

PR #604, opened 2025-10-14
fix: init groups server with new server.RouteOptions when Grouping

### LINKED ISSUE #478: Middleware not consistently applied in nested route groups

dylanhitt:
Closes #478 

There may be a more elegant way of doing this. But I'm not sure how.

### DIFF
--- a/net_http_mux.go
+++ b/net_http_mux.go
@@ -23,23 +23,23 @@ import (
 //	})
 //	s.Run()
 func Group(s *Server, path string, routeOptions ...func(*BaseRoute)) *Server {
-	if path == "/" {
+	switch cleaned, found := strings.CutSuffix(path, "/"); {
+	case path == "/":
 		path = ""
-	} else if path != "" && path[len(path)-1] == '/' {
-		slog.Warn("Group path should not end with a slash.", "path", path+"/", "new", path)
+	case found:
+		slog.Warn("Group path should not end with a slash.", "path", path, "new", cleaned)
 	}
 
-	ss := *s
-	newServer := &ss
+	newServer := *s
+	newServer.routeOptions = append([]func(*BaseRoute){}, s.routeOptions...)
 	newServer.basePath += path
 
 	if autoTag := strings.TrimLeft(path, "/"); !s.disableAutoGroupTags && autoTag != "" {
 		newServer.routeOptions = append(s.routeOptions, OptionTags(autoTag))
 	}
-
 	newServer.routeOptions = append(newServer.routeOptions, routeOptions...)
 
-	return newServer
+	return &newServer
 }
 
 // All captures all methods (GET, POST, PUT, PATCH, DELETE) and register a controller.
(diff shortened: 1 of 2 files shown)


dylanhitt:
@EwenQuim is this fine to merge?

net_http_mux.go
```diff
@@ -23,23 +23,23 @@ import (
...
 func Group(s *Server, path string, routeOptions ...func(*BaseRoute)) *Server {
-	if path == "/" {
+	switch {
+	case path == "/":
 		path = ""
-	} else if path != "" && path[len(path)-1] == '/' {
-		slog.Warn("Group path should not end with a slash.", "path", path+"/", "new", path)
+	case strings.HasSuffix(path, "/"):
+		slog.Warn("Group path should not end with a slash.", "path", path, "new", strings.TrimSuffix(path, "/"))
 	}
```
ccoVeille: Maybe you could use CutSuffix that returns a bool flag

This way it would look like this (pseudo code)

```suggestion
	case cleaned, found := strings.CutSuffix(path, "/"); found:
		slog.Warn("Group path should not end with a slash.", "path", path, "new", cleaned)
	}
```

dylanhitt: How does that look now? You can't instantiate inside case statement AFAIK.

PR #608, opened 2025-10-28
feat: add the ErrorWithTitle interface to set Titles of errors

dylanhitt:


errors.go
```diff
@@ -23,6 +23,11 @@ type ErrorWithDetail interface {
 	DetailMsg() string
 }
 
+type ErrorWithTitle interface {
+	error
+	TitleString() string
```
ccoVeille: I feel you used this because HTTP Error has a Title field

Why not staying consistent with Detail that had the same issue and used DetailMsg


```suggestion
	TitleMsg() string
```

dylanhitt: I can yeah. I really don’t like either 😞.

ccoVeille: Wording is a pain, I know

ccoVeille: ErrorTitle() string could be a solution maybe

dylanhitt: I debated on that. I do think DetailMsg should become the same. Albeit breaking, what are your thoughts?

errors.go
```diff
@@ -238,6 +245,12 @@ func HandleHTTPError(ctx context.Context, err error) error {
 		errResponse.Detail = errorDetail.DetailMsg()
 	}
 
+	// Check for title
+	var errorTitle ErrorWithTitle
+	if errors.As(err, &errorTitle) {
+		errResponse.Title = errorTitle.TitleString()
+	}
```
ccoVeille: I feel you are replicating the pattern that exists above, but I would prefer to apply the logic of error variable always starts with err


```suggestion
	// Check for title
	var errWithTitle ErrorWithTitle
	if errors.As(err, &errWithTitle) {
		errResponse.Title = errWithTitle.TitleString()
	}
```

errors.go
ccoVeille: I agree in a perfect words ErrorWithTitle and ErrorWithDetail should apply the sane logic.

But world is not prefect. I'm unsure you shlukc fight a breaking change dragon god such minor change

errors.go
dylanhitt: Updated with co author'd

ccoVeille (approved):

PR #616, opened 2025-11-19
feat: provide the ability to set http.Request in MockContext

### LINKED ISSUE #568: Allow setting of Request in MockContext
One of my controllers accesses `c.Request().RequestURI`, `c` being a `fuego.ContextNoBody`.
In my tests I use `fuego.NewMockContext()`.  `fuego.MockContext` has a private field `request` of type `*http.Request` that can be read with `mockCtx.Request()` but is not set during initialization, thus returning `nil`. 

I would like `fuego.MockContext.request` to be public / settable.

For now I use this hack:
```go
func setMockContextRequest[B, P any](ctx *fuego.MockContext[B, P], req *http.Request) {
	v := reflect.ValueOf(ctx).Elem().FieldByName("request")
	reflect.NewAt(v.Type(), unsafe.Pointer(v.UnsafeAddr())).Elem().Set(reflect.ValueOf(req))
}
```

dylanhitt:
closes #568

Elected to go with SetRequest instead of in the constructor. Making the field public is not an option as it has to adhere to the Context interface.

### DIFF
--- a/mock_context.go
+++ b/mock_context.go
@@ -107,6 +107,10 @@ func (m *MockContext[B, P]) PathParamInt(name string) int {
 	return 0
 }
 
+func (m *MockContext[B, P]) SetRequest(r *http.Request) {
+	m.request = r
+}
+
 // Request returns the mock request
 func (m *MockContext[B, P]) Request() *http.Request {
 	return m.request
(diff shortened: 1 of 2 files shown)


Resteklicken:
> @Resteklicken
> 
> I believe this should solve one of your issues.

Yup, thanks a lot for this!

PR #618, opened 2025-11-19
fix: omitempty should set spec as required not nullable

### LINKED ISSUE #609: Unexpectedly missing required attribute from OpenAPI-generated schema
```go
type UserResponse struct {
    ID        string `json:"id"`
    Email     string `json:"email"`
    FirstName string `json:"firstName"`
    LastName  string `json:"lastName"`
}


func (h *UserController) GetUser(c fuego.ContextNoBody) (UserResponse, error) {
    //...
```

I would expect fields like `email` above to be marked as required in the schema in the OpenAPI spec as they don't have `omitempty` and aren't nullable. If I specify `validate:"required"` on the struct annotations they will be marked as required, but that does not seem correct.

dylanhitt:
Closes #609

Fairly major change when it comes to spec generation. If a field does not contain the json tag of omitempty it will be considered required. If omitempty is present and `validate:"required"` is present is will be marked as required to maintain some backwards compatibility. The major breaking change here is that omitempty is no longer considerable nullable, which, it shouldn't be as a reference type is how kin-openapi determines if a filed is nullable already.

@andyfleming you may be interested in look at the implementation of this.

### DIFF
--- a/examples/petstore/lib/testdata/doc/openapi.golden.json
+++ b/examples/petstore/lib/testdata/doc/openapi.golden.json
@@ -41,15 +41,17 @@
 										"type": "string"
 									}
 								},
+								"required": [
+									"type",
+									"value"
+								],
 								"type": "object"
 							},
 							"treats": {
 								"items": {
-									"nullable": true,
 									"properties": {
 										"brand": {
 											"description": "The brand of the treat",
-											"nullable": true,
 											"type": "string"
 										},
 										"itemId": {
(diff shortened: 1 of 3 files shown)


dylanhitt:
Hey @EwenQuim you've seen pretty busy lately, but this is something I'd like a second set of eyes on. 

I believe this comment sums this up pretty well: https://github.com/go-fuego/fuego/issues/609#issuecomment-3554411860

EwenQuim:
I agree, LGTM!

The downside is that we will lose input validation: a field will be marked as required but we can send `undefined` in a field of the JSON payload and it will still pass validation, and it will be deserialized as an empty string. A solution can be to set min=1 which will cover 99% of the use-cases (why would anyone accept just an empty string).

I hope people will use different DTOs and will set the validate:"required" anyway for entity creation to be semantically correct instead of using min=1.

PR #686, opened 2026-02-16
chore: bump go version to 1.25.7

jub0bs:


dylanhitt:
How about go 1.25.7. This would be a regression of the latest crypto bug.

dylanhitt (approved):

PR #696, opened 2026-03-02
fix: use LastIndex for closing bracket in transformTypeName

### LINKED ISSUE #560: Documentation Generation Issue with Generic array Types in Controllers
### Issue Title
Documentation Generation Issue with Generic array Types in Controllers

#### To Reproduce

1. Create a controller that returns a complex type with generics, specifically an array within the generic type.
2. Generate the documentation using Fuego.
3. Observe that documentation for the array-type generics is not generated.

#### Expected behavior
A clear and concise description of what you expected to happen:

The documentation for controllers returning generic types should be generated correctly even if the type contains arrays, without requiring an alias for the array type.

#### Framework version (please check if it happens with the latest version before posting):
v0.18.8

#### Go version (please check if it happens with the latest Go version before posting):
v1.24.2

#### Additional context
Add any other context about the problem here:

In the case where a controller returns a complex type with generics, Fuego will generate the documentation correctly only if the generic type is not an array.

- Example working correctly: `(*dto.HttpResponse[model], error)` => documentation is generated.
- Example failing: `(*dto.HttpResponse[[]model], error)` => documentation is not generated, as the name of the resource is based on the generic type (which includes square brackets [] not allowed in OpenAPI 3.0) :
`ERROR Error validating spec error="invalid components: schema \"HttpResponse_[\": identifier \"HttpResponse_[\" is not supported by OpenAPIv3 standard (charset: [\"a-zA-Z0-9._-\"])"`

Defining an alias for the type can resolve the issue, but adds unnecessary verbosity:

- Aliased version: `(*dto.HttpResponse[ArrayModel], error)`.

0-don:
## Summary
- `transformTypeName` used `strings.Index` to find the closing `]` bracket, which matched the `]` from `[]` in slice type parameters (e.g. `Response[[]Model]`) instead of the actual closing generic bracket
- This caused all `Response[[]T]` variants to produce the same schema name `Response_[`, collapsing them into a single OpenAPI component schema
- Switched to `strings.LastIndex` so the outermost closing bracket is used

## Reproduction
Any controller returning a generic wrapper with a slice type parameter produces an invalid/colliding schema name:

```go
// These all produce schema name "Response_[" instead of unique names
func A(c ContextNoBody) (*Response[[]ModelA], error) { ... }
func B(c ContextNoBody) (*Response[[]ModelB], error) { ... }
```

## Changes
- `openapi.go`: Replace `strings.Index(s, "]")` with `strings.LastIndex(s, "]")` in `transformTypeName`
- `openapi_test.go`: Add `Test_transformTypeName` covering simple generics, package paths, and slice type parameters

## Example

| Input | Before (broken) | After (fixed) |
|---|---|---|
| `Response[model.User]` | `Response_model.User` | `Response_model.User` |
| `Response[[]model.QuotaData]` | `Response_[` | `Response_[]model.QuotaData` |
| `Response[[]model.OtherData]` | `Response_[` | `Response_[]model.OtherData` |

Fixes #560

### DIFF
--- a/examples/petstore/controllers/pets.go
+++ b/examples/petstore/controllers/pets.go
@@ -56,6 +56,10 @@ func (rs PetsResources) Routes(s *fuego.Server) {
 		option.Description("Generic request and response. Showcase Fuego's support for generic input & output."),
 	)
 
+	fuego.Get(petsGroup, "/all-wrapped", rs.getAllPetsWrapped,
+		option.Description("Get all pets in a generic wrapped response with a slice type parameter"),
+	)
+
 	fuego.Get(petsGroup, "/by-age", rs.getAllPetsByAge,
 		option.Description("Returns an array of pets grouped by age"),
 		option.Middleware(dummyMiddleware),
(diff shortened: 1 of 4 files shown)


dylanhitt (commented):
Thank you for this fix it's greatly appreciated. Just that one change

Issue #700, opened 2026-03-05
Trailer: Server-Timing header sent on 204 No Content responses breaks Node.js proxies


fank:
**To Reproduce**

1. Create a typed handler that returns 204 No Content:
```go
func (h *Handler) Logout(c fuego.ContextNoBody) (any, error) {
    c.SetStatus(http.StatusNoContent)
    return nil, nil
}

fuego.Post(s, "/api/v1/auth/logout", handler.Logout)
```
2. Proxy the Go server through a Node.js-based dev proxy (e.g., Vite dev server with `server.proxy`)
3. Call `POST /api/v1/auth/logout`
4. Node.js v24 crashes with:
```
Error [ERR_HTTP_TRAILER_INVALID]: Trailers are invalid with this transfer encoding
    at ServerResponse._storeHeader (node:_http_outgoing:522:11)
```

The raw response from Fuego includes:
```http
HTTP/1.1 204 No Content
Trailer: Server-Timing
Server-Timing: fuegoReqInit;dur=0
```

**Expected behavior**

Fuego should not set the `Trailer` header on responses with status codes that must not have a body (1xx, 204, 304), per RFC 9110 §6.5.1: trailers are only valid when the transfer coding is chunked.

**Framework version:** v0.19.0

**Go version:** go1.26.0 linux/amd64

**Additional context**

Node.js v24 tightened HTTP trailer validation, making this crash consistently. Older Node versions silently ignored the invalid trailer. The workaround is to use `PostStd` with a raw `http.HandlerFunc` instead of typed handlers for 204 responses.

PR #701, opened 2026-03-06
fix: do not set trailer header when no content

dylanhitt:
patched #700

@fank i believe this should fix your issue.


dylanhitt:
Okay to address your comment

1. this is a deeper issue with fuego in that SetStatus calls WriteHeader. Ideally if we wrapped responsewriter and our users called SetStatus and error was returned the error code wouldn't be one of the things you described, but this is a long standing bug and requires a bit of work to address. see: https://github.com/go-fuego/fuego/pull/376

2. I'll switch it.

PR #706, opened 2026-03-06
Feat/use schemarefs

dylanhitt:
We used to have support for this. At work a fellow dev was asking for the ability to use a nested structure in one our generated clients. I noticed a lot of libs from TS will make even nested structures turns out the feature has been in kin-openapi all along.


EwenQuim (approved):
Nice one!

PR #725, opened 2026-04-09
fix: prevent phantom contentless default response in OpenAPI spec

0-don:
## Summary

Every route in the generated OpenAPI spec includes a phantom contentless `"default"` response that no one explicitly added. This causes code generation tools to emit void unions for all response types.

### The bug origin

`kin-openapi`'s `NewResponses()` with no arguments auto-inserts `"default": {"description": ""}` with no content schema. In fuego, this gets triggered through two paths:

1. **`option.go`** (lines 411, 433, 477): `OptionAddError`, `OptionAddResponse`, and `OptionDefaultResponse` call `openapi3.NewResponses()` to initialize `Operation.Responses` when nil
2. **`openapi3.Operation.AddResponse`** (inside kin-openapi): also calls `NewResponses()` when `Responses` is nil, triggered when `addResponseIfNotSet` adds the global 400/500 responses

### Reproduction

Any route produces a spec with the phantom default:

```json
{
  "responses": {
    "200": { "description": "OK", "content": { ... } },
    "400": { "description": "Bad Request", "content": { ... } },
    "500": { "description": "Internal Server Error", "content": { ... } },
    "default": { "description": "" }
  }
}
```

Code generation tools (Orval, openapi-typescript, openapi-generator) interpret the contentless default as `void`:

```typescript
// Generated (broken):
type Response = void | ActualResponseType;
// Expected:
type Response = ActualResponseType;
```

### Changes

- **`openapi.go`**: Pre-initialize `Operation.Responses` with `NewResponsesWithCapacity(4)` in both `RegisterOpenAPIOperation` functions, before any `AddResponse` calls can hit kin-openapi's nil-check phantom injection
- **`option.go`**: Replace `NewResponses()` with `NewResponsesWithCapacity(1)` in the three option functions that initialize responses
- **`option_test.go`**: Adjust response count assertions (remove the phantom default from expected counts, e.g. `5` to `4` for `// 200, 400, 409, 500`)
- **`openapi.golden.json`**: Regenerated, 42 lines removed (7 phantom `"default"` response blocks)

### Before

```
route.Operation.Responses.Len() == 5  // 200, 400, 409, 500, default
```

### After

```
route.Operation.Responses.Len() == 4  // 200, 400, 409, 500
```

### Related

Root cause fix submitted to kin-openapi: [getkin/kin-openapi#1148](https://github.com/getkin/kin-openapi/pull/1148). This PR works independently as a workaround by ensuring `Responses` is never nil when responses are added.

### DIFF
--- a/examples/petstore/lib/testdata/doc/openapi.golden.json
+++ b/examples/petstore/lib/testdata/doc/openapi.golden.json
@@ -504,9 +504,6 @@
 							}
 						},
 						"description": "Internal Server Error _(panics)_"
-					},
-					"default": {
-						"description": ""
 					}
 				},
 				"summary": "filter pets",
@@ -630,9 +627,6 @@
 							}
 						},
 						"description": "Internal Server Error _(panics)_"
-					},
-					"default": {
-						"description": ""
 					}
 				},
 				"summary": "post pets",
@@ -784,9 +778,6 @@
 							}
 						},
 						"description": "Internal Server Error _(panics)_"
-					},
-					"default": {
-						"description": ""
 					}
 				},
 				"summary": "get all pets",
@@ -877,9 +868,6 @@
 							}
 						},
 						"description": "Internal Server Error _(panics)_"
-					},
-					"default": {
-						"description": ""
 					}
 				},
 				"summary": "get all pets wrapped",
@@ -981,9 +969,6 @@
 							}
 						},
 						"description": "Internal Server Error _(panics)_"
-					},
-					"default": {
-						"description": ""
 					}
 				},
 				"summary": "get all pets by age",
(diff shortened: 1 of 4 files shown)


dylanhitt:
@0-don I see you've got your fix in kin merged already. Maybe we should just wait til they cut a release? You can do the honors as well. 

Thank you for looking into this. It's really appreciated.

dylanhitt:
@0-don seems your change upstream was tagged, if you want to do the honors.

A side note. The new tag on kin allows for proper validation on on typed arrays for declaring nullable. So that should allow us to support proper 3.1 style nullable.

0-don:
> @0-don seems your change upstream was tagged, if you want to do the honors.
> 
> A side note. The new tag on kin allows for proper validation on on typed arrays for declaring nullable. So that should allow us to support proper 3.1 style nullable.

I am currently busy, its fine if you do it, we can close this pull then

dylanhitt:
Closing in favor of #741

Issue #707, opened 2026-03-15
Add `enum` support for parameters


mguentner:
I'd like to supply an `enum` when specifying a parameter like so:

```go
option.Query("foobar", "Some query param.", param.Enum("a", "b", "c")),
```

Spec: https://swagger.io/docs/specification/v3_0/data-models/enums/

dylanhitt:
Idea for making this work on request bodies as well: https://github.com/go-fuego/fuego/issues/222#issuecomment-2708482940

PR #745, opened 2026-05-03
feat: parse validate:"oneof=..." into OpenAPI enum

0-don:
## Summary

`parseValidate` handles `min=`, `max=`, and `required`, but ignores `oneof`. Runtime validation already enforces it via `go-playground/validator`, so the spec is weaker than the actual contract: codegen tools (Orval, openapi-generator) emit `string` instead of a typed union.

This PR populates `schema.Enum` from `oneof=...`, with values typed for `string`, `integer`, and `number` schemas. Same convention as `min=` / `max=` (validate-tag-based, no new tag).

Addresses part of #707 for body / query-struct fields. #735 covers the `option.Query(..., param.Enum(...))` path.

### Changes

- **schema_customizer.go**: parse `oneof=...` in `parseValidate`. Whitespace-split per `go-playground/validator`'s spec, values typed by schema kind
- **schema_customizer_test.go**: `TestParseValidateOneof` covering string, integer, number, `required,oneof` combo, and no-`oneof` no-op

### Before / After

```go
Bucket string `query:"bucket" validate:"oneof=1m 5m 15m 1h 1d"`
```

| | Before | After |
| --- | --- | --- |
| OpenAPI | `type: string` | `type: string` + `enum: [1m, 5m, 15m, 1h, 1d]` |
| Orval client | `bucket?: string` | `bucket?: '1m' \| '5m' \| '15m' \| '1h' \| '1d'` |

### Notes

- Bad numeric values are skipped with `slog.Warn`, matching existing `min=` / `max=` handling
- No new exported API, no breaking changes

### DIFF
--- a/schema_customizer.go
+++ b/schema_customizer.go
@@ -20,6 +20,7 @@ import (
 //   - min=1 => minLength=1 (for strings)
 //   - max=100 => max=100 (for integers)
 //   - max=100 => maxLength=100 (for strings)
+//   - oneof=a b c => enum=[a,b,c] (matches go-playground/validator)
 func parseValidate(tag reflect.StructTag, schema *openapi3.Schema) {
 	validateTag, ok := tag.Lookup("validate")
 	if !ok {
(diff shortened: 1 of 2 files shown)


dylanhitt (approved):
Awesome! Thank you.

Sorry this took me so long.

PR #746, opened 2026-05-03
feat: add MiddlewareConfig.DisableControllerSection

0-don:
## Summary

Adds a `DisableControllerSection` knob to `MiddlewareConfig`, mirroring the existing `DisableMiddlewareSection`. Lets users opt out of the auto-injected `#### Controller:` block in operation descriptions, which exposes Go package paths in the public OpenAPI spec.

When both sections are disabled, `DefaultDescription` returns an empty string so the route's own description flows through without a leading `---` separator.

### Changes

- **engine.go**: add `DisableControllerSection bool` to `MiddlewareConfig`, wire it through `WithMiddlewareConfig`
- **net_http_mux.go**: skip the `#### Controller:` line when disabled; return `""` when both sections are off
- **option_test.go**: 3 new subtests in `TestOptionDescription`: controller-only disabled, both disabled with description, both disabled without description

### DIFF
--- a/net_http_mux.go
+++ b/net_http_mux.go
@@ -206,16 +206,30 @@ func camelToHuman(s string) string {
 
 // DefaultDescription returns a default .md description for a controller
 func DefaultDescription[T any](handler string, middlewares []T, middlewareConfig *MiddlewareConfig) string {
-	// Start with the controller info
-	description := "#### Controller: \n\n`" + handler + "`"
+	middlewaresDisabled := middlewareConfig.DisableMiddlewareSection || len(middlewares) == 0
+
+	// If both sections are disabled, emit nothing so the route's own
+	// description flows through without a leading separator.
+	if middlewareConfig.DisableControllerSection && middlewaresDisabled {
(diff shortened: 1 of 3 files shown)


net_http_mux.go
```diff
@@ -206,16 +206,30 @@ func camelToHuman(s string) string {
 
 // DefaultDescription returns a default .md description for a controller
 func DefaultDescription[T any](handler string, middlewares []T, middlewareConfig *MiddlewareConfig) string {
-	// Start with the controller info
-	description := "#### Controller: \n\n`" + handler + "`"
+	middlewaresDisabled := middlewareConfig.DisableMiddlewareSection || len(middlewares) == 0
```
dylanhitt, line 209: Nice clean up with the len check 👍

dylanhitt (approved):
My only comment is maybe this isn't the best place for this. But I really don't see a better place for it yet. So for now lets keep it here and maybe some commonality will present itself with a different field

Thank you again 😄

PR #797, opened 2026-07-16
Propose to change schema name generation

tan-tony:
I had ran into an edge case where there are two (or more) structs with the same name but under different packages. 
The openapi generation would success but the output for the second struct is completely wrong. 
From investigation, it had lead me to conclude that the structs with same name would overlap each other as result of transformTypeName() outputting the same string. 

My thought here is to change the schema name to include the package to guarantee global uniqueness. One of the possible downside here would potentially be the schema name is now less human readable... Please have a review on this and let me know your thoughts. 

Thanks,

### DIFF
--- a/engine_test.go
+++ b/engine_test.go
@@ -92,10 +92,10 @@ func TestWithRequestContentType(t *testing.T) {
 
 		content := route.Operation.RequestBody.Value.Content
 		require.NotNil(t, content["application/json"])
-		assert.Equal(t, "#/components/schemas/ReqBody", content["application/json"].Schema.Ref)
+		assert.Equal(t, "#/components/schemas/github.com_go-fuego_fuego_ReqBody", content["application/json"].Schema.Ref)
 
 		require.NotNil(t, content["application/xml"])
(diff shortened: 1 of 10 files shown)


dylanhitt:
At one point we did include the full package name (i think). Regardless of it being not very readable, it also exposes internal details for private projects at the openapi spec level, many of the users of this project just produce spec for public consumption and do not want their internal details being leaked to the public.

tan-tony:
@dylanhitt  I see, that makes sense. What about the approach of generating an unique id, such as a hash based on the package. That way the full package path remains hidden and could still avoid collisions from same struct names. 

Also noting that [Controller:] and [Middlewares:] reveals full package path and func name too.

dylanhitt:
Hey @tan-tony, yes I'm aware of the controllers/middlewares both of these can be disabled and the route description will just be passed through. As for this case why not bring your own SchemaCustomizer. The feature was created for situations like this.

Also I found the original reasoning for this: https://github.com/go-fuego/fuego/issues/332#issuecomment-2577613584. May not be valid for this case though. Didn't dig that deep.

tan-tony:
Ok, I'll do that for my own use case for now. 
If the code change doesn't make sense, please feel free to close this.

PR #860, opened 2026-09-07
feat: add fuego.WithoutHeaders to disable default response headers

### LINKED ISSUE #839: Disable `Server-Timing` Trailers on server

Dart/Flutter does not support working with responses that include `Server-Timing` Trailers. I would like the ability to disable Server-Timing natively in fuego. 

An option in `fuego.NewServer` that is something to the effect of `fuego.WithoutTrailers()`

Right now I have to write a middleware to strip the headers.

```go
func stripTrailers(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		strip := httpsnoop.Hooks{
			WriteHeader: func(next httpsnoop.WriteHeaderFunc) httpsnoop.WriteHeaderFunc {
				return func(status int) {
					w.Header().Del("Trailer")
					next(status)
				}
			},
		}
		next.ServeHTTP(httpsnoop.Wrap(w, strip), r)
	})
}
```

**Additional context**
This also breaks with vite API proxies on all requests without stripping the headers.

dylanhitt:
Closes #839

Just went ahead and did an across the board disable on these headers.

@bwees I believe this will solve your issue.

### DIFF
--- a/engine.go
+++ b/engine.go
@@ -47,6 +47,7 @@ type Engine struct {
 
 	requestContentTypes  []string
 	responseContentTypes []string
+	disableFlowHeaders   bool
 }
 
 type OpenAPIConfig struct {
@@ -185,6 +186,12 @@ func WithErrorHandler(errorHandler func(ctx context.Context, err error) error) E
 
 		e.ErrorHandler = errorHandler
 	}
+}
+
+// WithoutFlowHeaders disables the default headers added by Fuego in the Flow
+// (e.g. X-Powered-By, Server-Timing).
+func WithoutFlowHeaders() EngineOption {
+	return func(e *Engine) { e.disableFlowHeaders = true }
 }
 
 // DisableErrorHandler overrides ErrorHandler with a simple pass-through
--- a/serve.go
+++ b/serve.go
@@ -153,34 +153,41 @@ type ContextFlowable[B, P any] interface {
 	SerializeError(err error)
 }
 
+func setFlowHeader[B, P any](e *Engine, ctx ContextFlowable[B, P], k, v string) {
+	if e.disableFlowHeaders {
+		return
+	}
+	ctx.SetHeader(k, v)
+}
+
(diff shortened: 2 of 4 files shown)


engine.go
```diff
@@ -187,6 +188,12 @@ func WithErrorHandler(errorHandler func(ctx context.Context, err error) error) E
 	}
 }
 
+// WithoutHeaders disables the default headers added by Fuego in the Flow
+// (e.g. X-Powered-By, Server-Timing).
+func WithoutHeaders() EngineOption {
```
bwees: I feel like this should be WithoutFlowHeaders because WithoutHeaders seems too generalized for what is actually being removed.

dylanhitt: I was thinking that, kinda thought people would ask 'what the hell is "flow"'.

dylanhitt: I do agree that WithoutHeaders is a bit to generalized. I was thinking 

- WithoutFuegoHeaders but that seems redundant given the call would look like `fuego.WithoutFuegoHeaders`

bwees: Reading WithoutHeaders seems like you are disabling all headers which isnt what this is doing. I feel like a small documentation section on this option should suffice.

bwees: > I do agree that WithoutHeaders is a bit to generalized. I was thinking
> 
> * WithoutFuegoHeaders but that seems redundant given the call would look like `fuego.WithoutFuegoHeaders`

I mean even that seems to general since this is only for configuring Flow parameters. I think its fine to name it WithoutFlowHeaders since thats exactly what it does. I'd rather have the function just say what it does rather than having to leave a comment next to it in my code.

dylanhitt: Updated to you suuggestion thanks for taking a look.

dylanhitt: I'll probably cut a tag at the end of week, feel free to respond here again if I forget.

